PR #1, opened 2021-06-14
Add readme and owners

sgreene570:
This PR adds preliminary OWNERS and README.md files.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,21 @@
+approvers:
+  - knobunc
+  - Miciah
+  - frobware
+  - danehans
+  - sgreene570
+  - candita
+  - rfredette
+  - alebedev87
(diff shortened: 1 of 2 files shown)


OWNERS
```diff
@@ -0,0 +1,21 @@
...
+  - knobunc
+  - Miciah
+  - frobware
+  - danehans
+  - sgreene570
+  - candita
+  - rfredette
+  - alebedev87
+  - miheer
+component: DNS
```
Miciah, line 21: We should think about adding a new Bugzilla component for ExternalDNS (and likewise for Contour).

sgreene570: Noted. We can worry about that once we are at a point where we are filing bugzilla reports :)

PR #2, opened 2021-06-15
Add Initial Kubebuilder scaffolding

sgreene570:
This PR lays out initial scaffolding for the operator via kubebuilder in the following commits:

---

**Initialize kubebuilder scaffolding**

This commit is the result of running the following commands
via kubebuilder v3.1.0:

```
kubebuilder init --domain openshift.io --repo openshift.io/external-dns-operator
kubebuilder create api --group operator --kind ExternalDNS --version v1alpha1
```

Also, this commit contains resources generated by running `make` and `make test`.

Signed-off-by: Stephen Greene <sgreene@redhat.com>

---

**go.mod: Update initial dependencies**

Update go.mod & go.sum via the following commands:

```
go get sigs.k8s.io/controller-runtime@v0.9.0
go mod tidy
```

Signed-off-by: Stephen Greene <sgreene@redhat.com>

---

**Update .gitignore**

Add `bin/` & `testbin/` to .gitignore to avoid committing binaries.

Signed-off-by: Stephen Greene <sgreene@redhat.com>

PR #3, opened 2021-06-17
Clean up kubebuilder scaffolding and add `vendor/`

sgreene570:
This PR is a follow-up to #2 and contains the following commits:

---

**Strip out leader election code and yaml** 

The initial kubebuilder scaffolding includes code and resources for managing operator leader election, which we probably don't need for the ExternalDNS operator.

**Add vendor directory for go modules** 

`vendor/`:

Run go mod vendor and commit the outputted `vendor` directory.

`.gitignore`:

Remove comments referencing `vendor/`.

**Add tools/tools.go for vendoring controller-gen** 

`tools/tools.go`:

Add `controller-gen` to a phony `tools` package so that `controller-gen` source is baked into `vendor/`.

`go.mod`/`go.sum`:

Explicitly add `controller-gen` dependency.

`vendor/`:

Vendor controller-gen dependency.

`bin/`:

Stop tracking the controller-gen binary.
 
**Update test Makefile directive** 

`Makefile`:

Add new `CONTROLLER_GEN` variable for running the vendored copy of controller-gen.

Remove `controller-gen` directive as we no longer need to fetch controller-gen during build time.

Modify the `test` directive to call `go test` (prior to this commit, the `test` directive pulled binaries to perform mock e2e tests. We will add e2e tests for real clusters in the near future).

`.gitignore`:

Remove `testbin/` since that directory is no longer used at all.

**Update base CRD controller-gen build version**


Miciah:
Can we omit the `bin/controller-gen` binary file?

sgreene570:
> Can we omit the `bin/controller-gen` binary file?

interesting, I dont think we can actually since that can't be downloaded during CI runs. Let me amend the .gitignore.

Miciah:
> > Can we omit the `bin/controller-gen` binary file?
> 
> interesting, I dont think we can actually since that can't be downloaded during CI runs. Let me amend the .gitignore.

You should be able to vendor the package (following the example here: https://github.com/openshift/cluster-ingress-operator/blob/master/tools/tools.go#L13) and then build it (or run it using `go run`).  (I'm surprised we _cannot_ download it during CI runs, but we shouldn't _need_ to download it if we package the source and build in CI.)

sgreene570:
> > > Can we omit the `bin/controller-gen` binary file?
> > 
> > 
> > interesting, I dont think we can actually since that can't be downloaded during CI runs. Let me amend the .gitignore.
> 
> You should be able to vendor the package (following the example here: https://github.com/openshift/cluster-ingress-operator/blob/master/tools/tools.go#L13) and then build it (or run it using `go run`). (I'm surprised we _cannot_ download it during CI runs, but we shouldn't _need_ to download it if we package the source and build in CI.)

This was my thinking as well. I'll update this Pr shortly! Thanks.

/hold

sgreene570:
Alright, thanks for the input @Miciah ! This PR should be good to go now, PTAL when you get the chance! :grin: 

/hold cancel

sgreene570:
> I'm surprised we _cannot_ download it during CI runs, but we shouldn't _need_ to download it if we package the source and build in CI.

I believe CI builds happen in a more or less disconnected environment since slorping binaries from the public internet into builds is absolutely a security concern.

PR #4, opened 2021-06-22
Nerf unit tests and fix build directive

sgreene570:
**Makefile: Force builds to use vendor/**

Add necessary go flags to the `make build` directive.

**Remove mock-cluster unit test** 

`rm controllers/suite_test.go`

The suite_test.go unit test file creates a mock cluster using etcd,
among other binaries. This does not intermix well with our CI
environment.

This commit removes this file. We will add unit tests in the near
future.


Makefile
```diff
@@ -60,8 +60,9 @@ test: manifests generate fmt vet ## Run unit tests
 
 ##@ Build
 
+GO=GO111MODULE=on GOFLAGS=-mod=vendor CGO_ENABLED=0 go
```
Miciah, line 63: Presumably `CGO_ENABLED=0` is key because the other variables are on by default with the version of Go we currently use, right?

sgreene570: That, and -mod=vendor

Miciah: According to https://golang.org/ref/mod#vendoring, "If the `vendor` directory is present in the main module's root directory, it will be used automatically if the `go` version in the main module's `go.mod` file is 1.14 or higher."  So it seems like specifying `GOFLAGS=-mod=vendor` is redundant—but it doesn't hurt anything.

sgreene570: Huh, TIL. Still, can't hurt to be explicit.

PR #6, opened 2021-06-25
OpenShift Dockerfile added

alebedev87:


Makefile
```diff
@@ -61,8 +61,9 @@ test: manifests generate fmt vet ## Run unit tests
 ##@ Build
 
 GO=GO111MODULE=on GOFLAGS=-mod=vendor CGO_ENABLED=0 go
-build: generate fmt vet ## Build manager binary.
+buildbin:
```
sgreene570, line 64: can we call this `build-manager` instead? 

And to be clear this new directive is just to avoid copying the `$(GO)` call into `Dockerfile.openshift`?
If so, can you put a comment above `buildbin` explaining it's purpose?

Dockerfile.openshift
```diff
@@ -0,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9 AS builder
+WORKDIR /external-dns-operator
+COPY . .
+RUN make buildbin
```
sgreene570, line 4: Should we make this change for `Dockerfile` as well? I don't have a strong preference between using the Makefile's compile call, or an explicit call to `go build` in the dockerfiles, but I do have a preference for consistency.

Makefile
alebedev87: Sure, will call it `build-manager`

Yes, to avoid the copying, and I didn't want to use `build` target as it depends on `generate` target which generates some code.

Dockerfile
```diff
@@ -2,26 +2,16 @@
...
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
+RUN make build-manager
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
-COPY --from=builder /workspace/manager .
+COPY --from=builder /workspace/bin/manager .
```
sgreene570, line 14: Does this require also changing ENTRYPOINT?

alebedev87: Changing ENTRYPOINT: no.
Yes we can, `bin/manager` is the output path which was generated by kubebuilder.
I'd prefer the binary to be compiled in `.` ?

sgreene570: oh, right, misread this. Looks good!

sgreene570:
thanks!
/lgtm

Dockerfile
sgreene570: For some reason I thought the `.` was before the `/workspace/..` argument :joy:

PR #9, opened 2021-07-01
Iterate on externaldnscontroller and add basic namespace reconciliation

sgreene570:
This PR consists of the following commits:

**Move kubebuilder RBAC tags to operator.go**

pkg/operator/controller/externaldns/controller.go:
Remove kubebuilder RBAC tags.

pkg/operator/operator.go:
Move kubebuilder RBAC tags to all live in a single
location in operator.go. Add RBAC tags for namespace permissions.

config/rbac/role.yaml:
Run `make manifests`.

**externaldnscontroller: Implement New(...) function**
6a6a96f
pkg/operator/controller/externaldns/controller.go:

Implement a `New(...)` function for creating the externaldns
controller with customizations.

pkg/operator/operator.go:
Update externaldns controller creation calls.

Use a non-caching client, similair to that in the
ingress and DNS operators.

**externaldns: Add basic namespace reconciliation logic** 

pkg/operator/controller/externaldns/namespace.go:

Add necessary functions to implement
`ensureExternalDNSNamespace`.

pkg/operator/controller/externaldns/controller.go:

Call `ensureExternalDNSNamespace` from the reconcile loop
of the externaldns controller.

### DIFF
--- a/pkg/operator/controller/externaldns/namespace.go
+++ b/pkg/operator/controller/externaldns/namespace.go
@@ -1,0 +1,80 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package externaldnscontroller
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ensureExternalDNSNamespace ensures that the externaldns namespace exists.
+func (r *reconciler) ensureExternalDNSNamespace(ctx context.Context, namespace string) (bool, *corev1.Namespace, error) {
+	desired := desiredExternalDNSNamespace(namespace)
+
+	haveNamespace, current, err := r.currentExternalDNSNamespace(ctx, namespace)
+	if err != nil {
+		return false, nil, err
+	}
+
+	if !haveNamespace {
+		if err := r.createExternalDNSNamespace(ctx, desired); err != nil {
+			return false, nil, err
+		}
+		return r.currentExternalDNSNamespace(ctx, namespace)
+	}
+
+	return true, current, nil
+}
+
+// currentExternalDNSNamespace  gets the current externalDNS namespace resource.
+func (r *reconciler) currentExternalDNSNamespace(ctx context.Context, namespace string) (bool, *corev1.Namespace, error) {
+	ns := &corev1.Namespace{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: namespace}, ns); err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil, nil
+		}
+		return false, nil, err
+	}
+	return true, ns, nil
+
+}
+
+// desiredExternalDNSNamespace returns the desired namespace resource.
+func desiredExternalDNSNamespace(namespace string) *corev1.Namespace {
+	return &corev1.Namespace{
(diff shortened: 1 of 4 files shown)


pkg/operator/controller/externaldns/namespace.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return false, nil, err
+	}
+
+	switch {
+	case !haveNamespace:
+		if err := r.createExternalDNSNamespace(desired); err != nil {
+			return false, nil, err
+		}
+		return r.currentExternalDNSNamespace(namespace)
```
alebedev87, line 43: `desired` variable already holds the instance of the namespace object, no need to get it from API again

pkg/operator/controller/externaldns/namespace.go
```diff
@@ -0,0 +1,81 @@
...
+// ensureExternalDNSNamespace ensures that the externaldns namespace exists.
+func (r *reconciler) ensureExternalDNSNamespace(namespace string) (bool, *corev1.Namespace, error) {
+	desired := desiredExternalDNSNamespace(namespace)
+
+	haveNamespace, current, err := r.currentExternalDNSNamespace(namespace)
+	if err != nil {
+		return false, nil, err
+	}
+
+	switch {
```
alebedev87, line 38: there is only one condition; do we really need a `switch` here?

pkg/operator/controller/externaldns/namespace.go
```diff
@@ -0,0 +1,81 @@
...
+		return r.currentExternalDNSNamespace(namespace)
+	}
+
+	return true, current, nil
+}
+
+// currentExternalDNSNamespace  gets the current externalDNS namespace resource.
+func (r *reconciler) currentExternalDNSNamespace(namespace string) (bool, *corev1.Namespace, error) {
+	ns := &corev1.Namespace{}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: namespace}, ns); err != nil {
```
alebedev87, line 52: `context` object can be passed to `ensureExternalDNSNamespace()` (and deeper) from `Reconcile`

pkg/operator/operator.go
```diff
@@ -52,6 +52,12 @@ type Operator struct {
 	log     logr.Logger
 }
 
+// Aggregate kubebuilder RBAC tags in one location for simplicity.
```
alebedev87, line 55: I thought that the idea of having these comments in the same file as `Reconcile()` was to keep them closer to where it's more visible which Kubernetes resources are used.

pkg/operator/controller/externaldns/namespace.go
sgreene570: Actually, some other controllers might add annotations to the namespace, so it's worth fetching it again IMO.

sgreene570: This is the pattern we use in the ingress and DNS operators :)

pkg/operator/controller/externaldns/namespace.go
sgreene570: Fixed. We can add the switch statement back need be in the future.

pkg/operator/controller/externaldns/namespace.go
sgreene570: done! Good idea, thanks for pointing that out.

pkg/operator/operator.go
sgreene570: I guess this is a subjective thing. IMO it's nicer to have all of the RBAC markers in the same spot so you don't have to bounce between files to find a specific RBAC tag. Thoughts? I don't feel super strongly either way, but I think I would prefer to leave all of the RBAC markers in one place, since this is how the contour operator does it https://github.com/projectcontour/contour-operator/blob/main/internal/operator/operator.go

alebedev87: Ok, let's keep it consistent between these 2 operators.

pkg/operator/controller/externaldns/namespace.go
sgreene570: Speaking of which, ive added the necessary change to use a non-caching client to this PR. Thanks again!

Should we design the externalDNS operator to (safely) use the default split client, or can we follow suit with the DNS and Ingress operators here?

sgreene570: My preference leans towards the approach used in the DNS/Ingress operators simply because it's a pattern I am comfortable with.

alebedev87: As discussed in Slack, for starters let's use the not caching client to be sure of the coherent sequential writes/reads.

PR #12, opened 2021-07-06
Add v1alpha1 ExternalDNS Operator API.

sgreene570:
**Add v1alpha1 ExternalDNS Operator API**

Adds a preliminary (v1alpha1) cluster-scoped ExternalDNS
Operator CRD under the `externaldns.olm.openshift.io` group name.

This commit is in support of
https://issues.redhat.com/browse/NE-640.

Enhancement proposal available here:
https://github.com/openshift/enhancements/blob/master/enhancements/dns/externaldns-operator.md

api/v1alpha1/externaldns_types.go:

Add Go code for ExternalDNS operator API.

api/v1alpha1/groupversion_info.go:

Set CRD group version.

api/v1alpha1/zz_generated.deepcopy.go:

Run `make generate`.

config/crd/bases/operator.openshift.io_externaldns.yaml:

Run `make generate` and pluralize `externaldnses` for the
`ExternalDNSList` type

**Update ExternalDNS API calling sites**

config/:
Use new `externaldns.olm.openshift.io` API group.
Delete the scaffolded kustomize patches as they are not needed.
Drop deleted field from sample externalDNS yaml.

pkg/operator/operator.go:
Update API group/resource in RBAC markers.

---

This PR supersedes https://github.com/openshift/api/pull/956


sgreene570:
We need to determine which API group to use for the new externalDNS operator CRD.

/hold

api/v1alpha1/groupversion_info.go
```diff
@@ -26,7 +26,8 @@ import (
 
 var (
 	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "operator.openshift.io", Version: "v1alpha1"}
+	// TODO: settle on an appropriate group name.
+	GroupVersion = schema.GroupVersion{Group: "externaldns.olm.openshift.io", Version: "v1alpha1"}
```
deads2k, line 30: this looks ok to me.  Check with @ecordell to see what's standard here.  operator.openshift.io is managed in openshift/api, so we want a unique group since this will not be coupled to an OCP payload and will update at a different cadence.

We can also carve off a subdomain under operator.openshfit.io if necessary, but your likely names may impact your future API choices that you *do* want coupled to OCP release cadence.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+// +genclient
```
deads2k, line 25: I expected this to be cluster-scoped.  You could have multiple instances in a single cluster, but control of DNS seems like a cluster-admin scoped power.

Were you thinking of namespace scoping?  If so, what sort of topology do you have in mind?  Something more akin to hypershift?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+
+	// Name is a string representing a single domain
+	// value.
+	//
+	// e.g. my-app.my-cluster-domain.com
+	//
+	// +optional
+	Name *string `json:"names,omitempty"`
+
+	// Pattern is a regular expression used to
```
deads2k, line 144: you need to specify the regex language. They aren't all the same.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	// Domains specifies which domains that ExternalDNS should
+	// create DNS records for. Multiple domain values
+	// can be specified such that subdomains of an included domain
+	// can effectively be ignored using the "Include" and "Exclude"
+	// domain filter options.
+	//
+	// An entry that excludes a domain that isn't a subdomain of some
+	// included domain will be ignored.
+	//
+	// An empty list of domains means ExternalDNS will create
```
deads2k, line 54: This behavior in combination with the exclude behavior of "an exclude that isn't included is ignored" results in seemingly pathological behavior where I have to go from, "everything by default" to "enumerate everything" before I can exclude anything.

Why not have an exclude always win?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+}
+
+// ExternalDNSSpec defines the desired state of the ExternalDNS.
 type ExternalDNSSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Domains specifies which domains that ExternalDNS should
+	// create DNS records for. Multiple domain values
+	// can be specified such that subdomains of an included domain
+	// can effectively be ignored using the "Include" and "Exclude"
```
deads2k, line 48: If I exclude a thing that I previously included, what happens to the DNS entry?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	ExternalDNSProviderUnion `json:",inline"`
+
+	// Credentials is a reference to a secret containing
+	// the relevant credentials for the given provider, to be
+	// used by an ExternalDNS controller for creating, updating,
+	// and viewing DNS records.
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Credentials *corev1.LocalObjectReference `json:"credentials"`
```
deads2k, line 179: Don't re-use this type.  If you want a reference to a secret, create a LocalSecretReference type in this package.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+
+// ExternalDNSProviderUnion describes optional fields for an ExternalDNS
+// provider that should be captured.
+// +union
+type ExternalDNSProviderUnion struct {
+	// Type describes which DNS provider
+	// ExternalDNS should publish records to.
+	// The following DNS providers are supported:
+	//
+	//  * AWS
```
deads2k, line 190: I'm surprised that you don't require structured information for each of these providers in order to make connections to them.  How do you find them?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	// More providers will ultimately be added in the future.
+)
+
+// ExternalDNSSource describes which Source resource
+// the ExternalDNS should create DNS records for.
+type ExternalDNSSource struct {
+	ExternalDNSSourceUnion `json:",inline"`
+
+	// HostnameAnnotationPolicy specifies whether or not ExternalDNS
+	// should ignore the "external-dns.alpha.kubernetes.io/hostname"
```
deads2k, line 220: This seems to give considerable, unvalidated power to individual services or routes.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	// ExternalDNS will be configured to create
+	// DNS records for.
+	//
+	// Multiple ExternalDNS CRs must be
+	// created if multiple ExternalDNS source resources
+	// are desired.
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Source ExternalDNSSource `json:"source"`
```
deads2k, line 79: Why am I only allowed to choose one?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	// against source resource annotations.
+	//
+	// +optional
+	AnnotationFilter map[string]string `json:"annotationFilter,omitempty"`
+
+	// Namespace instructs ExternalDNS to only acknowledge
+	// source resource instances in a specific namespace.
+	//
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
deads2k, line 261: This looks severely limiting.  I would expect namespace selection to use label selection where the individual namespace name could be used as a selector if desired.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	//
+	// If no service types are provided, ExternalDNS will be
+	// configured to create DNS records for LoadBalancer services
+	// only by default.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:={"LoadBalancer"}
+	// +kubebuilder:validation:MinItems=1
+	// +required
+	ServiceType []corev1.ServiceType `json:"serviceType,omitempty"`
```
deads2k, line 335: Why have this option in addition to publishpolicy?  This appears to cover all the cases that PublicExternal does with better granularity.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+
+type ServicePublishPolicy string
+
+// +kubebuilder:validation:Enum=PublishExternal;PublishInternalExternal
+const (
+	ServicePublishPolicyExternalOnly        ServicePublishPolicy = "PublishExternal"
+	ServicePublishPolicyExternalAndInternal ServicePublishPolicy = "PublishInternalExternal"
+)
+
+type ExternalDNSCRDSourceOptions struct {
```
deads2k, line 346: When using this option, how do you know the host you're supposed to add to DNS?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	// Version is the API version
+	// of the given resource kind for
+	// ExternalDNS to use.
+	//
+	// e.g. "externaldns.k8s.io/v1alpha1"
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Version string `json:"version"`
```
deads2k, line 367: Why is the version here valuable?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	// Kind is the kind of the CRD
+	// source resource type to be
+	// consumed by ExternalDNS.
+	//
+	// e.g. "DNSEndpoint"
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Kind string `json:"kind"`
```
deads2k, line 356: Why use a kind instead of a resource?  Kinds are ambiguous mappings.  Resources are unambiguous.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	LabelFilter map[string]string `json:"labelFilter,omitempty"`
+}
+
+// ExternalDNSStatus defines the observed state of ExternalDNS.
 type ExternalDNSStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Conditions is a list of operator-specific conditions
+	// and their status.
+	Conditions []ExternalDNSOperatorCondition `json:"conditions,omitempty"`
```
deads2k, line 382: metav1.Conditions

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+
+	// ObservedGeneration is the most recent generation observed.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Provider is the configured provider in use by ExternalDNS.
+	Provider ExternalDNSProvider `json:"provider,omitempty"`
```
deads2k, line 388: Why is this in status?  I'm betting you reject changes to it, but explain.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
 }
 
-//+kubebuilder:object:root=true
+var (
+	// Available indicates that the ExternalDNS is available.
+	ExternalDNSAvailableConditionType = "Available"
+
+	// AuthenticationFailed indicates that there were issues starting
+	// ExternalDNS pods related to the given provider credentials.
+	ExternalDNSProviderAuthFailedReasonType = "AuthenticationFailed"
```
deads2k, line 421: Change the world.  you can use this if there is some external force on you, but also use Available and Degraded, where degraded could have AuthenticationFailed as a reason.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Source ExternalDNSSource `json:"source"`
+
+	// Zones describes which DNS Zone IDs
+	// ExternalDNS should publish records to.
+	//
+	// +optional
+	Zones []string `json:"zones,omitempty"`
```
deads2k, line 85: You have exceeded my DNS knowledge. If zones change, you probably have to react.  You need to describe the cleanup process.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,412 @@ package v1alpha1
...
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+//
+// ExternalDNS describes a managed ExternalDNS controller instance for a cluster.
+// The controller is responsible for creating external DNS records in supported
+// DNS providers based off of instances of select Kubernetes resources.
+type ExternalDNS struct {
```
deads2k, line 33: When I delete my instance, how are you managing the cleanup?  My expectation is a finalizer as you tidy up, but how can you distinguish between a dns record this operand creates and a DNS record someone manually created?

api/v1alpha1/externaldns_types.go
sgreene570: [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) can only create DNS records for one source type per instance . We wanted to make the "fan out" of externalDNS pods more obvious by tieing a single deployment of ExternalDNS to a single instance of the ExternaLDNS CR.

api/v1alpha1/externaldns_types.go
sgreene570: Alas, this is a shortcoming of ExternalDNS, which can only observe all namespaces, or one namespace per instance (and on top of that, only one source resource per instance).
https://github.com/kubernetes-sigs/external-dns/issues/1454
We could support a namespace label selector here, but the fan out wouldn't be obvious.

api/v1alpha1/externaldns_types.go
sgreene570: Changed to cluster-scoped, since that makes more sense than namespace scoped. Thanks!

api/v1alpha1/externaldns_types.go
sgreene570: > Why not have an exclude always win?
Great idea, lets go with that. I've gone ahead and updated this logic. Thanks!

api/v1alpha1/externaldns_types.go
sgreene570: Right. I've spiced this comment up. Thanks!

api/v1alpha1/externaldns_types.go
sgreene570: In retrospect, im not sure why I thought adding an unnecessary publish policy was a good idea.

I've removed the publish policy field and elaborated on the implications of using the ClusterIP service type in the comments above.

api/v1alpha1/externaldns_types.go
sgreene570: I've gone ahead and created a `NamespacedSecretReference` type in this package since the CRD is now cluster-scoped. @deads2k does that sound like a reasonable thing to do?

sgreene570: If theres a better type I should use for a namespaced secret reference, let me know.

api/v1alpha1/externaldns_types.go
sgreene570: > don't require structured information for each of these providers
We don't at the moment, but that could change in the future. Is there something that immediately jumps out to you as missing? 
> How do you find them?
Could you elaborate on what you mean? 

cc @Miciah

api/v1alpha1/externaldns_types.go
sgreene570: The ExternalDNS Source implementation expects crd resources to have specific fields, including a DNSName field.

https://github.com/kubernetes-sigs/external-dns/blob/master/docs/contributing/crd-source.md

Does that answer your question?

api/v1alpha1/externaldns_types.go
sgreene570: We were planning on using this status field to reflect the currently configured provider, so a user would know if their provider credentials are valid.

api/v1alpha1/externaldns_types.go
sgreene570: this variable is for a reason type, maybe im missing something?

api/v1alpha1/externaldns_types.go
sgreene570: Switched to metav1.conditions, since that should be upstreamable.

Thanks.

api/v1alpha1/externaldns_types.go
sgreene570: Right, thats why we thought it would make sense to disable this functionality by default.

api/v1alpha1/externaldns_types.go
sgreene570: How would finalizers apply in this scenario? ExternalDNS creates DNS records for resources in the cluster, such as services or routes. It wouldn't make sense for ExternalDNS to delete the cluster resources themselves. Arguably, it might make sense to additionally leave external DNS record cleanup to a cluster-administrator.

api/v1alpha1/externaldns_types.go
sgreene570: Nothing. I've expanded on the docs here.

api/v1alpha1/externaldns_types.go
sgreene570: Right. I beefed up the docs to include a warning against enabling this. Thoughts?

api/v1alpha1/externaldns_types.go
sgreene570: For both the Kind and Version fields, this field parralell how ExternalDNS is configured via command-line flags.

With that in mind, is using Version and Kind flags acceptable (Is upstream using the wrong terminology, perhaps?)

e.g.

```
Version: "externaldns.k8s.io/v1alpha1"
Kind: "DNSEndpoint"
```

api/v1alpha1/externaldns_types.go
sgreene570: See https://github.com/openshift/external-dns-operator/pull/12#discussion_r665715937

api/v1alpha1/externaldns_types.go
sgreene570: I've elaborated in the godocs. I think it makes sense for the operator to do nothing in this case.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,400 @@ package v1alpha1
...
+	// +optional
+	Service *ExternalDNSServiceSourceOptions `json:"service,omitempty"`
+
+	// CRD describes source configuration options specific
+	// to the CRD source resource.
+	//
+	// +optional
+	CRD *ExternalDNSCRDSourceOptions `json:"crd,omitempty"`
+
+	// OpenShiftRoute source currently has no unique configuration options.
```
sgreene570, line 305: @deads2k in your opinion, should we stub this type here, or leave it out like it is right now.

api/v1alpha1/externaldns_types.go
sgreene570: Would it be best to have separate fields for resource group and resource version? Or no version field?

api/v1alpha1/externaldns_types.go
alebedev87: It looks like it's technically feasible to find all the records owned by an instance of ExternalDNS.
Each DNS record created by ExternalDNS is accompanied by a corresponding TXT record which contains the ownerID (must be unique for each ExternalDNS instance).
It's possible to list all the records of the given zones and filtering on TXT records which would allow us to check the ownership.
I think that I would prefer to see this clean-up functionality in ExteranlDNS rather than in the operator due to the number of supported providers.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,400 @@ package v1alpha1
...
+	ExternalDNSProviderUnion `json:",inline"`
+
+	// Credentials is a reference to a secret containing
+	// the relevant credentials for the given provider, to be
+	// used by an ExternalDNS controller for creating, updating,
+	// and viewing DNS records.
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Credentials NamespacedSecretReference `json:"credentials"`
```
alebedev87, line 189: The credentials format seem to vary from provider to provider. InfoBlox uses username/password, Azure and BlueCat some json file config, AWS (in case kube2iam is used) needs IAM role name, etc.
I can think of 2 options here:
- elaborate the comment with expected formats for different providers
- add dedicated fields for each supported provider

api/v1alpha1/externaldns_types.go
sgreene570: Right, my concern was more so figuring out how to tell ExternalDNS to explicitly delete a set of records. Do you know if there is a way to do that without modifying the record's source resources?

api/v1alpha1/externaldns_types.go
sgreene570: I believe the Bluecat provider is moving away from a json config file for storing credentials and other information. Let me double check how the other providers handle credentials and update this accordingly.

sgreene570:
Latest push (commit 0767d56) breaks out specific provider options into structured fields.

api/v1alpha1/externaldns_types.go
sgreene570: I've gone ahead and added fields for each providers specific configuration options.

api/v1alpha1/externaldns_types.go
sgreene570: I've added structured information for each provider.

api/v1alpha1/externaldns_types.go
Miciah: ExternalDNS lists all CRDs in the specified API group/version and uses the first one with a matching kind: https://github.com/kubernetes-sigs/external-dns/blob/2605c5ee93de4c04c62482211e887e7c7c3e95f5/source/crd.go#L82.  If we used resource here, the operator would have to translate resource to kind for ExternalDNS.  Using resource is semantically what we want, but using kind is what ExternalDNS does, so using resource here would add extra complexity and opportunities for failure.

api/v1alpha1/externaldns_types.go
Miciah: If credentials are invalid, that can be reflected in status conditions.  

I had in mind that `spec.provider` could be omitted, in which case the operator would supply defaults and use `status.provider` field to communicate those defaults.  However, my current thinking is that the operator may create a default ExternalDNS CR with `spec.provider` filled in, so defaulting just the field may be unnecessary, and I see `spec.provider` is required, so maybe `status.provider` can be omitted.

sgreene570: My mistake. 
For now I will omit `status.provider`. Thanks!

sgreene570:
Latest push added fixups for API calling sites / yaml. Seems to have thrown github off a bit, my apologies (you may now have to expand the api/v1alpha1/externaldns_types.go file tab to see the actual API changes in this PR).

api/v1alpha1/externaldns_types.go
alebedev87: Without modifying the source - no, I didn't find anything for it.

api/v1alpha1/groupversion_info.go
sgreene570: I see the SR-IOV operator  (also installed via OLM) uses `sriovnetwork.openshift.io`. Could we jump a level up here and use `externaldns.openshift.io`?
Thoughts @ecordell ? TIA

sgreene570:
@deads2k @Miciah PR is ready for review now that provider-specific parameters have been added.

config/rbac/role.yaml
```diff
@@ -18,7 +18,7 @@ rules:
   - update
   - watch
 - apiGroups:
-  - operator.openshift.io
+  - externaldns.olm.openshift.io
   resources:
   - externaldns
```
alebedev87, line 23: I faced this problem during the test of OLM bundle: I changed CRD name to `externaldnses.externaldns.olm.openshift.io` to use the future name.
The problem is that the resources is `externaldnsES` (as specified in CRD).

sgreene570:
@deads2k could you take another look when you get the chance?

api/v1alpha1/externaldns_types.go
deads2k: > [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) can only create DNS records for one source type per instance. We wanted to make the "fan out" of externalDNS pods more obvious by tieing a single deployment of ExternalDNS to a single instance of the ExternalDNS CR.

This API is for managing an operator, right?  So your operator could create however many instances are required, right?

sgreene570: Yes. We just wanted the # of ExternalDNS containers/pods to be more apparent to end users. cc @Miciah

api/v1alpha1/externaldns_types.go
deads2k: I don't see doc updates for this. But I think you need to
1. cleanup external-dns created DNS entries.  If this feature is missing now, it seem important to create rather than only having orphaning policy
2. the operator needs to clean up the external-dns controller itself.
it seems like a finalizer may be required.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,546 @@ package v1alpha1
...
+	// An empty list of domains means ExternalDNS will create
+	// DNS records for any included source resource regardless
+	// of the resource's desired hostname.
+	//
+	// Populating Domains with only excluded options means ExternalDNS
+	// will create DNS records for any included source resource that do not
+	// match the provided excluded domain options.
+	//
+	// Excluding DNS records that were previous included via a resource update
+	// will *not* result in the original DNS records being deleted.
```
deads2k, line 61: This seems like a design issue.  Is there an option to avoid orphaning entries?

api/v1alpha1/externaldns_types.go
deads2k: > We just wanted the # of ExternalDNS containers/pods to be more apparent to end users

Why is this an objective?  I would expect an API to be optimized for users to express their intent.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,546 @@ package v1alpha1
...
+	Source ExternalDNSSource `json:"source"`
 
-	// Foo is an example field of ExternalDNS. Edit externaldns_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// Zones describes which DNS Zone IDs
+	// ExternalDNS should publish records to.
+	//
+	// Updating this field after creation
+	// will cause all DNS records in the previous
+	// zone(s) to be left behind.
```
deads2k, line 91: The orphan policy on everything concerns me in terms of trying this feature out and later removing it.  Having an orphan policy makes sense, but I would expect a cleanup policy to be the default.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,546 @@ package v1alpha1
...
+	// https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts.keys
+	// and https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/gke.md
+	// for more information
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	Credentials NamespacedSecretReference `json:"credentials"`
+}
+
+type ExternalDNSAzureProviderOptions struct {
```
deads2k, line 273: These are a huge improvement.

api/v1alpha1/externaldns_types.go
deads2k: > I believe the Bluecat provider is moving away from a json config file for storing credentials and other information. Let me double check how the other providers handle credentials and update this accordingly.

If the format is not owned by us, how do you plan to plan to handle changes over time and smooth upgrades?

api/v1alpha1/externaldns_types.go
deads2k: > Right. I beefed up the docs to include a warning against enabling this. Thoughts?

Are you sure you even want this knob? Why not require a custom CR?

api/v1alpha1/externaldns_types.go
deads2k: > @deads2k in your opinion, should we stub this type here, or leave it out like it is right now.

I'm ok leaving it out for now.

api/v1alpha1/externaldns_types.go
deads2k: > The ExternalDNS Source implementation expects crd resources to have specific fields, including a DNSName field.
> 
> https://github.com/kubernetes-sigs/external-dns/blob/master/docs/contributing/crd-source.md
> 
> Does that answer your question?

Yes.  Can you add this to the godoc?  Also, indicating that such a resource will be privileged with precise control over DNS entries is important.

api/v1alpha1/externaldns_types.go
deads2k: Based your answers above, I gather the schema of a particular version can matter. Having two fields seems ok.

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,546 @@ package v1alpha1
...
+	// +required
+	Version string `json:"version"`
+
+	// LabelFilter specifies a label filter
+	// to be used to filter CRD resource instances.
+	// Only one label filter can be specified on
+	// an ExternalDNS instance.
+	//
+	// +optional
+	LabelFilter map[string]string `json:"labelFilter,omitempty"`
```
deads2k, line 533: Why isn't this filter the metav1 label selector type?

api/v1alpha1/externaldns_types.go
```diff
@@ -18,41 +18,546 @@ package v1alpha1
...
+type ExternalDNSStatus struct {
+	// Conditions is a list of operator-specific conditions
+	// and their status.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ObservedGeneration is the most recent generation observed.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Zones is the configured zones in use by ExternalDNS.
+	Zones []string `json:"zones,omitempty"`
```
deads2k, line 546: Why do we track this if your policy only supports orphaning?

api/v1alpha1/externaldns_types.go
deads2k: > this variable is for a reason type, maybe im missing something?

Nope, I didn't read the block carefully enough.

api/v1alpha1/externaldns_types.go
sgreene570: Thanks!

api/v1alpha1/externaldns_types.go
sgreene570: > Why not require a custom CR?

A custom CR would be painful when creating external DNS records for LoadBalancer services.

api/v1alpha1/externaldns_types.go
sgreene570: > If the format is not owned by us, how do you plan to plan to handle changes over time and smooth upgrades?

Good question. We will have to work with upstream around issues like this (perhaps we could maintain backwards compatibility upstream to help with upgrades).

api/v1alpha1/externaldns_types.go
sgreene570: Fixed, thanks for the ideas.

api/v1alpha1/externaldns_types.go
sgreene570: to make it more obvious to administrators which DNS Zones ExternalDNS has been successfully able to bind to?
I suppose we could use generic conditions to express that...
@Miciah thoughts?

api/v1alpha1/externaldns_types.go
sgreene570: Fixed!.

Miciah:
Let's merge this and iterate.
/lgtm

PR #13, opened 2021-07-07
Add verify command for OpenShift CI

sgreene570:
hack/verify-deps.sh:
Add script to verify `vendor/` when running on OpenShift CI.

hack/verify-gofmt.sh:
Add script to verify that all operator code is gofmt compliant.

hack/verify-generated.sh:
Add script to verify that generated manifests (e.g. the ExternalDNS CRD)
and deep copy functions are up to date.

Makefile:
Add new `verify` directive that calls the verify scripts in `hack/`.

/assign @alebedev87

### DIFF
--- a/hack/verify-deps.sh
+++ b/hack/verify-deps.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
+function print_failure {
+  echo "There are unexpected changes to the vendor tree following 'go mod vendor' and 'go mod tidy'. Please"
+  echo "run these commands locally and double-check the Git repository for unexpected changes which may"
+  echo "need to be committed."
+  exit 1
+}
+
+if [ "${OPENSHIFT_CI:-false}" = true ]; then
+  go mod vendor
+  go mod tidy
+
+  test -z "$(git status --porcelain)" || print_failure
+  echo "verified Go modules"
+fi
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,11 @@ test: manifests generate fmt vet ## Run unit tests
 		-coverprofile coverage.out \
 		./...
 
+verify:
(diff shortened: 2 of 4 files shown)


hack/verify-deps.sh
```diff
@@ -0,0 +1,17 @@
...
+
+function print_failure {
+  echo "There are unexpected changes to the vendor tree following 'go mod vendor' and 'go mod tidy'. Please"
+  echo "run these commands locally and double-check the Git repository for unexpected changes which may"
+  echo "need to be committed."
+  exit 1
+}
+
+if [ "${OPENSHIFT_CI:-false}" = true ]; then
+  go mod vendor
```
alebedev87, line 12: These ones change some file on `main`:
```
$ git st --porcelain
$ go mod vendor && go mod tidy
$ git st --porcelain
 M go.mod
 M vendor/modules.txt
```
I think we can add the changes to this PR to start clean.

sgreene570: We could start clean by using some temporary commit logic? Thoughts?

sgreene570: For reference, the OPENSHIFT_CI flag is what stops this script from interfering with any locally uncommitted changes (at the cost of not actually running when you run `make verify`). The expectation in CI is that all changes will be committed (since CI will do a git clone).

alebedev87: I was just thinking about the next guy who'll do a PR, it would fail due to `verify-deps.sh`.
But it won't as there is no CI job for it but once it's there it will.

Well, it's not a big deal anyway; LGTM.

sgreene570: Oh, I misunderstood. Ill open a new PR to fix that :grin:

PR #18, opened 2021-07-16
Begin implementing ExternalDNS Deployment logic

sgreene570:
config/rbac/role.yaml:

Update operator RBAC for deployments.

go.mod & vendor/modules.txt:

Explicitly add go-cmp dep.

pkg/operator/controller/externaldns/controller.go:

Call deployment logic.

pkg/operator/controller/externaldns/deployment.go:

Add initial deployment logic.

pkg/operator/controller/externaldns/deployment_test.go:

Add initial deployment logic unit tests.

pkg/operator/operator.go:

Add deployment RBAC markers.

PR #19, opened 2021-07-23
Begin implementing deployment logic

alebedev87:
Handover from https://github.com/openshift/external-dns-operator/pull/18.
Main differences from Steve's PR:
- Missing labelselector/labels added to deployment
- `ExternalDNS.Status` made optional
- Manifests updated to use `openshift-external-dns-operator` namespace

Tried on my local cluster:
```bash
$ oc logs deploy/external-dns-operator -c operator
2021-08-04T12:15:38.200Z	INFO	using operator namespace	{"namespace": "openshift-external-dns-operator"}
2021-08-04T12:15:38.200Z	INFO	using operand namespace	{"namespace": "external-dns"}
2021-08-04T12:15:38.200Z	INFO	using ExternalDNS image	{"image": "docker.io/bitnami/external-dns:latest"}
I0804 12:15:39.252345       1 request.go:668] Waited for 1.03609362s due to client-side throttling, not priority and fairness, request: GET:https://10.217.4.1:443/apis/apps.openshift.io/v1?timeout=32s
2021-08-04T12:15:40.761Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": "127.0.0.1:8080"}
2021-08-04T12:15:43.317Z	INFO	setup	starting externalDNS operator
2021-08-04T12:15:43.320Z	INFO	controller-runtime.manager	starting metrics server	{"path": "/metrics"}
2021-08-04T12:15:43.320Z	INFO	controller-runtime.manager.controller.externaldns_controller	Starting EventSource	{"source": "kind source: /, Kind="}
2021-08-04T12:15:43.320Z	INFO	controller-runtime.manager.controller.externaldns_controller	Starting Controller
2021-08-04T12:15:43.422Z	INFO	controller-runtime.manager.controller.externaldns_controller	Starting workers	{"worker count": 1}
2021-08-04T12:15:43.440Z	INFO	externaldns_controller	created externaldns namespace	{"namespace": "external-dns"}
2021-08-04T12:15:43.492Z	INFO	externaldns_controller	created externalDNS service account	{"namespace": "external-dns", "name": "external-dns-sample"}
2021-08-04T12:15:43.545Z	INFO	externaldns_controller	created externalDNS deployment	{"namespace": "external-dns", "name": "external-dns-sample"}

$ oc get externaldns
NAME     AGE
sample   115s

$ oc -n external-dns get deploy,sa
NAME                                  READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/external-dns-sample   0/1     1            0           73m

NAME                                 SECRETS   AGE
serviceaccount/builder               2         73m
serviceaccount/default               2         73m
serviceaccount/deployer              2         73m
serviceaccount/external-dns-sample   2         73m

```

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -9,7 +9,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+        image: quay.io/openshift/origin-kube-rbac-proxy:latest
         args:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
@@ -22,6 +22,3 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
-      - name: operator
-        args:
-        - "--metrics-bind-address=127.0.0.1:8080"
--- a/config/samples/operator_v1alpha1_externaldns.yaml
+++ b/config/samples/operator_v1alpha1_externaldns.yaml
@@ -1,5 +1,18 @@
 apiVersion: externaldns.olm.openshift.io/v1alpha1
 kind: ExternalDNS
 metadata:
-  name: externaldns-sample
+  name: sample
+  namespace: external-dns-operator
 spec:
+  provider:
+    type: AWS
+    aws:
+      credentials:
+        name: aws-access-key
+        namespace: external-dns-operator
+  source:
+    type: Service
+    namespace: "publish-external-dns"
+    annotationFilter:
+      external-dns.mydomain.org/publish: "yes"
+  zones: ["myzoneid"]
--- a/pkg/operator/config/config.go
+++ b/pkg/operator/config/config.go
@@ -19,8 +19,8 @@ package config
 const (
 	DefaultExternalDNSImage  = "docker.io/bitnami/external-dns:latest"
 	DefaultMetricsAddr       = "127.0.0.1:8080"
-	DefaultOperatorNamespace = "externaldns-operator"
-	DefaultOperandNamespace  = "externaldns"
+	DefaultOperatorNamespace = "external-dns-operator"
+	DefaultOperandNamespace  = "external-dns"
 )
 
 // Config is configuration of the operator.
--- a/pkg/operator/controller/externaldns/controller.go
+++ b/pkg/operator/controller/externaldns/controller.go
@@ -66,6 +66,7 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	if err := c.Watch(&source.Kind{Type: &operatorv1alpha1.ExternalDNS{}}, &handler.EnqueueRequestForObject{}); err != nil {
 		return nil, err
 	}
@@ -76,29 +77,29 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
 // Reconcile reconciles watched objects and attempts to make the current state of
 // the object match the desired state.
 func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	result := reconcile.Result{}
-
 	externalDNS := &operatorv1alpha1.ExternalDNS{}
 	if err := r.client.Get(ctx, req.NamespacedName, externalDNS); err != nil {
 		if errors.IsNotFound(err) {
 			r.log.Info("externalDNS not found; reconciliation will be skipped", "request", req)
-			return result, nil
+			return reconcile.Result{}, nil
 		}
-		return result, fmt.Errorf("failed to get externalDNS %s: %w", req, err)
+		return reconcile.Result{}, fmt.Errorf("failed to get externalDNS %s: %w", req, err)
 	}
 
 	if _, _, err := r.ensureExternalDNSNamespace(ctx, r.config.Namespace); err != nil {
-		// Return if the externalDNS namespace cannot be created since
-		// resource creation in a namespace that does not exist will fail.
-		return result, fmt.Errorf("failed to ensure externalDNS namespace: %v", err)
+		return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS namespace: %w", err)
 	}
 
-	haveServiceAccount, _, err := r.ensureExternalDNSServiceAccount(ctx, r.config.Namespace, externalDNS)
+	haveServiceAccount, sa, err := r.ensureExternalDNSServiceAccount(ctx, r.config.Namespace, externalDNS)
 	if err != nil {
-		return result, fmt.Errorf("failed to ensure externalDNS service account: %w", err)
+		return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS service account: %w", err)
 	} else if !haveServiceAccount {
-		return result, fmt.Errorf("failed to get externalDNS service account: %w", err)
+		return reconcile.Result{}, fmt.Errorf("failed to get externalDNS service account: %w", err)
 	}
 
-	return result, nil
+	if _, _, err := r.ensureExternalDNSDeployment(ctx, r.config.Namespace, r.config.Image, sa, externalDNS); err != nil {
+		return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS deployment: %w", err)
+	}
+
+	return reconcile.Result{}, nil
 }
--- a/pkg/operator/controller/externaldns/deployment.go
+++ b/pkg/operator/controller/externaldns/deployment.go
@@ -163,7 +163,7 @@ func desiredExternalDNSDeployment(namespace, image string, serviceAccount *corev
 
 // WIP function for generating container specs for one container at a time.
 func buildExternalDNSContainer(seq int, image, zone, provider, source string, externalDNS *operatorv1alpha1.ExternalDNS) (*corev1.Container, error) {
-	name := fmt.Sprintf("externaldns-%d", seq+1)
+	name := controller.ExternalDNSContainerName(zone)
 
 	args := []string{
 		fmt.Sprintf("--metrics-address=127.0.0.1:%d", metricsStartPort+seq),
@@ -257,23 +257,71 @@ func (r *reconciler) updateExternalDNSDeployment(ctx context.Context, current, d
 // externalDNSDeploymentChanged evaluates whether or not a deployment update is necessary.
 // Returns a boolean if an update is necessary, and the deployment resource to update to.
 func externalDNSDeploymentChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
-	changed := false
 	updated := current.DeepCopy()
 
-	if len(current.Spec.Template.Spec.Containers) > 0 && len(expected.Spec.Template.Spec.Containers) > 0 {
-		if current.Spec.Template.Spec.Containers[0].Image != expected.Spec.Template.Spec.Containers[0].Image {
-			updated.Spec.Template.Spec.Containers[0].Image = expected.Spec.Template.Spec.Containers[0].Image
-			changed = true
-		}
-		currArgs := append([]string{}, current.Spec.Template.Spec.Containers[0].Args...)
-		expArgs := append([]string{}, expected.Spec.Template.Spec.Containers[0].Args...)
-		sort.Strings(currArgs)
-		sort.Strings(expArgs)
-		if !cmp.Equal(currArgs, expArgs) {
-			updated.Spec.Template.Spec.Containers[0].Args = expected.Spec.Template.Spec.Containers[0].Args
-			changed = true
+	return externalDNSContainersChanged(current, expected, updated), updated
+}
+
+// externalDNSContainersChanged returns true if the current containers differ from the expected
+func externalDNSContainersChanged(current, expected, updated *appsv1.Deployment) bool {
+	changed := false
+
+	// number of container is different: let's reset them all
+	if len(current.Spec.Template.Spec.Containers) != len(expected.Spec.Template.Spec.Containers) {
+		updated.Spec.Template.Spec.Containers = expected.Spec.Template.Spec.Containers
+		return true
+	}
+
+	currentContMap := buildIndexedContainerMap(current.Spec.Template.Spec.Containers)
+	expectedContMap := buildIndexedContainerMap(expected.Spec.Template.Spec.Containers)
+
+	// let's check that all the current containers have the desired values set
+	for currName, currCont := range currentContMap {
+		// if the current container is expected: check its fields
+		if expCont, found := expectedContMap[currName]; found {
+			if currCont.Image != expCont.Image {
+				updated.Spec.Template.Spec.Containers[currCont.Index].Image = expCont.Image
+				changed = true
+			}
+			if !equalStringSliceContent(expCont.Args, currCont.Args) {
+				updated.Spec.Template.Spec.Containers[currCont.Index].Args = expCont.Args
+				changed = true
+			}
+		} else {
+			// if the current container is not expected: let's not dig deeper - reset all
+			updated.Spec.Template.Spec.Containers = expected.Spec.Template.Spec.Containers
+			return true
 		}
 	}
 
-	return changed, updated
+	return changed
+}
+
+// indexedContainer is the standard core POD's container with additional index field
+type indexedContainer struct {
+	corev1.Container
+	Index int
+}
+
+// buildIndexedContainerMap builds a map from the given list of containers
+// key is the container name
+// value is the indexed container with index being the sequence number of the given list
+func buildIndexedContainerMap(containers []corev1.Container) map[string]indexedContainer {
+	m := map[string]indexedContainer{}
+	for i, cont := range containers {
+		m[cont.Name] = indexedContainer{
+			Container: cont,
+			Index:     i,
+		}
+	}
+	return m
+}
+
+// equalStringSliceContent returns true if 2 string slices have the same content (order doesn't matter)
+func equalStringSliceContent(sl1, sl2 []string) bool {
+	copy1 := append([]string{}, sl1...)
+	copy2 := append([]string{}, sl2...)
+	sort.Strings(copy1)
+	sort.Strings(copy2)
+	return cmp.Equal(copy1, copy2)
 }
--- a/pkg/operator/controller/externaldns/deployment.go
+++ b/pkg/operator/controller/externaldns/deployment.go
@@ -1,0 +1,279 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package externaldnscontroller
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	"github.com/google/go-cmp/cmp"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	operatorv1alpha1 "github.com/openshift/external-dns-operator/api/v1alpha1"
+	controller "github.com/openshift/external-dns-operator/pkg/operator/controller"
+)
+
+const (
+	metricsStartPort    = 7979
+	appNameLabel        = "app.kubernetes.io/name"
+	appInstanceLabel    = "app.kubernetes.io/instance"
+	masterNodeRoleLabel = "node-role.kubernetes.io/master"
+	osLabel             = "kubernetes.io/os"
+	linuxOS             = "linux"
+)
+
+// providerStringTable maps ExternalDNSProviderType values from the
+// ExternalDNS operator API to the provider string argument expected by ExternalDNS.
+var providerStringTable = map[operatorv1alpha1.ExternalDNSProviderType]string{
+	operatorv1alpha1.ProviderTypeAWS:      "aws",
+	operatorv1alpha1.ProviderTypeGCP:      "google",
+	operatorv1alpha1.ProviderTypeAzure:    "azure",
+	operatorv1alpha1.ProviderTypeBlueCat:  "bluecat",
+	operatorv1alpha1.ProviderTypeInfoblox: "infoblox",
+}
+
+// sourceStringTable maps ExternalDNSSourceType values from the
+// ExternalDNS operator API to the source string argument expected by ExternalDNS.
+var sourceStringTable = map[operatorv1alpha1.ExternalDNSSourceType]string{
+	operatorv1alpha1.SourceTypeRoute:   "openshift-route",
+	operatorv1alpha1.SourceTypeService: "service",
+	// TODO: Add CRD source support
+}
+
+// ensureExternalDNSDeployment ensures that the externalDNS deployment exists.
+// Returns a Boolean value indicating whether the deployment exists, a pointer to the deployment, and an error when relevant.
+func (r *reconciler) ensureExternalDNSDeployment(ctx context.Context, namespace, image string, serviceAccount *corev1.ServiceAccount, externalDNS *operatorv1alpha1.ExternalDNS) (bool, *appsv1.Deployment, error) {
+	nsName := types.NamespacedName{Namespace: namespace, Name: controller.ExternalDNSResourceName(externalDNS)}
+
+	desired, err := desiredExternalDNSDeployment(namespace, image, serviceAccount, externalDNS)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to build externalDNS deployment: %w", err)
+	}
+
+	exist, current, err := r.currentExternalDNSDeployment(ctx, nsName)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to get externalDNS deployment: %w", err)
+	}
+
+	// create the deployment
+	if !exist {
+		if err := r.createExternalDNSDeployment(ctx, desired); err != nil {
+			return false, nil, err
+		}
+		// get the deployment from API to catch up the fields added/updated by API and webhooks
+		return r.currentExternalDNSDeployment(ctx, nsName)
+	}
+
+	// update the deployment
+	if _, err := r.updateExternalDNSDeployment(ctx, current, desired); err != nil {
+		return true, current, err
+	}
+
+	// get the deployment from API to catch up the fields added/updated by API and webhooks
+	return r.currentExternalDNSDeployment(ctx, nsName)
+}
+
+// currentExternalDNSDeployment gets the current externalDNS deployment resource.
+func (r *reconciler) currentExternalDNSDeployment(ctx context.Context, nsName types.NamespacedName) (bool, *appsv1.Deployment, error) {
+	depl := &appsv1.Deployment{}
+	if err := r.client.Get(ctx, nsName, depl); err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil, nil
+		}
+		return false, nil, err
+	}
+	return true, depl, nil
+}
+
+// desiredExternalDNSDeployment returns the desired deployment resource.
+func desiredExternalDNSDeployment(namespace, image string, serviceAccount *corev1.ServiceAccount, externalDNS *operatorv1alpha1.ExternalDNS) (*appsv1.Deployment, error) {
+	replicas := int32(1)
+
+	matchLbl := map[string]string{
+		appNameLabel:     controller.ExternalDNSBaseName,
+		appInstanceLabel: externalDNS.Name,
+	}
+
+	nodeSelectorLbl := map[string]string{
+		osLabel:             linuxOS,
+		masterNodeRoleLabel: "",
+	}
+
+	depl := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      controller.ExternalDNSResourceName(externalDNS),
+			Namespace: namespace,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &replicas,
+			Selector: &metav1.LabelSelector{
+				MatchLabels: matchLbl,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: matchLbl,
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: serviceAccount.Name,
+					NodeSelector:       nodeSelectorLbl,
+				},
+			},
+		},
+	}
+
+	provider, ok := providerStringTable[externalDNS.Spec.Provider.Type]
+	if !ok {
+		return nil, fmt.Errorf("unsupported provider: %q", externalDNS.Spec.Provider.Type)
+	}
+	source, ok := sourceStringTable[externalDNS.Spec.Source.Type]
+	if !ok {
+		return nil, fmt.Errorf("unsupported source type: %q", externalDNS.Spec.Source.Type)
+	}
+
+	for i, zone := range externalDNS.Spec.Zones {
+		container, err := buildExternalDNSContainer(i, image, zone, provider, source, externalDNS)
+		if err != nil {
+			return nil, err
+		}
+		depl.Spec.Template.Spec.Containers = append(depl.Spec.Template.Spec.Containers, *container)
+	}
+
+	return depl, nil
+}
+
+// WIP function for generating container specs for one container at a time.
+func buildExternalDNSContainer(seq int, image, zone, provider, source string, externalDNS *operatorv1alpha1.ExternalDNS) (*corev1.Container, error) {
+	name := fmt.Sprintf("externaldns-%d", seq+1)
+
+	args := []string{
+		fmt.Sprintf("--metrics-address=127.0.0.1:%d", metricsStartPort+seq),
+		fmt.Sprintf("--txt-owner-id=externaldns-%s", externalDNS.Name),
+		fmt.Sprintf("--zone-id-filter=%s", zone),
+		fmt.Sprintf("--provider=%s", provider),
+		fmt.Sprintf("--source=%s", source),
+		"--policy=sync",
+		"--registry=txt",
+		"--log-level=debug",
+	}
+
+	//TODO: Add provider credentials logic
+
+	if externalDNS.Spec.Source.Namespace != nil && len(*externalDNS.Spec.Source.Namespace) > 0 {
+		args = append(args, fmt.Sprintf("--namespace=%s", *externalDNS.Spec.Source.Namespace))
+	}
+
+	if len(externalDNS.Spec.Source.AnnotationFilter) > 0 {
+		annotationFilter := ""
+		for key, value := range externalDNS.Spec.Source.AnnotationFilter {
+			annotationFilter += fmt.Sprintf("%s=%s,", key, value)
+		}
+		args = append(args, fmt.Sprintf("--annotation-filter=%s", annotationFilter[0:len(annotationFilter)-1]))
+	}
+
+	if externalDNS.Spec.Source.Service != nil && len(externalDNS.Spec.Source.Service.ServiceType) > 0 {
+		serviceTypeFilter, publishInternal := "", false
+		for _, serviceType := range externalDNS.Spec.Source.Service.ServiceType {
+			serviceTypeFilter += string(serviceType) + ","
+			if serviceType == corev1.ServiceTypeClusterIP {
+				publishInternal = true
+			}
+		}
+
+		// avoid having a trailing comma
+		args = append(args, fmt.Sprintf("--service-type-filter=%s", serviceTypeFilter[0:len(serviceTypeFilter)-1]))
+
+		if publishInternal {
+			args = append(args, "--publish-internal-services")
+		}
+	}
+
+	if externalDNS.Spec.Source.HostnameAnnotationPolicy == operatorv1alpha1.HostnameAnnotationPolicyIgnore {
+		args = append(args, "--ignore-hostname-annotation")
+	}
+
+	//TODO: Add logic for the CRD source.
+
+	return &corev1.Container{
+		Name:                     name,
+		Image:                    image,
+		ImagePullPolicy:          corev1.PullIfNotPresent,
+		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+		Args:                     args,
+	}, nil
+}
+
+// createExternalDNSDeployment creates the given deployment using the reconciler's client.
+func (r *reconciler) createExternalDNSDeployment(ctx context.Context, depl *appsv1.Deployment) error {
+	if err := r.client.Create(ctx, depl); err != nil {
+		return fmt.Errorf("failed to create externalDNS deployment %s/%s: %w", depl.Namespace, depl.Name, err)
+	}
+	r.log.Info("created externalDNS deployment", "namespace", depl.Namespace, "name", depl.Name)
+	return nil
+}
+
+// updateExternalDNSDeployment updates the in-cluster externalDNS deployment.
+// Returns a boolean if an update was made, and an error when relevant.
+func (r *reconciler) updateExternalDNSDeployment(ctx context.Context, current, desired *appsv1.Deployment) (bool, error) {
+	// don't always update (or do simple DeepEqual with) the operand's deployment
+	// as this may result into a "fight" between API/admission webhooks and this controller
+	// example:
+	//  - this controller creates a deployment with the desired fields A, B, C
+	//  - API adds some default fields D, E, F (e.g. metadata, imagePullPullPolicy, dnsPolicy)
+	//  - mutating webhooks add default values to fields E, F
+	//  - this controller gets into reconcile loop and starts all from step 1
+	//  - checking that fields A, B, C are the same as desired would save us the before mentioned round trips:
+	changed, updated := externalDNSDeploymentChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+
+	if err := r.client.Update(ctx, updated); err != nil {
+		return false, fmt.Errorf("failed to update externalDNS deployment %s/%s: %w", desired.Namespace, desired.Name, err)
+	}
+	r.log.Info("updated externalDNS deployment", "namespace", desired.Namespace, "name", desired.Name)
+	return true, nil
+}
+
+// externalDNSDeploymentChanged evaluates whether or not a deployment update is necessary.
+// Returns a boolean if an update is necessary, and the deployment resource to update to.
+func externalDNSDeploymentChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
+	changed := false
+	updated := current.DeepCopy()
+
+	if len(current.Spec.Template.Spec.Containers) > 0 && len(expected.Spec.Template.Spec.Containers) > 0 {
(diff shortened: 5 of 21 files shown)


config/samples/operator_v1alpha1_externaldns.yaml
```diff
@@ -1,5 +1,15 @@
 apiVersion: externaldns.olm.openshift.io/v1alpha1
 kind: ExternalDNS
 metadata:
-  name: externaldns-sample
+  name: sample
+  namespace: external-dns-operator
```
candita, line 5: I'll wait for @Miciah to weigh in, but I think you should use `openshift-externaldns-operator` for the namespace

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		// getting the deployment from API in case a third party updated it after the creation
+		return r.currentExternalDNSDeployment(ctx, namespace, externalDNS)
+	}
+
+	// update the deployment
+	if _, err := r.updateExternalDNSDeployment(ctx, desired); err != nil {
+		return true, current, err
+	}
+
```
candita, line 77: If there any validation that might return an error here?  If so, return the error to the caller so they can figure out what might be wrong.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,224 @@
...
+		if err := r.createExternalDNSDeployment(ctx, desired); err != nil {
+			return false, nil, err
+		}
+		// getting the deployment from API in case a third party updated it after the creation
+		return r.currentExternalDNSDeployment(ctx, namespace, externalDNS)
+	}
+
+	// update the deployment
+	if _, err := r.updateExternalDNSDeployment(ctx, desired); err != nil {
+		return true, current, err
```
candita, line 75: It returns true if there was an error?

config/samples/operator_v1alpha1_externaldns.yaml
alebedev87: I was thinking about it too. However OpenShift gave me an error while I tried to create `openshift-external-operator`  namespace for external-dns-opertor and I was cluster-admin. It seems to be that there is some hard coded check for `openshift-*` namespaces.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Yes, it was like this before but I'm not particularly against it. This boolean indicates the existence of the deployment, not the error.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,224 @@
...
+	if externalDNS.Spec.Source.Namespace != nil && len(*externalDNS.Spec.Source.Namespace) > 0 {
+		args = append(args, fmt.Sprintf("--namespace=%s", *externalDNS.Spec.Source.Namespace))
+	}
+
+	if externalDNS.Spec.Source.Service != nil && len(externalDNS.Spec.Source.Service.ServiceType) > 0 {
+		serviceTypeFilter, publishInternal := "", false
+		for i, serviceType := range externalDNS.Spec.Source.Service.ServiceType {
+			serviceTypeFilter += string(serviceType)
+			// avoid having a trailing comma
+			if len(externalDNS.Spec.Source.Service.ServiceType)-1 > i {
```
candita, line 178: You can just do this once after the loop exits.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,224 @@
...
+				serviceTypeFilter += ","
+			}
+			args = append(args, fmt.Sprintf("--service-type-filter=%s", serviceTypeFilter))
+
```
candita, line 182: Do you want one instance of the `service-type-filter` parameter, or several?

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,224 @@
...
+		Args:                     args,
+	}, nil
+}
+
+// createExternalDNSDeployment creates the given deployment using the reconciler's client.
+func (r *reconciler) createExternalDNSDeployment(ctx context.Context, depl *appsv1.Deployment) error {
+	if err := r.client.Create(ctx, depl); err != nil {
+		return fmt.Errorf("failed to create externalDNS deployment %s/%s: %w", depl.Namespace, depl.Name, err)
+	}
+	r.log.Info("created externalDNS deployment", "namespace", depl.Namespace, "name", depl.Name)
```
candita, line 212: For Info logs, better to use klog and the flags for verbosity: `klog.V(4).Infof`

pkg/operator/controller/externaldns/deployment.go
alebedev87: It's possible to specify several but I think that it's a little more convenient to see them all on one line.

alebedev87: Ah, now I see, yeah, it appends every time which doesn't make sense, good catch!

pkg/operator/controller/externaldns/deployment.go
alebedev87: Moved the append of the flag outside of `for` loop.

pkg/operator/controller/externaldns/deployment.go
alebedev87: `controller-runtime` based operators and controller use [logr](https://pkg.go.dev/github.com/go-logr/logr) compatible structured logging, not `klog`. About the verbosity level, here were are logging at the default level (without specifying `V()`) so that this will be seen in both [development or production](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.9.5/pkg/log/zap#Options).

config/samples/operator_v1alpha1_externaldns.yaml
Miciah: Yeah, the operator should use the `openshift-externaldns-operator` namespace.  What was the error?

alebedev87: My bad, I naively tested the namespace creation with `oc new-project` which gave me this error:
```bash
$ oc new-project openshift-externaldns-operator
Error from server (Forbidden): project.project.openshift.io "openshift-externaldns-operator" is forbidden: cannot request a project starting with "openshift-"
```
Will change the namespace to `openshift-externaldns-operator`.

pkg/operator/controller/externaldns/deployment.go
candita: Nit: the len doesn't change, you can just set a variable once outside of the loop.  Or just append the comma every time and remove the last one afterward.

alebedev87: Done, remove the trailing comma after the loop

alebedev87:
After a discussion with @Miciah, added back the logic for the comparison: current vs desired.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,261 @@
...
+	return nil
+}
+
+// updateExternalDNSDeployment updates the in-cluster externalDNS deployment.
+// Returns a boolean if an update was made, and an error when relevant.
+func (r *reconciler) updateExternalDNSDeployment(ctx context.Context, current, desired *appsv1.Deployment) (bool, error) {
+	// don't always update (or do simple DeepEqual with) the operand's deployment
+	// as this may result into a "fight" between API/admission webhooks and this controller
+	// example:
+	//  - this contorller creates a deployment with the desired fields A, B, C
```
candita, line 222: contorller -> controller

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,261 @@
...
+	//  - API adds some default fields D, E, F (i.e. metadata, imagePullPullPolicy, dnsPolicy)
+	//  - mutating webhooks add default values to fields E, F
+	//  - this controller gets into reconcile loop and starts all from step1
+	//  - checking that fields A, B, C are the same as desired would save us the before mentioned round trips:
+	changed, updated := externalDNSDeploymentChanged(current, desired)
+	if !changed {
+		return false, nil
+	}
+
+	if err := r.client.Update(ctx, updated); err != nil {
```
candita, line 232: Shouldn't there be some kind of validation on the changes prior to making the update?

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	return true, depl, nil
+}
+
+// desiredExternalDNSDeployment returns the desired deployment resource.
+func desiredExternalDNSDeployment(namespace, image string, serviceAccount *corev1.ServiceAccount, externalDNS *operatorv1alpha1.ExternalDNS) (*appsv1.Deployment, error) {
+	replicas := int32(1)
+
+	matchLbl := map[string]string{
+		"app.kubernetes.io/name":     "external-dns",
```
candita, line 101: Would be nice to have all these strings declared as consts

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,261 @@
...
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: serviceAccount.Name,
+					NodeSelector:       nodeSelectorLbl,
+				},
+			},
+		},
+	}
+
+	metricsPort := 7979
```
candita, line 132: Maybe this should be declared as a const too

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,261 @@
...
+	for i, zone := range externalDNS.Spec.Zones {
```
candita, line 133: Do we have a sense of how many Zones might there be?  Is is possible that there might be port conflicts if we start venturing into the hundreds of Zones?

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,261 @@
...
+func externalDNSDeploymentChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
+	changed := false
+	updated := current.DeepCopy()
+
+	if len(current.Spec.Template.Spec.Containers) > 0 && len(expected.Spec.Template.Spec.Containers) > 0 {
+		if current.Spec.Template.Spec.Containers[0].Image != expected.Spec.Template.Spec.Containers[0].Image {
+			updated.Spec.Template.Spec.Containers[0].Image = expected.Spec.Template.Spec.Containers[0].Image
+			changed = true
+		}
+		if !cmp.Equal(current.Spec.Template.Spec.Containers[0].Args, expected.Spec.Template.Spec.Containers[0].Args) {
```
candita, line 250: I suppose if there's any chance the args are ordered differently, it might be better to compare them iteratively.

pkg/operator/controller/externaldns/deployment.go
alebedev87: I don't think that we have an idea about how many zones can be here, in my mind it's quite a small number (1, 2 digits). I think we may have port conflicts in almost any range. It's the proximity to the popular `8000` range which you think is not good? I didn't see externalDNS listening on `8xxx` ports so far.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Well, yes, theoretically you can edit or patch the deployment which would change the order of args. This would result into an update. Don't know about the purpose of such a change though but it can be done accidentally.

pkg/operator/controller/externaldns/deployment.go
alebedev87: I put a constant for externalDNS to `names.go` file in `pkg/operator/controller/` package.
For the label names, I think it's fine to have them defined here. I don't see a big advantage of putting them as `const` at the beginning of the file, I see them more useful to be defined closer to the instantiation of `appsv1.Deployment`.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Steve didn't do any validation and for the moment I cannot come up with any either. Also, I think that if something will come we'd better to put it into an admission webhook (validating or mutating for defaults).

pkg/operator/controller/externaldns/deployment.go
alebedev87: Just noticed that in my next PR I used the same value for the toleration's key as the node selector's label.
Will move this value to const and to keep it consistent will do this for all.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Changed to the comparison of the sorted arg slices

pkg/operator/controller/externaldns/deployment.go
candita: We usually do some validation around `Reconcile`, but it could be picked up in a later task.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	operatorv1alpha1 "github.com/openshift/external-dns-operator/api/v1alpha1"
+	controller "github.com/openshift/external-dns-operator/pkg/operator/controller"
+)
+
+const (
+	startPortMetrics    = 7979
```
candita, line 37: Just for clarity, please rename to metricsPort or metricsStartPort

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if err != nil {
+		t.Errorf("expected no error from calling desiredExternalDNSDeployment, but received %v", err)
+	}
+
+	if len(depl.Spec.Template.Spec.Containers) != len(externalDNS.Spec.Zones) {
+		t.Errorf("expected externalDNS deployment to have %d containers, but found %d", len(externalDNS.Spec.Zones), len(depl.Spec.Template.Spec.Containers))
+	}
+
+	port := 7979
```
candita, line 84: 7979 -> metricsPort

pkg/operator/controller/externaldns/deployment.go
Miciah: On OpenShift, there will be 1 or 2 zones, and I doubt any practical use cases outside of OpenShift would call for more than a few zones.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	// getting the deployment from API in case a third party updated it after the update
+	return r.currentExternalDNSDeployment(ctx, namespace, externalDNS)
+}
+
+// currentExternalDNSDeployment gets the current externalDNS deployment resource.
+func (r *reconciler) currentExternalDNSDeployment(ctx context.Context, namespace string, externalDNS *operatorv1alpha1.ExternalDNS) (bool, *appsv1.Deployment, error) {
+	depl := &appsv1.Deployment{}
+	if err := r.client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: externalDNS.Name}, depl); err != nil {
```
Miciah, line 97: This should use the same name as `desiredExternalDNSDeployment`:
```suggestion
	if err := r.client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: controller.ExternalDNSResourceName(externalDNS)}, depl); err != nil {
```

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -22,6 +22,3 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
-      - name: operator
-        args:
-        - "--metrics-bind-address=127.0.0.1:8080"
```
Miciah, line 27: Why delete this?

config/samples/operator_v1alpha1_externaldns.yaml
```diff
@@ -1,5 +1,15 @@
...
 spec:
+  provider:
+    type: AWS
+    aws:
+      credentials:
+        name: aws-access-key
+        namespace: openshift-external-dns-operator
+  source:
+    type: Service
+  zones: ["dnszone"]
```
Miciah, line 15: Would it be better to leave `spec.zones` unspecified?  On platforms where zones can be autodetected, leaving it unspecified will allow this sample to work out of the box, if the credentials can be configured appropriately.  On other platforms, the sample won't work anyway.

pkg/operator/config/config.go
```diff
@@ -17,10 +17,11 @@ limitations under the License.
 package config
 
 const (
-	DefaultExternalDNSImage  = "docker.io/bitnami/external-dns:latest"
-	DefaultMetricsAddr       = "127.0.0.1:8080"
-	DefaultOperatorNamespace = "externaldns-operator"
-	DefaultOperandNamespace  = "externaldns"
+	DefaultExternalDNSImage     = "docker.io/bitnami/external-dns:latest"
```
Miciah, line 20: We should use a quay.io pullspec, but maybe that should be in a follow-up.

pkg/operator/config/config.go
```diff
@@ -17,10 +17,11 @@ limitations under the License.
...
+	DefaultMetricsAddr          = "127.0.0.1:8080"
+	DefaultOperatorNamespace    = "openshift-external-dns-operator"
+	DefaultOperandNamespace     = "external-dns"
```
Miciah, line 23: The goal being to make this upstreamable, would it make more sense for the default namespace not to include "openshift-", and to include a set of manifests or patches to configure the operator and operand to use "openshift-"-prefixed namespaces?

pkg/operator/controller/externaldns/controller.go
```diff
@@ -77,6 +81,7 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
 // the object match the desired state.
 func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	result := reconcile.Result{}
+	requeueResult := reconcile.Result{RequeueAfter: time.Duration(r.config.RequeuePeriodSeconds) * time.Second, Requeue: true}
```
Miciah, line 84: Specifying `Requeue` is superfluous if you specify a non-zero value for `RequeueAfter`.

pkg/operator/controller/externaldns/controller.go
```diff
@@ -88,16 +93,16 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	if _, _, err := r.ensureExternalDNSNamespace(ctx, r.config.Namespace); err != nil {
-		// Return if the externalDNS namespace cannot be created since
-		// resource creation in a namespace that does not exist will fail.
-		return result, fmt.Errorf("failed to ensure externalDNS namespace: %v", err)
+		return requeueResult, fmt.Errorf("failed to ensure externalDNS namespace: %w", err)
```
Miciah, line 96: Specifying a non-zero value result is superfluous when the error value is non-nil.  Specifying `Requeue` or `RequeueAfter` is only really needed if there was no error but you want to terminate reconciliation and start it over from the beginning.

pkg/operator/controller/externaldns/controller.go
```diff
@@ -88,16 +93,16 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 	}
 
-	haveServiceAccount, _, err := r.ensureExternalDNSServiceAccount(ctx, r.config.Namespace, externalDNS)
+	_, sa, err := r.ensureExternalDNSServiceAccount(ctx, r.config.Namespace, externalDNS)
 	if err != nil {
-		return result, fmt.Errorf("failed to ensure externalDNS service account: %w", err)
-	} else if !haveServiceAccount {
-		return result, fmt.Errorf("failed to get externalDNS service account: %w", err)
```
Miciah, line 100: Seems like it makes sense to keep this check.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+
+	operatorv1alpha1 "github.com/openshift/external-dns-operator/api/v1alpha1"
+	controller "github.com/openshift/external-dns-operator/pkg/operator/controller"
+)
+
+const (
+	metricsStartPort    = 7979
+	appNameLabel        = "app.kubernetes.io/name"
+	appInstanceLabel    = "app.kubernetes.io/instance"
+	masterNodeRoleLabel = "node-role.kubernetes.io/master"
```
Miciah, line 40: This label is deprecated in Kubernetes 1.20:
```suggestion
	controlPlaneNodeRoleLabel = "node-role.kubernetes.io/control-plane"
```

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+// sourceStringTable maps ExternalDNSSourceType values from the
+// ExternalDNS operator API to the source string argument expected by ExternalDNS.
+var sourceStringTable = map[operatorv1alpha1.ExternalDNSSourceType]string{
+	operatorv1alpha1.SourceTypeRoute:   "openshift-route",
+	operatorv1alpha1.SourceTypeService: "service",
+	// TODO: Add CRD source support
+}
+
+// ensureExternalDNSDeployment ensures that the externalDNS deployment exists.
+// Returns a boolean if the deployment exists, and an error when relevant.
```
Miciah, line 64:
```suggestion
// Returns a Boolean value indicating whether the deployment exists, a pointer to the deployment, and an error when relevant.
```

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to get externalDNS deployment: %w", err)
+	}
+
+	// create the deployment
+	if !exist {
+		if err := r.createExternalDNSDeployment(ctx, desired); err != nil {
+			return false, nil, err
+		}
+		// getting the deployment from API in case a third party updated it after the creation
```
Miciah, line 81: The reason for doing another get here is so that we have the object with fields that the API fills in, such as `resourceVersion` and defaulted fields, without which a subsequent update would fail.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+			return nil, err
+		}
+		depl.Spec.Template.Spec.Containers = append(depl.Spec.Template.Spec.Containers, *container)
+	}
+
+	return depl, nil
+}
+
+// WIP function for generating container specs for one container at a time.
+func populateExternalDNSContainer(image string, metricsPort int, zone string, externalDNS *operatorv1alpha1.ExternalDNS) (*corev1.Container, error) {
```
Miciah, line 154: Is this still WIP?

Would it make sense to return a non-pointer value?

Tiny suggestion: Would it make sense to name this `buildExternalDNSContainer`, to be consistent with the terminology in the "failed to build" error message in `ensureExternalDNSDeployment`?

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+	name := "externaldns-" + zone
```
Miciah, line 155: Is `zone` necessarily valid in a container name?  It might be better to hash the name, or just use a number ("externaldns-1", "externaldns-2", ...).

pkg/operator/controller/externaldns/deployment.go
Miciah: We should specify `MaxItems` for `spec.zones` in the CRD though.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+		"--log-level=debug",
+	}
+
+	provider, ok := providerStringTable[externalDNS.Spec.Provider.Type]
+	if !ok {
+		return nil, fmt.Errorf("%s is not a supported provider", externalDNS.Spec.Provider.Type)
+	}
+	args = append(args, fmt.Sprintf("--provider=%s", provider))
+
+	//TODO: Add provider credentials logic
```
Miciah, line 172: The caller could look up credentials.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+
+	source, ok := sourceStringTable[externalDNS.Spec.Source.Type]
```
Miciah, line 174: The caller could do the source lookup, same way and with the same benefits that I described in my comment about the provider lookup.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+	if !ok {
+		return nil, fmt.Errorf("%s is not a supported source type", externalDNS.Spec.Source.Type)
```
Miciah, line 176:
```suggestion
		return nil, fmt.Errorf("unsupported source type: %q", externalDNS.Spec.Source.Type)
```

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+			if serviceType == corev1.ServiceTypeClusterIP {
+				publishInternal = true
+			}
+		}
+
+		// avoid having a trailing comma
+		args = append(args, fmt.Sprintf("--service-type-filter=%s", serviceTypeFilter[0:len(serviceTypeFilter)-1]))
+
+		if publishInternal {
+			args = append(args, "--publish-internal-services")
```
Miciah, line 197: Any idea why `--publish-internal-services` is a separate flag from `--service-type-filter`?  Is it a legacy flag, or is publishing ClusterIP services dangerous or something such that it calls for a separate flag?

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// updateExternalDNSDeployment updates the in-cluster externalDNS deployment.
+// Returns a boolean if an update was made, and an error when relevant.
+func (r *reconciler) updateExternalDNSDeployment(ctx context.Context, current, desired *appsv1.Deployment) (bool, error) {
+	// don't always update (or do simple DeepEqual with) the operand's deployment
+	// as this may result into a "fight" between API/admission webhooks and this controller
+	// example:
+	//  - this controller creates a deployment with the desired fields A, B, C
+	//  - API adds some default fields D, E, F (i.e. metadata, imagePullPullPolicy, dnsPolicy)
```
Miciah, line 232:
```suggestion
	//  - API adds some default fields D, E, F (e.g. metadata, imagePullPullPolicy, dnsPolicy)
```

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	if err := r.client.Update(ctx, updated); err != nil {
+		return false, fmt.Errorf("failed to update externalDNS deployment %s/%s: %w", desired.Namespace, desired.Name, err)
+	}
+	r.log.Info("updated externalDNS deployment", "namespace", desired.Namespace, "name", desired.Name)
+	return true, nil
+}
+
+// externalDNSDeploymentChange evaluates whether or not a deployment update is necessary.
```
Miciah, line 248:
```suggestion
// externalDNSDeploymentChanged evaluates whether or not a deployment update is necessary.
```

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		currArgs := append([]string{}, current.Spec.Template.Spec.Containers[0].Args...)
+		expArgs := append([]string{}, expected.Spec.Template.Spec.Containers[0].Args...)
+		sort.Strings(currArgs)
+		sort.Strings(expArgs)
+		if !cmp.Equal(currArgs, expArgs) {
+			updated.Spec.Template.Spec.Containers[0].Args = expected.Spec.Template.Spec.Containers[0].Args
+			changed = true
+		}
+	}
```
Miciah, line 267: We want to be as impermissive of changes to operator-managed resources as possible—stomp any and all changes by users, but allow for defaulting behavior by the API and mutating admission plugins.  We can tighten the update logic up in a follow-up, but once people start using the operator, they may start doing all sorts of weird and unanticipated hacks if we allow it, and then it becomes much more difficult to tighten the operator's control over its resources.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+
+	if !changed {
+		updated = nil
+	}
```
Miciah, line 271: This seems unnecessary—once the function returns, `updated` goes away anyway.

cmd/external-dns-operator/main.go
```diff
@@ -41,6 +41,7 @@ func main() {
 	flag.StringVar(&opCfg.OperatorNamespace, "operator-namespace", operatorconfig.DefaultOperatorNamespace, "The namespace that the operator is running in.")
 	flag.StringVar(&opCfg.OperandNamespace, "operand-namespace", operatorconfig.DefaultOperandNamespace, "The namespace that ExternalDNS containers should run in.")
 	flag.StringVar(&opCfg.ExternalDNSImage, "externaldns-image", operatorconfig.DefaultExternalDNSImage, "The container image used for running ExternalDNS.")
+	flag.IntVar(&opCfg.RequeuePeriodSeconds, "requeue-period", operatorconfig.DefaultRequeuePeriodSeconds, "How much time to wait after a failed reconcile (in seconds).")
```
Miciah, line 44: What's the use case for setting a different requeue period?

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -0,0 +1,274 @@
...
+
+// ensureExternalDNSDeployment ensures that the externalDNS deployment exists.
+// Returns a boolean if the deployment exists, and an error when relevant.
+func (r *reconciler) ensureExternalDNSDeployment(ctx context.Context, namespace, image string, serviceAccount *corev1.ServiceAccount, externalDNS *operatorv1alpha1.ExternalDNS) (bool, *appsv1.Deployment, error) {
+	desired, err := desiredExternalDNSDeployment(namespace, image, serviceAccount, externalDNS)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to build externalDNS deployment: %w", err)
+	}
+
+	exist, current, err := r.currentExternalDNSDeployment(ctx, namespace, externalDNS)
```
Miciah, line 71: Instead of passing `namespace` and `externalDNS` to `currentExternalDNSDeployment` and other functions which then need to construct the namespaced name, it would be simpler to compute a `types.NamespacedName` value at the top of `ensureExternalDNSDeployment` and pass it down and use it throughout this function and its callees.  (BTW, a `types.NamespacedName` value is stringified as `namespace/name`, so it is convenient to use in error messages and log messages as well.)

Miciah (commented):
`TestExternalDNSDeploymentChanged` should have test cases for zone added, changed, or deleted.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Webhooks can modify the fields too, that's why I wrote it less precise. But it's kinda misleading, I agree.
Let me rephrase it.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Well, yes, it still needs the credentials logic.
Renamed to `buildExternalDNSContainer`.
Returning a value would make sense to not deference the pointer when it's added to `Spec.Containers`. Don't know whether something like return value optimization exists in Golang, if it does then it doesn't make much sense to return by pointer indeed.

pkg/operator/config/config.go
alebedev87: I'd like to do it too but I put on hold the setup of the mirroring for ART built external-dns images.

pkg/operator/controller/externaldns/controller.go
alebedev87: I thought of not re-queuing right after the failure (of any secondary resource creation). Some errors would not change after waiting for some time or not but some might disappear. Also, if it's not a fixable problem logging an error with some re-queue time in between would not pollute the logs that fast.

config/default/manager_auth_proxy_patch.yaml
alebedev87: I thought it was added to add the patch to find the context of where auth proxy container has to be added. But it works without it. Also `metrics-bind-address` is set to `127.0.0.1:8080` by default.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Not sure I understood what you meant.
I saw this in the enhancement doc:
> On OpenShift, the ExternalDNS Operator could create a credentialsrequest.cloudcredential.openshift.io resource so that the cluster administrator would not have to worry about providing cloud credentials by hand when creating an ExternalDNS operator resource. The ExternalDNS Operator would the be responsible for configuring the relevant ExternalDNS deployments to consume the secret produced from the credentialsrequest.cloudcredential.openshift.io resource.

But I thought this would be an option for OpenShift deployments, for the main (Kubernetes) case I thought we would still wait for the creator of `externaldns` resource to provide a secret.

pkg/operator/config/config.go
alebedev87: Hm, that sounds familiar. https://github.com/openshift/external-dns-operator/pull/19#discussion_r682107340

Miciah: Oops!  Hm, I guess we'll need to de-OpenShift-ize some other things anyway, so I guess using "openshift-" is fine.  However, if the operator namespace has the "openshift-" prefix, so should the operand namespace.

pkg/operator/controller/externaldns/controller.go
Miciah: I mean, you're returning `fmt.Errorf(...)`, which is a non-nil error value, and controller-runtime will requeue the reconciliation request immediately anyway when it gets that non-nil error value, so using `return result, fmt.Errorf(...)` should have the same effect as using `return requeueResult, fmt.Errorf(...)`.

pkg/operator/controller/externaldns/deployment.go
Miciah: That's strange.  I'll try to find out why that label isn't there.

pkg/operator/controller/externaldns/deployment.go
Miciah: I mean, the "TODO" could be in `desiredExternalDNSDeployment`, and once `desiredExternalDNSDeployment` had logic to look up the credentials, it could pass them to `buildExternalDNSContainer`.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Added `MaxItems` with the value of `10`. @Miciah, let me know which value makes sense here.

pkg/operator/controller/externaldns/controller.go
alebedev87: It's a pity it's not possible to specify `RequeueAfter` for error cases. I removed all requeue logic as for the moment it doesn't make much sense for the success case.

cmd/external-dns-operator/main.go
alebedev87: Removed the requeue logic as it doesn't make much sense for the success case, at least so far.

pkg/operator/controller/externaldns/deployment.go
alebedev87: They preferred this to be an opt-in flag: https://github.com/kubernetes-sigs/external-dns/pull/278

pkg/operator/controller/externaldns/controller.go
alebedev87: I don't know, just to return a different error message? `failed to get` vs `failed to ensure`?

config/samples/operator_v1alpha1_externaldns.yaml
alebedev87: What I had in my mind is more like having the sample which encourages the user to think about where the records should be published. The reason for this is when I first managed to get all the credentials setup ExternalDNS started to try to publish DNS records for all the services in the whole cluster and I was rather thankful of having the dns zone which didn't match any real zone ID. Once the user feels more comfortable with how ExternalDNS works, the filters can be removed.
I actually added `namespace` and `annotationFilter` to the sample too, for the same reason.

pkg/operator/config/config.go
alebedev87: I'm fine to revert all back to `external-dns-operator` namespace.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Replaced the zone with a sequence number.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Yes, I can add a TODO for the retrieval of the credentials secret into `desiredExternalDNSDeployment`. Unfortunately it's the only common thing between them, how the credentials should be consumed in externalDNS POD  is specific to each provider. So, `buildExternalDNSContainer` would still need to manage these differences therefore still needs a TODO.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Done, for service account and namespace files too

pkg/operator/controller/externaldns/deployment.go
alebedev87: You mean you'd prefer to see something like:
```golang
func buildExternalDNSContainer(image, zone, provider string, source *ExternalDNSSource)
```
as a declaration instead of passing the whole `ExternalDNS` instance?

pkg/operator/controller/externaldns/deployment.go
alebedev87: Do you have in mind some concrete fields we can additionally check?

config/samples/operator_v1alpha1_externaldns.yaml
Miciah: That makes sense.  The README suggested running `oc apply -k config/samples`, so I figured the sample was supposed to be ready to use out of the box, but it makes sense not to have the out-of-the-box example automatically do _too_ much.

pkg/operator/controller/externaldns/controller.go
Miciah: If the serviceaccount isn't there, `sa` will be nil, and `desiredExternalDNSDeployment` will deference a nil pointer.  (Granted, it's extremely unlikely that `ensureExternalDNSServiceAccount` will return false and a nil error value, but it could happen if it does a `Create` and a subsequent `Get` and the `Get` gets a non-found response.  Anyway, we get a Boolean return value here, so it makes sense to check it.)

pkg/operator/controller/externaldns/deployment.go
Miciah: I'm just suggesting that the `sourceStringTable[externalDNS.Spec.Source.Type]` lookup could be in `desiredExternalDNSDeployment`, and then `buildExternalDNSContainer` couldn't get an error.  If `desiredExternalDNSDeployment` did both the provider and source map lookups, then `buildExternalDNSContainer` would never return a non-nil error value, and the error return value could be deleted.  (Maybe as more options are added, more reasons to return an error will arise though?)

pkg/operator/controller/externaldns/deployment.go
Miciah: So it looks like it's a flag to say, "Yes, I really did mean to include ClusterIP services."  Do we need the same kind of "Yes, I really mean it" flag in the CRD?  I'd rather not, but the fact that upstream added the flag makes me wary.

pkg/operator/controller/externaldns/controller.go
alebedev87: Yes, you are right, good catch, thank you!

pkg/operator/controller/externaldns/deployment.go
alebedev87: Ah ok, I see. As a matter of fact the provider mapping can be replaced with a simple `ToLower()` as the values are already validated by OpenAPI of CRD and their ExternalDNS's equivalents are the same but in lower case.
The source unfortunately has 1 option which is drastically different in API and ExternalDNS: `OpenShiftRoute` vs `openshift-route`. The rest is also just a different case.
Well, ok, I'll pass them both as params, this will make the function a pure builder with no error handling, we'll see if this will work out (number of input params vs error output param).

pkg/operator/config/config.go
alebedev87: It's now `external-dns-operator`, will have to do something for OpenShift intergrations in the future.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Yeah, you are right, these conditions need to be handled and set in the status. I think this can be attacked in the followups.

pkg/operator/controller/externaldns/deployment.go
alebedev87: No, didn't see anything like this for CRD. There are some other guard flags for headless services but that not quite the same case as for `ClusterIP` as headless is not a service type so you cannot set in `--service-type-filter`.
```bash
  --publish-host-ip              Allow external-dns to publish host-ip for 
                                 headless services (optional)
  --always-publish-not-ready-addresses  
                                 Always publish also not ready addresses for 
                                 headless services (optional)
```

pkg/operator/controller/externaldns/deployment.go
alebedev87: Updated the change detection logic and unit tests. Hashing of the zone is added too (needed for the change detecction).
Put everything in a separate commit, just in case: https://github.com/openshift/external-dns-operator/pull/19/commits/c0c6c2b55751ad45c310a3872a138ad941cc8aa6

pkg/operator/controller/externaldns/deployment.go
alebedev87: Looked at this thing one more time. Actually you were right about the legacy.

At the time when `--publish-internal-services` was added there was no `--service-type-filter` flag. There was no way to fine tune deeper than the source level, so you could only say: I want services as source (`--source=service`) and yes I want internal ClusterIPs to be published too (`--publish-internal-services`).

Then they added  `--service-type-filter` flag in [PR#653](https://github.com/kubernetes-sigs/external-dns/pull/653) which defaulted to `all` which kept the old combination of `--source=service --publish-internal-services` still relevant. However if the filter is set to `ClusterIP` you have to specify the unnecessary `--publish-internal-services`, otherwise the IP [will not be found](https://github.com/kubernetes-sigs/external-dns/blob/master/source/service.go#L461).

pkg/operator/controller/externaldns/deployment.go
alebedev87: As per discussion in a chat, let's keep it set to 10 for the moment and we'll see if it'll need a change.

pkg/operator/controller/externaldns/deployment.go
alebedev87: It's a gap on OCP side, [BZ filed](https://bugzilla.redhat.com/show_bug.cgi?id=1995858).

pkg/operator/controller/externaldns/deployment.go
alebedev87: Added the handling of the status to TOD in README.md in the [next PR](https://github.com/openshift/external-dns-operator/pull/20).

pkg/operator/controller/externaldns/deployment.go
alebedev87: I think that how and where we lookup the credentials can be revisited when we get closer to the implementation, in the followup PRs. And I'll keep in mind this remark.

pkg/operator/controller/externaldns/deployment.go
alebedev87: As discussed in the chat, the last implementation from https://github.com/openshift/external-dns-operator/commit/7ac3c115f2696eb94d4ed5f1298374a19eb6c20d looks fine to begin with.

candita:
> `TestExternalDNSDeploymentChanged` should have test cases for zone added, changed, or deleted.

I think this is one comment that was overlooked.  Otherwise, lgtm.

alebedev87:
@candita: I addressed the comment about the zones in `deployment_test.go`. It's just that the new test cases don't mention the zones but containers, as zone == container in terms of the externalDNS deployment. The test cases are:
- `if externalDNS misses container` when a new zone was added
- `if externalDNS has extra container` when a zone was deleted

The other test cases cover the update of a zone: image or args (the only 2 things we care about for the moment).

PR #22, opened 2021-09-20
Functional operator

alebedev87:
Last peace of code to make ExternalDNS Operator functional (deployment/upgrade/removal).

What's inside:
- New controller to copy the secret from the operator into the operand namespace. Hence the change in the API for the credentials: no namespace can be specified.
- Adding of the owner reference to all the secondary resources which are related to the primary resource. This is needed for the cascade deletions. The shared resources like cluster role or namespace are not set to any ExternalDNS instance hence not removed.
- `FQDNTemplate` is now a slice in order to be able to specify multiple values separately, not comma separated like before

```
$ oc -n external-dns-operator get pods
NAME                                     READY   STATUS    RESTARTS   AGE
external-dns-operator-7d8c765bcc-btg7s   2/2     Running   0          21m

$ oc get externaldns sample
NAME     AGE
sample   93s

$ oc -n external-dns get pods
NAME                                  READY   STATUS    RESTARTS   AGE
external-dns-sample-bb5f9969d-wst8c   1/1     Running   0          2m9s

$ oc -n external-dns get secret external-dns-credentials-sample
NAME                              TYPE     DATA   AGE
external-dns-credentials-sample   Opaque   2      2m42s

$ curl -L http://alebedev-httpserver.devcluster.openshift.com
Accept=*/*
User-Agent=curl/7.61.1
Now: http alebedev-httpserver.devcluster.openshift.com
Before: http alebedev-httpserver.devcluster.openshift.com

$ oc delete externaldns sample
externaldns.externaldns.olm.openshift.io "sample" deleted

$ oc -n external-dns get deploy external-dns-sample
Error from server (NotFound): namespaces "external-dns" not found

$ oc -n external-dns get sa external-dns-sample
Error from server (NotFound): namespaces "external-dns" not found

$ oc get clusterrolebinding external-dns-sample
Error from server (NotFound): clusterrolebindings.rbac.authorization.k8s.io "external-dns-sample" not found
```


miheer:
@alebedev87  are we updating the status of the operator once we change something in the CRD ?

config/rbac/role.yaml
```diff
@@ -21,6 +21,7 @@ rules:
   - ""
   resources:
   - namespaces
+  - secrets
```
Miciah, line 24: If the secret is always in the operator's namespace, you might not need this.

Miciah: Ah, is this is so that you can create the copy in the operand's namespace?

pkg/operator/controller/credentials-secret/controller.go
```diff
@@ -0,0 +1,229 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	operatorv1alpha1 "github.com/openshift/external-dns-operator/api/v1alpha1"
+)
+
+const (
+	controllerName                  = "credentials_secret_controller"
+	finalizer                       = "externaldns.operator.kubernetes.io/finalizer-credentials-secret"
```
Miciah, line 45:
```suggestion
	finalizer                       = "externaldns.olm.openshift.io/finalizer-credentials-secret"
```

pkg/operator/controller/credentials-secret/controller.go
```diff
@@ -0,0 +1,229 @@
...
+		requests := []reconcile.Request{}
+		if err := reconciler.cache.List(context.Background(), externalDNSList, listOpts); err != nil {
+			log.Error(err, "failed to list externalDNS for secret")
+			return requests
+		}
+		for _, ed := range externalDNSList.Items {
+			log.Info("queueing externalDNS", "name", ed.Name)
+			request := reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: ed.Namespace,
```
Miciah, line 139: This is superfluous as `externaldnses` is cluster-scoped.

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -0,0 +1,125 @@
...
+	destName := controller.ExternalDNSDestCredentialsSecretName(r.config.TargetNamespace, extDNS.Name)
+	// desired is created from source
+	desired := desiredCredentialsSecret(source, destName)
+
+	if err := controllerutil.SetControllerReference(extDNS, desired, r.scheme); err != nil {
+		return false, nil, fmt.Errorf("failed to set the controller reference for credentials secret: %w", err)
+	}
+
+	// check if the destination secret exists
+	destExists, dest, err := r.currentCredentialsSecret(ctx, destName)
```
Miciah, line 54: Check `err`.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -164,8 +165,8 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
 		args = append(args, "--ignore-hostname-annotation")
 	}
 
-	if len(b.externalDNS.Spec.Source.FQDNTemplate) > 0 {
-		args = append(args, fmt.Sprintf("--fqdn-template=%s", b.externalDNS.Spec.Source.FQDNTemplate))
+	if b.externalDNS.Spec.Source.FQDNTemplate != nil && len(b.externalDNS.Spec.Source.FQDNTemplate) > 0 {
```
Miciah, line 168: Checking for `nil` is superfluous.  `len(b.externalDNS.Spec.Source.FQDNTemplate) > 0` is sufficient.

pkg/operator/controller/externaldns/rbac_test.go
```diff
@@ -49,12 +40,17 @@ func TestEnsureExternalDNSClusterRole(t *testing.T) {
...
 	}{
 		{
 			name:            "Does not exist",
 			existingObjects: []runtime.Object{},
 			expectedExist:   true,
 			expectedRole: rbacv1.ClusterRole{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ClusterRole",
+					APIVersion: "rbac.authorization.k8s.io/v1",
+				},
```
Miciah, line 53: Out of curiosity, why is `TypeMeta` needed now?

Miciah (commented):
The new controller should also watch the operator's secret to revert any changes that a user makes to it, but that isn't essential for basic functionality.  

For the test code, you might want to consider using helper functions to construct the expected objects in order to reduce code duplication.  For example, you could use a simpler helper like https://github.com/openshift/cluster-dns-operator/blob/5b0d54e348bf0f74aaa63e089b75b98793396cb7/pkg/operator/controller/dns_status_test.go#L200-L214, or you could use the pattern  in https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/service/jig.go where you have a `CreateFoo` method or function that takes a "tweak" function and returns a "tweaked" version of some general object.  

I didn't see any major problems, and this PR has had a few reviews now, so feel free to cancel the hold and address any remaining issues with a follow-up if needed.

pkg/operator/controller/externaldns/rbac_test.go
alebedev87: Good question, I didn't find the answer. It's somewhere in the difference of `reflect.DeepEqual` (used before) and `go-cmp` (used now).
I don't know why but `go-cmp` manages to compare `TypeMeta` and `ObjectMeta` embedded structs while reflect's `DeepEqual` does not.

I added TypeMeta fields to the ignore list to safe some space. I think that the type check is enough.

alebedev87:
Answer to https://github.com/openshift/external-dns-operator/pull/22#issuecomment-926710634
@miheer : no, not yet, it' something we just started to work on.

config/rbac/role.yaml
alebedev87: Yes, it's for the operand namespace's secret.
Technically we need the access to 1 namespace only but this namespace is not fixed but configurable via a flag.

alebedev87:
Answer to https://github.com/openshift/external-dns-operator/pull/22#pullrequestreview-763496707.

@Miciah: yes, the amount of the boilerplate code in the tests can be optimized. I already did this for [the deployment test](https://github.com/openshift/external-dns-operator/blob/f6360b2168bce719407df9e344aafec37cca10c2/pkg/operator/controller/externaldns/deployment_test.go#L994-L1010) but not everywhere. Some tests were quite small, creating helper functions for them didn't make much sense but now they grew up and this may become a problem.
This will need to be addressed in the follow ups.

miheer:
> Answer to [#22 (comment)](https://github.com/openshift/external-dns-operator/pull/22#issuecomment-926710634)
> @miheer : no, not yet, it' something we just started to work on.

OK do let me know if you need help on this as it is a big component.

pkg/operator/controller/externaldns/rbac_test.go
miheer: Were the fields rbacv1.ClusterRole{}, "ResourceVersion", "Kind", "APIVersion"  ignored  when you used reflect.DeepEqual ?

miheer: in the test case I did the following change

if gotExist {
                //diffOpts := cmpopts.IgnoreFields(rbacv1.ClusterRole{}, "ResourceVersion", "Kind", "APIVersion")
                //if diff := cmp.Diff(tc.expectedRole, *gotRole, diffOpts); diff != "" {
                if diff := reflect.DeepEqual(tc.expectedRole, *gotRole); diff != true {
                    t.Errorf("unexpected cluster role (-want +got):\n%v\n%v", tc.expectedRole, *gotRole)
                }
            }
        })
    }

and in rbac.go

//if diff := cmp.Diff(expectedRuleMap, currentRuleMap); diff != "" {
if diff := reflect.DeepEqual(expectedRuleMap, currentRuleMap); diff != true {
   return true, fmt.Sprintf("diff found in the policy rules: \n%v\n%v", expectedRuleMap, currentRuleMap)
}

I got this ->

rbac_test.go:176: unexpected cluster role (-want +got):
        {{ } {external-dns      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}
        {{ClusterRole rbac.authorization.k8s.io/v1} {external-dns     1000 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}

in the expected {ClusterRole rbac.authorization.k8s.io/v1} is missing

that is why the reflect is failing.

Using cmp as you ignore these fields then the test will pass

miheer: test output when I use reflect

[miheer@localhost external-dns-operator]$ 
[miheer@localhost external-dns-operator]$ GO111MODULE=on go test -v -run TestEnsureExternalDNSClusterRole ./pkg/...
?   	github.com/openshift/external-dns-operator/pkg/operator	[no test files]
?   	github.com/openshift/external-dns-operator/pkg/operator/config	[no test files]
testing: warning: no tests to run
PASS
ok  	github.com/openshift/external-dns-operator/pkg/operator/controller	(cached) [no tests to run]
testing: warning: no tests to run
PASS
ok  	github.com/openshift/external-dns-operator/pkg/operator/controller/credentials-secret	(cached) [no tests to run]
=== RUN   TestEnsureExternalDNSClusterRole
=== RUN   TestEnsureExternalDNSClusterRole/Does_not_exist
2021-09-28T00:45:51.495+1000	INFO	created externalDNS cluster role	{"name": "external-dns"}
    rbac_test.go:176: unexpected cluster role (-want +got):
        {{ } {external-dns      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}
        {{ClusterRole rbac.authorization.k8s.io/v1} {external-dns     1 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}
=== RUN   TestEnsureExternalDNSClusterRole/Exists_and_as_expected
    rbac_test.go:176: unexpected cluster role (-want +got):
        {{ } {external-dns      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}
        {{ClusterRole rbac.authorization.k8s.io/v1} {external-dns     999 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}
=== RUN   TestEnsureExternalDNSClusterRole/Exists_and_needs_to_be_updated
2021-09-28T00:45:51.496+1000	INFO	updated externalDNS cluster role	{"name": "external-dns", "reason": "diff found in the policy rules: \nmap[/endpoints:[get list watch] /nodes:[get list watch] /pods:[get list watch] /services:[get list watch] networking.k8s.io/ingresses:[get list watch]]\nmap[/endpoints:[get list watch] /pods:[get list watch] /services:[get list watch] networking.k8s.io/ingresses:[get list watch]]"}
    rbac_test.go:176: unexpected cluster role (-want +got):
        {{ } {external-dns      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}
        {{ClusterRole rbac.authorization.k8s.io/v1} {external-dns     1000 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} [{[get list watch] [networking.k8s.io] [ingresses] [] []} {[get list watch] [] [endpoints services pods nodes] [] []}] nil}
--- FAIL: TestEnsureExternalDNSClusterRole (0.00s)
    --- FAIL: TestEnsureExternalDNSClusterRole/Does_not_exist (0.00s)
    --- FAIL: TestEnsureExternalDNSClusterRole/Exists_and_as_expected (0.00s)
    --- FAIL: TestEnsureExternalDNSClusterRole/Exists_and_needs_to_be_updated (0.00s)
=== RUN   TestEnsureExternalDNSClusterRoleBinding
=== RUN   TestEnsureExternalDNSClusterRoleBinding/Does_not_exist
2021-09-28T00:45:51.496+1000	INFO	created externalDNS cluster role binding	{"name": "external-dns-test"}
=== RUN   TestEnsureExternalDNSClusterRoleBinding/Exists_and_as_expected
=== RUN   TestEnsureExternalDNSClusterRoleBinding/Exists_and_needs_to_be_updated
2021-09-28T00:45:51.496+1000	INFO	updated externalDNS cluster role binding	{"name": "external-dns-test", "reason": "following fields changed: role-name,subject-name,subject-namespace"}
--- PASS: TestEnsureExternalDNSClusterRoleBinding (0.00s)
    --- PASS: TestEnsureExternalDNSClusterRoleBinding/Does_not_exist (0.00s)
    --- PASS: TestEnsureExternalDNSClusterRoleBinding/Exists_and_as_expected (0.00s)
    --- PASS: TestEnsureExternalDNSClusterRoleBinding/Exists_and_needs_to_be_updated (0.00s)
FAIL
FAIL	github.com/openshift/external-dns-operator/pkg/operator/controller/externaldns	0.012s
?   	github.com/openshift/external-dns-operator/pkg/operator/controller/externaldns/test	[no test files]
FAIL
[miheer@localhost external-dns-operator]$

alebedev87: Thanks, @miheer , your test was right. It's my previous code (before `go-cmp` change) which was not doing the right condition (`!` was missing before `DeepEqual` call).
That's why `TypeMeta` had to be added in this PR - because I not only migrated to `go-cmp` but also corrected the `if` condition.

PR #23, opened 2021-09-27
Set the domain filter flags based on Domain CR fields

arjunrn:
The change adds arguments to the external-dns container based on the values of the fields set in the External DNS custom resource.


lmzuccarelli:
a nit: the const "/etc/kubernetes" is declared 3 times, in the pod.go,  could we declare it once only (code smell issue refactoring).  Older commit I know.

lmzuccarelli:
Executed `make test` all tests passing. Also pod.go has about 97.3% code coverage - nice :)

pkg/operator/controller/externaldns/pod.go
```diff
@@ -93,19 +94,19 @@ func newExternalDNSContainerBuilder(image, provider, source string, secret *core
 }
 
 // build returns the definition of a single container for the given DNS zone with unique metrics port
-func (b *externalDNSContainerBuilder) build(zone string) *corev1.Container {
+func (b *externalDNSContainerBuilder) build(zone string) (*corev1.Container, error) {
```
alebedev87, line 97: I don't know how much it's useful to specify both `zone-id-filter` and `domain-filter` but our API and external-dns allow this, so should we I think.

However I think the other use case is more likely to be used for the domains: no `zone-id-filter`, only domain filters. And this use case needs to be covered too.

For the moment, the composition of the desired deployment is built around the zones (1 container per zone). I think that we need to add the special case when no zone is specified but there are domains.

pkg/operator/controller/externaldns/pod_test.go
```diff
@@ -0,0 +1,230 @@
...
+					ExternalDNSDomainUnion: v1alpha1.ExternalDNSDomainUnion{
+						MatchType: v1alpha1.DomainMatchTypeExact,
+						Name:      pointer.StringPtr("def.com"),
+					},
+					FilterType: v1alpha1.FilterTypeExclude,
+				},
+				{
+					ExternalDNSDomainUnion: v1alpha1.ExternalDNSDomainUnion{
+						MatchType: v1alpha1.DomainMatchTypeRegex,
+						Pattern:   pointer.StringPtr("*.ghi.com"),
```
alebedev87, line 189: This is a glob, not a regexp.

pkg/operator/controller/externaldns/pod.go
arjunrn: IIUC they should not interfere with each other. I think when there is no zone specified then `external-dns` will attempt to reconcile all potential resources. And will filter on the `domain-filter` values if specified.

pkg/operator/controller/externaldns/pod_test.go
arjunrn: Fixed.

pkg/operator/controller/externaldns/pod.go
alebedev87: Yes, they don't interfere. And yes, if only domain filter is specified external-dns will list all hosted zones and filter on the given domain.

The problem is that if no `.Spec.Zones` was set in CR the external-dns deployment will not have a single container.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -170,7 +171,58 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
...
+				}
+				args = append(args, fmt.Sprintf("--domain-filter=%s", *d.Name))
+			case operatorv1alpha1.DomainMatchTypeRegex:
+				if d.Pattern == nil {
+					return nil, fmt.Errorf("pattern for domain cannot be empty")
+				}
+				if regexDomainIncludeSet {
+					return nil, fmt.Errorf("more than one domain pattern cannot be specified for inclusion")
+				}
+				args = append(args, fmt.Sprintf("--regex-domain-filter=%s", *d.Pattern))
```
alebedev87, line 205: Just noticed that `--regexp-*` flags were added in version `0.8.0` ([link](https://github.com/kubernetes-sigs/external-dns/releases/tag/v0.8.0)) and we are still using `0.7.1` ([link](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/config/config.go#L21)).

I think it's fine to upgrade to `0.8.0` in our code but it would be good to test it at least manually.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -170,7 +171,58 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
...
+				if d.Name == nil {
+					return nil, fmt.Errorf("name for domain cannot be empty")
+				}
+				args = append(args, fmt.Sprintf("--domain-filter=%s", *d.Name))
+			case operatorv1alpha1.DomainMatchTypeRegex:
+				if d.Pattern == nil {
+					return nil, fmt.Errorf("pattern for domain cannot be empty")
+				}
+				if regexDomainIncludeSet {
+					return nil, fmt.Errorf("more than one domain pattern cannot be specified for inclusion")
```
alebedev87, line 203: Good catch! I think that we need to add this into the API comment.

pkg/operator/controller/externaldns/pod.go
arjunrn: I can write a card for that case. Let's address that in a separate PR.

pkg/operator/controller/externaldns/pod.go
arjunrn: Hmm, then I wonder why it was added to the CRD. I will update the external-dns version.

pkg/operator/controller/externaldns/pod.go
arjunrn: Added as part of doc.

pkg/operator/controller/externaldns/pod.go
arjunrn: Updated.

alebedev87: Yeah, I put the same version as in our downstream fork of external-dns but it seems that the API was updated with the regexp filters after the fork was created.

pkg/operator/controller/externaldns/pod.go
alebedev87: Xref the card: [CFE-142](https://issues.redhat.com/browse/CFE-142).

alebedev87:
/lgtm

Let's see what NetworkEdge guys think @Miciah , @miheer .

pkg/operator/controller/externaldns/pod.go
```diff
@@ -170,7 +171,58 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
...
+		args                  []string
+	)
+	for _, d := range b.externalDNS.Spec.Domains {
+		switch d.FilterType {
+		case operatorv1alpha1.FilterTypeInclude:
+			switch d.MatchType {
+			case operatorv1alpha1.DomainMatchTypeExact:
+				if d.Name == nil {
+					return nil, fmt.Errorf("name for domain cannot be empty")
+				}
```
Miciah, line 196: In the future, this validation could be handled by an admission webhook (which could also validate that `d.Name` *is* nil when `d.MatchType` is *not* `operatorv1alpha1.DomainMatchTypeExact`).  Likewise for the regex case.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -170,7 +171,58 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
...
+			}
+		case operatorv1alpha1.FilterTypeExclude:
+			switch d.MatchType {
+			case operatorv1alpha1.DomainMatchTypeExact:
+				args = append(args, fmt.Sprintf("--exclude-domains=%s", *d.Name))
+			case operatorv1alpha1.DomainMatchTypeRegex:
+				if regexDomainExcludeSet {
+					return nil, fmt.Errorf("more than one domain pattern cannot be specified for exclusion")
+				}
+				args = append(args, fmt.Sprintf("--regex-domain-exclusion=%s", *d.Pattern))
```
Miciah, line 218: Don't we need the same nil checks for `d.Name` and `d.Pattern` in exclusions that you have for inclusions?

Miciah (commented):
Need to fix the error handling in `buildSeq` and nil checks in `domainFilters`.  We can address my question about the API in a follow-up if you prefer.

api/v1alpha1/externaldns_types.go
```diff
@@ -113,6 +113,7 @@ type ExternalDNSDomain struct {
 	//
 	//  "Exclude": Exclude the domain set specified
 	//  by name or pattern.
+	// Only one "Pattern" filter types may be present for inclusion and exclusion.
```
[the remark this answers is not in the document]
Miciah: To elaborate on what I mean by combining patterns, take the "multiple regex include filter" test case for example; the input is this:
```go
			domainInput: []v1alpha1.ExternalDNSDomain{
				{
					ExternalDNSDomainUnion: v1alpha1.ExternalDNSDomainUnion{
						MatchType: v1alpha1.DomainMatchTypeRegex,
						Pattern:   pointer.StringPtr("(.*)\\.abc\\.com"),
					},
					FilterType: v1alpha1.FilterTypeInclude,
				},
				{
					ExternalDNSDomainUnion: v1alpha1.ExternalDNSDomainUnion{
						MatchType: v1alpha1.DomainMatchTypeRegex,
						Pattern:   pointer.StringPtr("(.*)\\.def\\.com"),
					},
					FilterType: v1alpha1.FilterTypeInclude,
				},
			},
```
Right now, this is expected to fail:
```go
			expectErr:     true,
			expectedError: "more than one domain pattern cannot be specified for inclusion",
```
We could just as easily take that input and produce the following output:
```go
			expectedArgs: []string{"--regex-domain-filter=(((.*)\\.abc\\.com)|((.*)\\.def\\.com))"},
```
That is, slap some parentheses around each pattern and join the parenthesized patterns using `|`.

BTW, you might want to use ``` `` ``` for regexps to conserve backslashes (e.g., ```Pattern:   pointer.StringPtr("(.*)\\.def\\.com"),``` → ```Pattern:   pointer.StringPtr(`(.*)\.def\.com`),```).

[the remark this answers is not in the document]
arjunrn: Done, I've also added validation of the regex's so that they can be joined with some degree of certainty that the resulting expression is also valid. Also switched to backticks.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -93,19 +94,19 @@ func newExternalDNSContainerBuilder(image, provider, source string, secret *core
...
 
 // buildSeq returns the definition of a single container for the given DNS zone
 // sequence param is used to create the unique metrics port
-func (b *externalDNSContainerBuilder) buildSeq(seq int, zone string) *corev1.Container {
+func (b *externalDNSContainerBuilder) buildSeq(seq int, zone string) (*corev1.Container, error) {
 	container := b.defaultContainer(controller.ExternalDNSContainerName(zone))
 	b.fillProviderAgnosticFields(seq, zone, container)
 	b.fillProviderSpecificFields(container)
-	return container
+	return container, nil
```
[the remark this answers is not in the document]
arjunrn: Fixed

pkg/operator/controller/externaldns/pod.go
arjunrn: I wanted to do most of the validation in a webhook but we were not sure if that would be ok with netedge.

pkg/operator/controller/externaldns/pod.go
arjunrn: Added.

api/v1alpha1/externaldns_types.go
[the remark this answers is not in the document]
Miciah: This comment: "Only one 'Pattern' filter types may be present for inclusion and exclusion." can be deleted now.

pkg/operator/controller/externaldns/pod_test.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+			if !tc.expectErr {
+				if !reflect.DeepEqual(args, tc.expectedArgs) {
+					t.Errorf("expected arguments %v, got %v", tc.expectedArgs, args)
+				}
+			} else {
+				if err == nil {
+					t.Errorf("expected error, got nil")
+					return
+				}
```
Miciah, line 257: This is redundant; line 246 already performs this check.

arjunrn: Fixed

api/v1alpha1/externaldns_types.go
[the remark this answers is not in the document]
arjunrn: Fixed

Miciah:
Thanks!
/lgtm

PR #29, opened 2021-09-29
Added usage docs for Infoblox and BlueCat providers

arjunrn:
Added a usage.md file which contains instructions on how an `ExternalDNS` resource can be created which are configured with BlueCat and Infoblox DNS providers.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -1,0 +1,109 @@
+# Usage
+
+### Credentials for DNS providers
+
+The _external-dns-operator_ manages external-dns deployments. It creates pods with correct credentials based on the
+provider configuration in the `ExternalDNS` resource. However, it does not provision the credentials themselves, instead
+it expects the credentials to be in the same namespace as the operator itself. It then copies over the credentials into
+the namespace where the _external-dns_ deployments are created so that they can be mounted by the pods.
+
+## Infoblox
+
+Before creating an `ExternalDNS` resource for the [Infoblox](https://www.infoblox.com/wp-content/uploads/infoblox-deployment-infoblox-rest-api.pdf)
+the following information is required:
+
+1. Grid Master Host
+2. WAPI version
+3. WAPI port
+4. WAPI username
+5. WAPI password
+
+Create a secret with the username and password as follows:
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: infoblox-credentials
+  namespace: #operator namespace
+data:
+  EXTERNAL_DNS_INFOBLOX_WAPI_USERNAME: # Base-64 encoded username
+  EXTERNAL_DNS_INFOBLOX_WAPI_PASSWORD: # Base-64 encoded password
+```
+
+Then create an `ExternalDNS` resource as follows:
+
+```yaml
+apiVersion: externaldns.olm.openshift.io/v1alpha1
+kind: ExternalDNS
+metadata:
+  name: infoblox-example
+spec:
+  provider:
+    type: Infoblox
+    infoblox:
+      credentials:
+        name: infoblox-credentials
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -0,0 +1,111 @@
...
+```yaml
+
+apiVersion: externaldns.olm.openshift.io/v1alpha1
+kind: ExternalDNS
+metadata:
+  name: bluecat-example
+spec:
+  provider:
+    type: BlueCat
+    infoblox:
```
lmzuccarelli, line 105: should this be blueCat (according to api) ?

arjunrn: Fixed

docs/usage.md
```diff
@@ -0,0 +1,111 @@
...
+  provider:
+    type: Infoblox
+    infoblox:
+      credentials:
+        name: infoblox-credentials
+      gridHost: # the grid master host from the previous step. eg: 172.26.1.200
+      wapiPort: # the WAPI port, eg: 80, 443, 8080
+      wapiVersion: # the WAPI version, eg: 2.11, 2.3.1
+  zones: # Replace with the desired hosted zones
+    - "test1.com"
```
alebedev87, line 51: `Spec.Zones` field is for zone IDs, so better to make them look like IDs rather than domain names.

docs/usage.md
```diff
@@ -0,0 +1,111 @@
...
+  "gatewayUsername": "user",
+  "gatewayPassword": "pass",
+  "dnsConfiguration": "Example",
+  "dnsView": "Internal",
+  "rootZone": "example.com",
+  "skipTLSVerify": false
+}
+```
+
+Then create a secret in the operator name with the command
```
alebedev87, line 84:
```suggestion
Then create a secret in the operator namespace with the command
```

docs/usage.md
alebedev87, line 51: `Spec.Zones` field is for zoneIDs, so it's better to make them look like IDs rather than domains.

arjunrn: In most cloud providers the zone id would be the domain or subdomain name itself. I'm not sure what it looks like for BlueCat.

alebedev87: Seems to be a number for BlueCat: [filtering](https://github.com/kubernetes-sigs/external-dns/blob/master/provider/bluecat/bluecat.go#L373),  [zone](https://github.com/kubernetes-sigs/external-dns/blob/master/provider/bluecat/bluecat.go#L94). However there is a [TODO](https://github.com/kubernetes-sigs/external-dns/blob/master/provider/bluecat/bluecat.go#L372) to change to the name :)

For infoblox, a [reference is used](https://github.com/kubernetes-sigs/external-dns/blob/master/provider/infoblox/infoblox.go#L280) which I fear is of that terrifying format:
```
Object Reference

Every WAPI Object is referenced by a unique Object Reference. WAPI returns this reference when an object is
created, modified, deleted or read. This reference is used to identify the object for the same operations.
An object reference is a string with the following format, without spaces:
wapitype / refdata [ : name1 [ { / nameN }... ] ]

Example
networkview/ZG5zLm5ldHdvcmtfdmlldyQw:default
```
I found the example of Infoblox API output for the zone creation:
```
$curl -k -u admin:infoblox -H 'content-type: application/json' -X POST "https://gridmaster/wapi/v2.11/zone_auth?_return_fields%2B=fqdn&_return_as_object=1" -d '{"fqdn": "info.com"}'

{"result":
 {"_ref": "zone_auth/ZG5zLnpvbmUkLl9kZWZhdWx0LmNvbS5pbmZv:info.com/default",
 "fqdn": "info.com",
 "view": "default"
}}
```

arjunrn: I've updated the examples.

PR #31, opened 2021-10-04
Added webhook for validating ExternalDNS CRs

arjunrn:
Used kubebuilder to add a validating webhook for `ExternalDNS` resources. The webhook will intercept all calls to create or update `ExternalDNS` resources and ensure that the values in the resources are specified correctly.
The change also contains envtests for testing the functionality of the webhook.


config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,25 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
+# More document can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
+apiVersion: cert-manager.io/v1
```
josefkarasek, line 4: Are we planning to use cert-manager or OLM?

arjunrn: We can specify a dependency on cert-manager operator in the OLM manifest.

josefkarasek: The reason I'm asking is OLM serves these certs. There's no reason to offload it to cert-manager.

All you need to do is include ValidatingWebhookConfiguration in kustomize. operator-sdk/OLM will pick it up. Here's an [example](https://github.com/josefkarasek/projectrequestlimits/blob/main/bundle/manifests/projectrequestlimits.clusterserviceversion.yaml#L99) (generated by operator-sdk)

Makefile
```diff
@@ -58,12 +58,11 @@ fmt: ## Run go fmt against code.
...
-test: manifests generate fmt vet ## Run unit tests
-		go test \
-		-race \
-		-covermode=atomic \
-		-coverprofile coverage.out \
-		./...
+ENVTEST_ASSETS_DIR ?= $(shell pwd)/testbin
+test: manifests generate fmt vet ## Run tests.
+	mkdir -p ${ENVTEST_ASSETS_DIR}
+	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.8.3/hack/setup-envtest.sh
```
lmzuccarelli, line 64: I noticed k8s_version default is set 1.19.2 - should we set this variable (i.e to later version 1.21.1 used in version 4.8.12 of openshift) ?

config/certmanager/certificate.yaml
arjunrn: Cool, didn't know OLM can do that. Will modify the manifests.

Makefile
arjunrn: Where do you see the version being hardcoded? Do you mean the `controller-runtime` version used here?

lmzuccarelli: in the script setup-envtest.sh  -> k8s_version="${ENVTEST_K8S_VERSION:-1.19.2}" - that's the default. I was asking if we need to set the envar ENVTEST_K8S_VERSION

arjunrn: Looks like there is a [tool](https://github.com/kubernetes-sigs/controller-runtime/tree/master/tools/setup-envtest) for managing the binaries for the envtests. I've switched the make target and set the version.

api/v1alpha1/externaldns_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+//+kubebuilder:webhook:path=/validate-externaldns-olm-openshift-io-v1alpha1-externaldns,mutating=false,failurePolicy=fail,sideEffects=None,groups=externaldns.olm.openshift.io,resources=externaldns,verbs=create;update,versions=v1alpha1,name=vexternaldns.kb.io,admissionReviewVersions={v1,v1beta1}
+
+var _ webhook.Validator = &ExternalDNS{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *ExternalDNS) ValidateCreate() error {
+	webhookLog.Info("validate create", "name", r.Name)
+	return r.validateFilters()
```
alebedev87, line 46: Can we add more validations?
- `FQDNTemplate` has to be there if HostnameAnnotationPolicy is `Ignore`
- Credentials/configs have to be there for each provider

bundle/manifests/external-dns-operator_clusterserviceversion.yaml
```diff
@@ -88,6 +88,9 @@ spec:
                       fieldPath: metadata.namespace
                 image: quay.io/openshift/origin-external-dns-operator:latest
                 name: operator
+                ports:
+                  - containerPort: 9443
+                    name: https
```
alebedev87, line 93:
```suggestion
                    name: webhook-server
```

config/default/manager_webhook_patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: external-dns-operator
+  namespace: external-dns-operator
+spec:
+  template:
+    spec:
+      containers:
+      - name: external-dns-operator
+        ports:
+        - containerPort: 9443
```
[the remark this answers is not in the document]
arjunrn: It's setup later
https://github.com/openshift/external-dns-operator/blob/caea16a2fdf02b0921fd607f4ac59c64615b9a9d/pkg/operator/operator.go#L96-L100

config/webhook/service.yaml
```diff
@@ -0,0 +1,12 @@
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-service
+  namespace: system
```
alebedev87, line 6:
```suggestion
  namespace: external-dns-operator
```

config/default/manager_webhook_patch.yaml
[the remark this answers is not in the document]
alebedev87: Ok, I see, `9443` is the default port ([link](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/server.go#L42)).

bundle/manifests/external-dns-operator_clusterserviceversion.yaml
```diff
@@ -119,3 +122,23 @@ spec:
...
+          apiVersions:
+            - v1alpha1
+          operations:
+            - CREATE
+            - UPDATE
+          resources:
+            - externaldnses
+      sideEffects: None
+      type: ValidatingAdmissionWebhook
+      webhookPath: /validate-externaldns-olm-openshift-io-v1alpha1-externaldns
```
alebedev87, line 144: This path is generated from the comment marker but still, wouldn't it be more approachable to make it shorter?
If you would like to test the webhook server manually, for instance (rare but still possible case).

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: validating-webhook-configuration
+webhooks:
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
```
alebedev87, line 12: We need to think of how to guide non-OLM users to add `caBundle` here, a kustomize patch?
BTW I hope that CA bundle is managed automagically by OLM.

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,29 @@
...
+    service:
+      name: webhook-service
+      namespace: system
```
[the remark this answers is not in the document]
arjunrn: This cannot be changed. However that's not a problem because the namespace is actually injected by kustomize using the configuration:
https://github.com/openshift/external-dns-operator/blob/main/config/crd/kustomizeconfig.yaml

bundle/manifests/external-dns-operator_clusterserviceversion.yaml
```diff
@@ -119,3 +122,23 @@ spec:
   provider:
     name: Red Hat, Inc.
   version: 0.0.1
+  webhookdefinitions:
+    - admissionReviewVersions:
+        - v1
+      deploymentName: external-dns-operator
+      failurePolicy: Fail
+      generateName: vexternaldns.kb.io
```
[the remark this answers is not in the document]
arjunrn: v=validate
externaldns=ExternalDNS resource
kb=kubebuilder
It's how it was generated. We can change it.

bundle/manifests/external-dns-operator_clusterserviceversion.yaml
arjunrn: Unfortunately, which registering the webhook in the operator the path is generated:

https://github.com/kubernetes-sigs/controller-runtime/blob/e1880f5026f6e2d8968b64c05cb66b91bce383ec/pkg/builder/webhook.go#L141-L149

And this cannot be overriden easily unless we don't use the helper functions.

config/webhook/manifests.yaml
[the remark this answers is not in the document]
alebedev87: Didn't quite get how CRD's kustomizeconfig changes the clientConfig in ValidatingWebHook.
The namespace in the clientConfig of the webhook seems to come from [this configuration](https://github.com/openshift/external-dns-operator/blob/666d26cc7b7a04cf565068bf23fa4ef4541ae271/config/default/kustomization.yaml#L2) somehow.

Service namespace is `andrey` but webhook's client config is still `system` because no global namespace is given:
```bash
$ git diff -U0
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -2 +2 @@
-namespace: external-dns-operator
+#namespace: external-dns-operator

--- a/config/webhook/service.yaml
+++ b/config/webhook/service.yaml
@@ -6 +6 @@ metadata:
-  namespace: external-dns-operator
+  namespace: andrey

$ oc apply -k config/default/ --dry-run=client -o yaml | grep -A4 clientConfig:
    clientConfig:
      service:
        name: webhook-service
        namespace: system
        path: /validate-externaldns-olm-openshift-io-v1alpha1-externaldns
```

But when the global namespace is there, it's used:
```bash
$ git diff -U0
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -2 +2 @@
-namespace: external-dns-operator
+namespace: external-dns-operator-andrey

$ $ oc apply -k config/default/ --dry-run=client -o yaml | grep -A4 clientConfig:
    clientConfig:
      service:
        name: webhook-service
        namespace: external-dns-operator-andrey
        path: /validate-externaldns-olm-openshift-io-v1alpha1-externaldns
```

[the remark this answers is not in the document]
alebedev87: Oouf, I think I got it. It's so called custom transformer which can modify the meaning of the transformation.
In case of the webhook's kustomize directory the custom transformer was used to "redirect" the namespace definition to a different place, not the standard `metadata.namespace` but [here](https://github.com/openshift/external-dns-operator/blob/666d26cc7b7a04cf565068bf23fa4ef4541ae271/config/webhook/kustomizeconfig.yaml#L21). So, the namespace specified in `config/default` was added to the custom path for the validating webhook.

Anyway, you are right, we can keep it set to `system`.

config/default/manager_webhook_patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+          protocol: TCP
+        volumeMounts:
+        - mountPath: /tmp/k8s-webhook-server/serving-certs
+          name: cert
+          readOnly: true
+      volumes:
+      - name: cert
+        secret:
+          defaultMode: 420
+          secretName: webhook-server-cert
```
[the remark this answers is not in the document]
arjunrn: I've added a convenience script to generate the certifcates and patch the webhook for local deployments.

config/webhook/manifests.yaml
arjunrn: >  BTW I hope that CA bundle is managed automagically by OLM.
It [does](https://github.com/openshift/external-dns-operator/pull/31#discussion_r722233346)

config/webhook/kustomizeconfig.yaml
```diff
@@ -0,0 +1,25 @@
+# the following config is for teaching kustomize where to look at when substituting vars.
+# It requires kustomize v2.1.0 or newer to work properly.
+nameReference:
+- kind: Service
+  version: v1
+  fieldSpecs:
+  - kind: MutatingWebhookConfiguration
+    group: admissionregistration.k8s.io
+    path: webhooks/clientConfig/service/name
```
Miciah, line 9: Are these stanzas for a mutating webhook needed even though we only actually have a validating webhook?

Makefile
```diff
@@ -70,12 +72,10 @@ fmt: ## Run go fmt against code.
...
 
-test: manifests generate fmt vet ## Run unit tests
-		go test \
-		-race \
-		-covermode=atomic \
-		-coverprofile coverage.out \
-		./...
+ENVTEST_ASSETS_DIR ?= $(shell pwd)/testbin
+test: manifests generate fmt vet ## Run tests.
+	mkdir -p $(ENVTEST_ASSETS_DIR)
```
Miciah, line 77:
```suggestion
	mkdir -p "$(ENVTEST_ASSETS_DIR)"
```

Miciah (commented):
/lgtm
/hold
Just some minor suggestions and requests for clarification.  Feel free to remove the hold and address any real issues in follow-up PRs.

api/v1alpha1/externaldns_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+			if f.Name == nil || *f.Name == "" {
+				return errors.New(`"Name" cannot be empty when match type is "Exact"`)
+			}
+		case DomainMatchTypeRegex:
+			if f.Pattern == nil || *f.Pattern == "" {
+				return errors.New(`"Pattern" cannot be empty when match type is "Pattern"`)
+			}
+			_, err := regexp.Compile(*f.Pattern)
+			if err != nil {
+				return fmt.Errorf(`invalid pattern for "Pattern" match type: %w`, err)
```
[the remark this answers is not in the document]
Miciah: If the constituent regexps are valid, then the concatenated pattern should be valid, and unit tests already cover the concatenation logic, so I think validating the concatenated pattern in the webhook would be superfluous.

hack/generate-certs.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Taken from https://github.com/newrelic/k8s-webhook-cert-manager/blob/master/generate_certificate.sh
+# and modified for use with newer versions of Kubernetes/Openshift.
+
+set -euo pipefail
+
+usage() {
+  cat <<EOF
+Generate certificate suitable for use with any Kubernetes Mutating Webhook.
+This script uses k8s' CertificateSigningRequest API to a generate a
```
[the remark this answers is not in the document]
arjunrn: My bad. The script I copied this from did indeed use CSR resources. But this approach doesn't work with newer versions of Kubernetes. So I modified the script to generate the certificate itself, but forgot to update the help line.

README.md
```diff
@@ -22,6 +22,16 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
...
        *Note*: For other providers, see options listed in `api/v1alpha1/externaldns_types.go`.
 
+4. The previous step deploys the validation webhook which requires TLS authentication for the webhook server. The
+   manifests do not contain a valid certificate and key and these should be provisioned in the cluster through other tools.
+   For local testing however the script `hack/generate-cert.sh` can be used to generate the certificate bundle and patch
+   the validation webhook config. Run:
+   ```bash
+   hack/generate_certificate.sh --service webhook-service --webhook validating-webhook-configuration \
+   --secret webhook-server-cert --namespace external-dns-operator
+   ```
```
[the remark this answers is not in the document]
arjunrn: I've updated the steps to highlight that this helper script should not be used in production clusters.

config/webhook/kustomizeconfig.yaml
arjunrn: These were generated by kubebuilder.

PR #36, opened 2021-10-07
Allow deployments with no zones

josefkarasek:


pkg/operator/controller/externaldns/deployment.go
```diff
@@ -176,14 +176,21 @@ func desiredExternalDNSDeployment(namespace, image, secretName string, serviceAc
 	depl.Spec.Template.Spec.Volumes = append(depl.Spec.Template.Spec.Volumes, volumes...)
 
 	cbld := newExternalDNSContainerBuilder(image, provider, source, secretName, volumes, externalDNS)
-	for _, zone := range externalDNS.Spec.Zones {
-		container, err := cbld.build(zone)
+	if len(externalDNS.Spec.Zones) == 0 {
+		container, err := cbld.build("")
```
alebedev87, line 180: Can you please add the empty string zone case into [ExternalDNSContainerName's test](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/names_test.go#L24)?

PR #37, opened 2021-10-07
Implement externalDNS.Status update

sherine-k:
Update externalDNS custom resource status from Reconcile
Adds:
- Conditions: "PodsScheduled","DeploymentAvailable","DeploymentReplicasMinAvailable", "DeploymentReplicasAllAvailable"
- zones
- observedGeneration

### DIFF
--- a/pkg/operator/controller/externaldns/status.go
+++ b/pkg/operator/controller/externaldns/status.go
@@ -197,7 +197,7 @@ func computeDeploymentPodsScheduledCondition(ctx context.Context, cl client.Clie
 	}
 	if len(unscheduled) != 0 {
 		var haveUnschedulable bool
-		message := "Some pods are not scheduled: "
+		message := "Some pods are not scheduled:"
 		// Sort keys so that the result is deterministic.
 		keys := make([]*corev1.Pod, 0, len(unscheduled))
 		for pod := range unscheduled {
@@ -213,9 +213,9 @@ func computeDeploymentPodsScheduledCondition(ctx context.Context, cl client.Clie
 			cond := unscheduled[pod]
 			if cond.Reason == corev1.PodReasonUnschedulable {
 				haveUnschedulable = true
-				message += fmt.Sprintf("Pod %q cannot be scheduled: %s", pod.Name, cond.Message)
+				message += fmt.Sprintf(" Pod %q cannot be scheduled: %s", pod.Name, cond.Message)
 			} else {
-				message += fmt.Sprintf("Pod %q is not yet scheduled: %s: %s", pod.Name, cond.Reason, cond.Message)
+				message += fmt.Sprintf(" Pod %q is not yet scheduled: %s: %s", pod.Name, cond.Reason, cond.Message)
 			}
 		}
 		if haveUnschedulable {


pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,668 @@
+package externaldnscontroller
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
```
alebedev87, line 7: I didn't see much of added value of this package. It's used to compare strings and lengths of slices and doing the error logging.

I would avoid adding another test package and would suggest to fallback to the standard comparisons and `testing.T.Errorf()`. Because `gomega` and `ginkgo` are coming with Arjun's webhook PR which will increase the number of the testing approaches to 3: go native, testify, gingko.

I think that `ginkgo` will dominate this repository at some moment but go native way is also a possible option in some cases.

pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,668 @@
...
+			existingDeployment: appsv1.Deployment{
+				Spec: appsv1.DeploymentSpec{
+					Selector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{
+							"name": "external-dns-operator",
+						},
+					},
+				},
+			},
+			existingPods: []corev1.Pod{
```
alebedev87, line 202: Can you use a function similar to `fakeDeployment` for PODs too?
This would save a lot of space and make test more readable.

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,285 @@
...
+	if err != nil {
+		return metav1.Condition{
+			Type:    ExternalDNSDeploymentReplicasMinAvailableConditionType,
+			Status:  metav1.ConditionUnknown,
+			Reason:  "InvalidMaxUnavailableValue",
+			Message: fmt.Sprintf("invalid value for max unavailable: %v", err),
+		}
+	}
+	if maxSurge == 0 && maxUnavailable == 0 {
+		//TODO understand why this is the default
```
alebedev87, line 98: This is actually a blocking situation when you have both `maxSurge` and `maxUnavailable` set to 0. `maxSurge=0` says "do the in-place upgrade" (don't spawn new PODs, replace existing ones) while `maxUnavailable=0` says "and please don't make any POD unavailable". I have to try but I think that the rolling update with these values will never pass and will timeout.

So, some minimal number was still added.

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,285 @@
...
+	err := client.Status().Update(ctx, externalDNS)
+	return err
+}
+
+func getPodsList(client client.Client, ctx context.Context) ([]corev1.Pod, error) {
+	pods := &corev1.PodList{}
+	if err := client.List(ctx, pods); err != nil {
+		return nil, err
+	}
+	return pods.DeepCopy().Items, error(nil)
```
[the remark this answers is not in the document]
sherine-k: Done. to be checked and resolved

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,285 @@
...
+	if deploymentExists {
+
+		externalDNS.Status.Conditions = MergeConditions(externalDNS.Status.Conditions, computeDeploymentAvailableCondition(currentDeployment))
+		externalDNS.Status.Conditions = MergeConditions(externalDNS.Status.Conditions, computeMinReplicasCondition(currentDeployment))
+		externalDNS.Status.Conditions = MergeConditions(externalDNS.Status.Conditions, computeAllReplicasCondition(currentDeployment))
+
+		pods, errgpl := getPodsList(client, ctx)
+		if errgpl != nil {
+			return errgpl
+			// question: we dont update what we have so far though?
```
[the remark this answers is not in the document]
sherine-k: Done. to be checked and resolved

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,285 @@
...
+	externalDNS.Status.ObservedGeneration = externalDNS.Generation
+	externalDNS.Status.Zones = externalDNS.Spec.DeepCopy().Zones
+
+	err := client.Status().Update(ctx, externalDNS)
+	return err
+}
+
+func getPodsList(client client.Client, ctx context.Context) ([]corev1.Pod, error) {
+	pods := &corev1.PodList{}
+	if err := client.List(ctx, pods); err != nil {
```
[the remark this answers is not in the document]
sherine-k: Done. to be checked and resolved

pkg/operator/controller/externaldns/status_test.go
sherine-k: Done. to be checked and resolved

pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,668 @@
...
+			existingObjects: fakePodListAsRuntimeObject(),
+			expectedResult:  fakePodList(),
+			errExpected:     false,
+		},
+	}
+	for _, tc := range testCases {
+		//Question: is it better to use envtest.Environment instead
+		cl := fake.NewClientBuilder().WithScheme(test.Scheme).WithRuntimeObjects(tc.existingObjects...).Build()
+
+		podList, err := getPodsList(cl, context.TODO())
```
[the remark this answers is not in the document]
sherine-k: updated for getFilteredPodsList (replaces getPodsList). To be checked and resolved

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+		Type:    ExternalDNSDeploymentReplicasMinAvailableConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  "DeploymentMinimumReplicasMet",
+		Message: "Minimum replicas requirement is met",
+	}
+}
+
+// computeAllReplicasCondition returns an externalDNS condition based on the deployment, its number of desired pods
+// and the number of available replicas.
+// the condition is true if the number of available replicas is greater than number of desired replicas
```
alebedev87, line 119:
```suggestion
// the condition is true if the number of available replicas is same or greater than number of desired replicas
```

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+		Type:    ExternalDNSDeploymentReplicasAllAvailableConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  "DeploymentReplicasAvailable",
+		Message: "All replicas are available",
+	}
+}
+
+// computeDeploymentPodsScheduledCondition lists the pods matching the namespace and the label selector of the deployment
+// returns condition true when all matching pods are scheduled
+// returns condition false if no matching pods were found, or if any of the pods are unscheduled
```
alebedev87, line 145:
```suggestion
// returns condition false if no matching pods were found, or if any of the pods is unscheduled
```

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+			Type:    ExternalDNSPodsScheduledConditionType,
+			Status:  metav1.ConditionTrue,
+			Reason:  "AllPodsScheduled",
+			Message: "All pods are scheduled",
+		}
+	}
+
+}
+
+// MergeConditions updates the conditions list with a new condition
```
alebedev87, line 218:
```suggestion
// mergeConditions updates the conditions list with new conditions
```

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+		LastTransitionTime: metav1.NewTime(clock.Now()),
+	}
+}
+
+// updateExternalDNSStatus recomputes all conditions given the current deployment and its status
+// and pushes the new externalDNS custom resource with updated status through a call to the client.Update function
+func (r *reconciler) updateExternalDNSStatus(ctx context.Context, externalDNS *operatorv1alpha1.ExternalDNS, currentDeployment *appsv1.Deployment) error {
+	extDNSWithStatus := externalDNS.DeepCopy()
+	if currentDeployment != nil {
+
```
alebedev87, line 262: unnecessary blank line

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+		extDNSWithStatus.Status.Conditions = mergeConditions(extDNSWithStatus.Status.Conditions, computeDeploymentAvailableCondition(currentDeployment))
+		extDNSWithStatus.Status.Conditions = mergeConditions(extDNSWithStatus.Status.Conditions, computeMinReplicasCondition(currentDeployment))
+		extDNSWithStatus.Status.Conditions = mergeConditions(extDNSWithStatus.Status.Conditions, computeAllReplicasCondition(currentDeployment))
+
```
alebedev87, line 266: blank line

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func createPodsScheduledUnknownCondition(reason, message string) metav1.Condition {
+	return metav1.Condition{
+		Type:    ExternalDNSPodsScheduledConditionType,
+		Status:  metav1.ConditionUnknown,
+		Reason:  reason,
+		Message: message,
+	}
+
```
alebedev87, line 284: blank line

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+		Status:             metav1.ConditionUnknown,
+		Reason:             "DeploymentAvailabilityUnknown",
+		Message:            "The deployment has no Available status condition set",
+		LastTransitionTime: metav1.NewTime(clock.Now()),
+	}
+}
+
+// updateExternalDNSStatus recomputes all conditions given the current deployment and its status
+// and pushes the new externalDNS custom resource with updated status through a call to the client.Update function
+func (r *reconciler) updateExternalDNSStatus(ctx context.Context, externalDNS *operatorv1alpha1.ExternalDNS, currentDeployment *appsv1.Deployment) error {
```
alebedev87, line 259: Can you please move this method to the top of the file?
Would be a little more readable as it's the main function of the file which calls all the others.

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+		extDNSWithStatus.Status.Conditions = mergeConditions(extDNSWithStatus.Status.Conditions, computeAllReplicasCondition(currentDeployment))
+
+		extDNSWithStatus.Status.Conditions = mergeConditions(extDNSWithStatus.Status.Conditions, computeDeploymentPodsScheduledCondition(ctx, r.client, currentDeployment))
+	} else {
+		extDNSWithStatus.Status.Conditions = mergeConditions(extDNSWithStatus.Status.Conditions, createDeploymentAvailabilityUnknownCondition())
+	}
+	extDNSWithStatus.Status.ObservedGeneration = extDNSWithStatus.Generation
+	extDNSWithStatus.Status.Zones = extDNSWithStatus.Spec.Zones
+
+	return r.client.Status().Update(ctx, extDNSWithStatus)
```
alebedev87, line 274: Would be good to add a check of the changed status similar to what cluster-ingress-operator does [here](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/status.go#L71).

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,296 @@
...
+// returns condition false if no matching pods were found, or if any of the pods are unscheduled
+func computeDeploymentPodsScheduledCondition(ctx context.Context, cl client.Client, deployment *appsv1.Deployment) metav1.Condition {
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil || selector.Empty() {
+		return createPodsScheduledUnknownCondition("InvalidLabelSelector", "Deployment has an invalid label selector.")
+	}
+	pods, err := getFilteredPodsList(ctx, cl, deployment.Namespace, selector)
+	if err != nil {
+		return createPodsScheduledUnknownCondition("PodScheduledUnknown", "unable to list pods")
+	} else {
```
alebedev87, line 154: More go idiomatic way of doing this would be to remove `else` as previous `if` would return if there is an error.

pkg/operator/controller/externaldns/status.go
sherine-k: removed

pkg/operator/controller/externaldns/status.go
sherine-k: ok

pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,681 @@
...
+}
+
+func TestComputeDeploymentPodsScheduledCondition(t *testing.T) {
+	testCases := []struct {
+		name               string
+		existingDeployment appsv1.Deployment
+		existingPods       []corev1.Pod
+		expectedResult     metav1.Condition
+	}{
+		//all podScheduledConditions are true
```
Miciah, line 220: These comments look redundant; the information is already conveyed in each test case's `name` field.

pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,681 @@
...
+			},
+		}
+		labelSelector, err := metav1.LabelSelectorAsSelector(aSelector)
+		if err != nil {
+			t.Error("Unable to create labelSelector for label external-dns-operator")
+		}
+		podList, err := getFilteredPodsList(context.TODO(), cl, "external-dns-operator", labelSelector)
+		if tc.errExpected && err == nil {
+			t.Error("expected an error but got none")
+		} else {
```
Miciah, line 441:
```suggestion
		} else if !tc.errExpected {
```

pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,681 @@
...
+			if err != nil {
+				t.Errorf("expected no error but got %v", err)
+			}
+			if len(podList) != 1 {
+				t.Errorf("expected podList to contain 2 pods but got %v", len(podList))
```
Miciah, line 446:
```suggestion
			if len(podList) != len(tc.expectedResult) {
				t.Errorf("expected podList to contain %d pods but got %d", len(tc.expectedResult), len(podList))
```

pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,681 @@
...
+			client: cl,
+			scheme: test.Scheme,
+			config: testConfig(),
+			log:    zap.New(zap.UseDevMode(true)),
+		}
+
+		err := r.updateExternalDNSStatus(context.TODO(), tc.existingExtDNS, &tc.existingDeployment)
+		if tc.errExpected && err == nil {
+			t.Error("expected an error but got none")
+		} else {
```
Miciah, line 503:
```suggestion
		} else if !tc.errExpected {
```

Miciah (commented):
Looks good overall.  I've commented with some minor suggestions and noted a few logic errors in the test code.

pkg/operator/controller/externaldns/status.go
```diff
@@ -0,0 +1,320 @@
...
+// returns condition true when all matching pods are scheduled
+// returns condition false if no matching pods were found, or if any of the pods is unscheduled
+func computeDeploymentPodsScheduledCondition(ctx context.Context, cl client.Client, deployment *appsv1.Deployment) metav1.Condition {
+	selector, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil || selector.Empty() {
+		return createPodsScheduledUnknownCondition("InvalidLabelSelector", "Deployment has an invalid label selector.")
+	}
+	pods, err := getFilteredPodsList(ctx, cl, deployment.Namespace, selector)
+	if err != nil {
+		return createPodsScheduledUnknownCondition("PodScheduledUnknown", "unable to list pods")
```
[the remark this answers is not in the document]
sherine-k: Picked the first choice - keeping the error in the condition message

pkg/operator/controller/externaldns/status_test.go
```diff
@@ -0,0 +1,634 @@
...
+						isConditionTypeFound = true
+						if diff := cmp.Diff(expectedCondition, cond, ignoreTimeOpt); diff != "" {
+							t.Errorf("expected condition %v; got condition %v: \n %s", tc.expectedResult, cond, diff)
+						}
+					}
+				}
+				if !isConditionTypeFound {
+					t.Errorf("expected condition type %v was not found in the result", expectedCondition.Type)
+				}
+			}
```
alebedev87, line 363: All this can be replaced by a single Diff:
```go
	conditionCmpOpts := []cmp.Option{
		cmpopts.EquateEmpty(),
		ignoreTimeOpt,
		cmpopts.SortSlices(func(a, b metav1.Condition) bool { return a.Type < b.Type }),
	}
	if diff := cmp.Diff(expectedCondition, cond, conditionCmpOpts...); diff != "" {
	}
```

PR #45, opened 2021-10-20
If the platform is OCP then we fetch the credentials from the secret we requested from cloud credentials operator.

miheer:
If the source is openshift-route then we get the credentials from the secret we requested from cloud credentials operator.

This fixes https://issues.redhat.com/browse/CFE-135


miheer:
@Miciah PTAL

pkg/operator/controller/credentials-secret/controller.go
```diff
@@ -200,9 +282,13 @@ func isInNS(namespace string) func(o client.Object) bool {
...
 
 // getExternalDNSCredentialsSecretName returns the name of the credentials secret retrieved from externalDNS resource
-func getExternalDNSCredentialsSecretName(externalDNS *operatorv1alpha1.ExternalDNS) string {
+func (r *reconciler) getExternalDNSCredentialsSecretName(externalDNS *operatorv1alpha1.ExternalDNS) string {
+	if ocpVersion := getOCPMajorVersion(r.kubeclient); ocpVersion == ocpMajorVersion4 {
+		return secretFromCloudCredentialsOperator
+	}
 	switch externalDNS.Spec.Provider.Type {
 	case operatorv1alpha1.ProviderTypeAWS:
+		fmt.Printf("In switch")
```
arjunrn, line 291: Debug statement should be removed.

pkg/operator/controller/credentials-secret/controller_test.go
```diff
@@ -124,14 +160,23 @@ func TestReconcile(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+
 		t.Run(tc.name, func(t *testing.T) {
 			cl := fake.NewClientBuilder().WithScheme(testScheme).WithRuntimeObjects(tc.existingObjects...).Build()
+			kubeFakeClient := kubernetesFakeClient.NewSimpleClientset()
+
+			if tc.name == "Bootstrap OCP" {
```
arjunrn, line 168: Tests should not have the same name. Instead there could be a bool flag indicating if the fake resources should be loaded.

miheer:
@arjunrn  PTAL

pkg/operator/controller/credentials-secret/controller_test.go
miheer: @arjunrn  done

pkg/operator/controller/credentials-secret/controller.go
miheer: @arjunrn done

dgoodwin:
@miheer we think this may have caused https://bugzilla.redhat.com/show_bug.cgi?id=2019001

possible we will have to open a revert PR until this can be resolved as it's blocking some things which are blocking some important things. :)

PR #47, opened 2021-10-26
CFE-118: Add E2E test suite

rfredette:
Currently includes 2 tests:

- `TestOperatorAvailable`: Verify that the external dns operator is running

- `TestExternalDNSRecordCreate`: Create an external-dns instance, plus a service with a backend pod. Verify that DNS records are created for the service, and that the service is reachable at the appropriate domain name.
  - ~~This test currently fails at verifying that the service is reachable.~~ This failure seems to be that something my local DNS config was caching NXDOMAIN responses for longer than the test would wait. I was able to get my local config to be more responsive to domain name changes, but also I increased the timeout when querying the service to allow for some variance.

In addition to the current issue in `TestExternalDNSRecordCreate`, there are a couple other loose ends before I'd remove the WIP label:
- ~~Move utility functions to a separate file, such as `test/e2e/util.go`~~ Done
- ~~If the external-dns instance is deleted before it can clean up all records in the hosted zone, hosted zone cleanup also fails. `deleteHostedZone` should remote all non-essential records in the hosted zone before attempting to delete the zone itself.~~ Done


alebedev87:
@rfredette would be nice to add a new Prow job which runs `make tets-e2e` to see the e2e in action.

test/e2e/operator_test.go
```diff
@@ -0,0 +1,290 @@
...
+		os.Exit(1)
+	}
+
+	defaultParentDomainName = os.Getenv("EXTDNS_PARENT_DOMAIN")
+	if defaultParentDomainName == "" {
+		fmt.Printf("No parent domain specified. Please set the environment variable EXTDNS_PARENT_DOMAIN\n")
+		os.Exit(1)
+	}
+	defaultParentZoneID = os.Getenv("EXTDNS_PARENT_ZONEID")
+	if defaultParentDomainName == "" {
```
alebedev87, line 142:
```suggestion
	if defaultParentZoneID == "" {
```

test/e2e/operator_test.go
```diff
@@ -0,0 +1,290 @@
...
+		os.Exit(1)
+	}
+	kclient = kubeClient
+
+	if err := kclient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &infraConfig); err != nil {
+		fmt.Printf("failed to get infrastructure config: %v\n", err)
+		os.Exit(1)
+	}
+
+	defaultParentDomainName = os.Getenv("EXTDNS_PARENT_DOMAIN")
```
alebedev87, line 136: where `EXTDNS_*` variables are set?

test/e2e/operator_test.go
```diff
@@ -0,0 +1,290 @@
...
+		fmt.Printf("Unsupported Provider")
+		os.Exit(1)
+	}
+
+	dialer = &net.Dialer{
+		Resolver: &net.Resolver{
+			PreferGo: true,
+			Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+				d := net.Dialer{
+					Timeout: 5000 * time.Millisecond,
```
alebedev87, line 168: would be nice to have this as a variable for easier configuration.

test/e2e/operator_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	os.Exit(m.Run())
+}
+
+func TestOperatorAvailable(t *testing.T) {
+	expected := []appsv1.DeploymentCondition{
+		{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionTrue},
+	}
+	if err := waitForOperatorDeploymentStatusCondition(t, kclient, expected...); err != nil {
```
alebedev87, line 189: I don't quite get where the operator's deployment is created.

test/e2e/operator_test.go
```diff
@@ -0,0 +1,290 @@
...
+	domainName = generateDomainName()
+
+	platform := infraConfig.Status.PlatformStatus
+	if platform == nil {
+		fmt.Printf("platform status is missing for infrastructure %s", infraConfig.Name)
+		os.Exit(1)
+	}
+	switch platform.Type {
+	case configv1.AWSPlatformType:
+		mySession := awsSession.Must(awsSession.NewSession())
```
alebedev87, line 156: I'm wondering whether this thing needs `AWS_ACCESS_*/AWS_SECRET_*` variables set beforehand.

test/e2e/operator_test.go
```diff
@@ -0,0 +1,290 @@
+//+build e2e
```
alebedev87, line 1: These tests seem to be quite OpenShift oriented (infrastructure config, aws credentials secret). Maybe it makes sense to tag this as openshift e2e.

test/e2e/util.go
```diff
@@ -0,0 +1,448 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/aws/aws-sdk-go/service/route53"
+)
+
+var defaultExtDNSTemplate = operatorv1alpha1.ExternalDNS{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "test-extdns",
+		Namespace: "publish-external-dns",
```
alebedev87, line 31: ExternalDNS custom resource is a cluster level object. No need to specify a namespace.

test/e2e/util.go
```diff
@@ -0,0 +1,448 @@
...
+		Source: operatorv1alpha1.ExternalDNSSource{
+			ExternalDNSSourceUnion: operatorv1alpha1.ExternalDNSSourceUnion{
+				Type: operatorv1alpha1.SourceTypeService,
+				Service: &operatorv1alpha1.ExternalDNSServiceSourceOptions{
+					ServiceType: []corev1.ServiceType{
+						corev1.ServiceTypeLoadBalancer,
+						corev1.ServiceTypeClusterIP,
+					},
+				},
+				Namespace: StringRef("publish-external-dns"),
```
alebedev87, line 43: We thought of removing the namespace filter from the API. Better to remove it, this is not going to change the rest of the tests.

test/e2e/operator_test.go
```diff
@@ -0,0 +1,290 @@
...
+	if err := kclient.Create(context.TODO(), &defaultBackend); err != nil {
+		t.Fatalf("Failed to create default backend: %v", err)
+	}
+	defer kclient.Delete(context.TODO(), &defaultBackend)
+	if err := kclient.Create(context.TODO(), &defaultService); err != nil {
+		t.Fatalf("Failed to create default service: %v", err)
+	}
+	defer kclient.Delete(context.TODO(), &defaultService)
+	expectedDomainName := fmt.Sprintf("%s.%s", defaultService.ObjectMeta.Name, domainName)
+	err = wait.PollImmediate(1*time.Second, 1*time.Minute, func() (bool, error) {
```
alebedev87, line 223: would be nice to have this timeout as a variable too, somewhere at the beginning of the file for readability/ease of configuration.

test/e2e/util.go
```diff
@@ -0,0 +1,448 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("aws-access-key-%s", randomURLString(16)),
+			Namespace: "external-dns-operator",
+		},
+		Data: map[string][]byte{
+			"aws_access_key_id":     originalCredsSecret.Data["aws_access_key_id"],
+			"aws_secret_access_key": originalCredsSecret.Data["aws_secret_access_key"],
+		},
+	}
+	if err := cl.Create(context.TODO(), &credsSecret); err != nil {
```
alebedev87, line 399: The external dns operator syncs the secret mentioned in `provider.aws.credentials` into the operator's namespace. I guess that this may be enough to just reference `kube-system/aws-creds` in ExternalDNS CR.

test/e2e/util.go
```diff
@@ -0,0 +1,448 @@
...
+		input := route53.ListResourceRecordSetsInput{
+			HostedZoneId:    &zoneID,
+			MaxItems:        StringRef("10"),
+			StartRecordName: &hostRecord,
+		}
+		output, err := dnsClient.aws.ListResourceRecordSets(&input)
+		if err != nil {
+			return false, err
+		}
+		for len(output.ResourceRecordSets) != 0 {
```
alebedev87, line 312: You wanted to say `if`?

test/e2e/operator_test.go
rfredette: The user needs to set them to point to a public zone. I've added a section to the readme with a bit of an explanation.

test/e2e/operator_test.go
rfredette: good catch; I've fixed it in the latest fixup

test/e2e/operator_test.go
rfredette: Either that, or it needs the same config that the aws cli uses, with the credentials at `.aws/credentials`. I will add some mention that aws creds (or equivalent for the cloud service being used) are needed in the README

Miciah:
Thanks!  Let's merge and iterate.  
/lgtm

test/e2e/util.go
rfredette: fixed, thanks

test/e2e/util.go
rfredette: That makes sense. I'll try this in a future iteration and simplify it if it works

PR #50, opened 2021-11-01
If the platform is OCP then we get the credentials from the secret we requested from cloud credentials operator

miheer:
If the platform is OCP then we get the credentials from the secret we requeste from cloud credentials operator.

This fixes https://issues.redhat.com/browse/CFE-135
Adds kubernetes fake package for mocking clients and objects as it is required for unit tests.

This PR unreverts commit 127835f3136ca10eea04dc8f1ee276d0f16cd447


miheer:
@arjunrn @Miciah

miheer:
@Miciah working on supporting GCP and Azure

miheer:
@alebedev87  @arjunrn PTAL

pkg/operator/controller/names.go
```diff
@@ -20,19 +20,63 @@ import (
...
+	for _, apiResource := range resources.APIResources {
+		if apiResource.Kind == "OpenShiftAPIServer" {
+			return OcpMajorVersion4
+		}
+	}
+	return NotocpMajorVersion4
+}
+
+func ExternalDNSCredentialsRequestName(externalDNS *operatorv1alpha1.ExternalDNS) types.NamespacedName {
+	fmt.Printf("externalDNS %v\n", externalDNS)
```
alebedev87, line 60: Debug leftover?

api/v1alpha1/externaldns_webhook.go
```diff
@@ -129,3 +148,21 @@ func (r *ExternalDNS) validateProviderCredentials() error {
 	}
 	return nil
 }
+
+// Returns Openshift Container Platform Major Version
+func GetOCPMajorVersion(kubeClient discovery.DiscoveryInterface) int {
```
alebedev87, line 153: I think that the function name doesn't reflect what it really does. For example, major version 3 will never be returned.
It may be more logical to name the function to something binary: `IsOCP4` or similar.

api/v1alpha1/externaldns_webhook.go
```diff
@@ -129,3 +148,21 @@ func (r *ExternalDNS) validateProviderCredentials() error {
...
+	// Since, CRD for OpenShift API Server was introduced in OCP v4.x we can verify if the current cluster is on OCP v4.x by
+	// ensuring that resource exists against Group(operator.openshift.io), Version(v1) and Kind(OpenShiftAPIServer)
+	// In case it doesn't exist we assume that external dns is running on non OCP 4.x environment
+	resources, err := kubeClient.ServerResourcesForGroupVersion("operator.openshift.io/v1")
+	if err != nil {
+		return NotocpMajorVersion4
+	}
+
+	for _, apiResource := range resources.APIResources {
+		if apiResource.Kind == "OpenShiftAPIServer" {
```
alebedev87, line 163: Can you make `OpenShiftAPIServer` a constant?

api/v1alpha1/externaldns_webhook.go
miheer: Actually OCP4 name will be specific 4th version. So, I think in the future we may have to add 5 as major version.

alebedev87: Yes, you are right, my example was too specific to OCP4.
What I wanted to say was more about the nature of the function. It's binary (yes/no answer) by nature for the moment, it doesn't return version 3 or won't return version 5 when we have it out, it just checks the presence of the OCP API resource.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		return false, nil, err
+	}
+	return true, cr, nil
+}
+
+// createExternalDNSCredentialsRequest creates the given credentials request.
+func (r *reconciler) createExternalDNSCredentialsRequest(ctx context.Context, desired *cco.CredentialsRequest) error {
+	if err := r.client.Create(ctx, desired); err != nil {
+		return fmt.Errorf("failed to create externalDNS cluster role %s: %w", desired.Name, err)
```
alebedev87, line 170:
```suggestion
		return fmt.Errorf("failed to create externalDNS credentials request %s: %w", desired.Name, err)
```

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,174 @@
...
+			},
+			PredefinedRoles: []string{
+				"roles/dns.admin",
+			},
+		})
+	return awsProvSpec, err
+}
+
+// desiredCredentialsRequestName returns the desired credentials request definition for externalDNS
+func desiredCredentialsRequestName(name types.NamespacedName, externalDNS *operatorv1alpha1.ExternalDNS) *cco.CredentialsRequest {
```
alebedev87, line 104: This function returns the whole instance of `CredentialsRequest` not just its name.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,174 @@
...
+			SecretRef: k8sv1.ObjectReference{
+				Name:      "externaldns-cloud-credentials",
+				Namespace: "external-dns-operator",
+			},
+		},
+	}
+
+	codec, err := cco.NewCodec()
+	if err != nil {
+		fmt.Printf("error creating new codec: %v", err)
```
alebedev87, line 121: 1. You should not use `fmt` for logging, use the logger given by the controller runtime, it's in `reconciler` object.
2. Do you need to log here? I think you can simply return an error with descriptive message.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,174 @@
...
+			},
+			RoleBindings: []cco.RoleBinding{
+				{Role: "Contributor"},
+			},
+		})
+	return awsProvSpec, err
+}
+
+func gcpProviderConfig(codec *cco.ProviderCodec) (*runtime.RawExtension, error) {
+	awsProvSpec, err := codec.EncodeProviderSpec(
```
alebedev87, line 91: It seems to be a copy paste mistake: `awsProvSpec`. If you just return the result of `EncodeProviderSpec` you won't have to worry about the name of the local variable:
```go
return codec.EncodeProviderSpec()
```

api/v1alpha1/externaldns_webhook.go
```diff
@@ -104,6 +120,9 @@ func (r *ExternalDNS) validateHostnameAnnotationPolicy() error {
 }
 
 func (r *ExternalDNS) validateProviderCredentials() error {
+	if ocpVersion == OcpMajorVersion4 {
```
alebedev87, line 123: We should not skip the validation for Infoblox and BlueCat DNS providers because CCO covers only AWS/Azure/GCP.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,174 @@
...
+				Kind: "AWSProviderSpec",
+			},
+			StatementEntries: []cco.StatementEntry{
+				{
+					Effect: "Allow",
+					Action: []string{
+						"elasticloadbalancing:DescribeLoadBalancers",
+						"route53:ListHostedZones",
+						"route53:ChangeResourceRecordSets",
+						"tag:GetResources"},
```
alebedev87, line 72: 1. Why do we need `elasticloadbalancing:DescribeLoadBalancers` ?
2. Why do we need `tag:GetResources`?
3. From [ExternalDNS AWS tutorial](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md#iam-policy) I see that `ListResourceRecordSets` action is also needed.
4. From the same tutorial I see that `ChangeResourceRecordSets` is limited to `arn:aws:route53:::hostedzone/*` resources, here we have it for all resources `*`, I'm wondering whether they are equivalent.

pkg/operator/controller/credentials-secret/controller.go
```diff
@@ -52,24 +58,30 @@ type Config struct {
...
-	scheme *runtime.Scheme
-	client client.Client
-	config Config
-	log    logr.Logger
+	cache      cache.Cache
+	scheme     *runtime.Scheme
+	client     client.Client
+	config     Config
+	log        logr.Logger
+	kubeclient discovery.DiscoveryInterface
```
alebedev87, line 66: The name doesn't seem to reflect well the functionality behind. Something like `kubeDiscovery` would be more relevant, wdyt?

api/v1alpha1/externaldns_webhook.go
```diff
@@ -22,16 +22,32 @@ import (
...
 var webhookLog = logf.Log.WithName("validating-webhook")
 
+var ocpVersion bool
+
 func (r *ExternalDNS) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	kubeClient, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return err
+	}
+	ocpVersion = isOCP(kubeClient)
```
alebedev87, line 50: I think `ocpVersion` name is no longer relevant after the function name change.

pkg/operator/controller/names.go
```diff
@@ -20,19 +20,63 @@ import (
...
+			Namespace: CredentialsRequestNamespace,
+			Name:      "externaldns-credentials-request-" + strings.ToLower(string(operatorv1alpha1.ProviderTypeAzure)),
+		}
+	case operatorv1alpha1.ProviderTypeGCP:
+		return types.NamespacedName{
+			Namespace: CredentialsRequestNamespace,
+			Name:      "externaldns-credentials-request-" + strings.ToLower(string(operatorv1alpha1.ProviderTypeGCP)),
+		}
+	}
+	return types.NamespacedName{}
```
arjunrn, line 77: Since we are only calling this function when the provider is of type AWS, Azure or GCP:
```suggestion
	return types.NamespacedName{
		Namespace: CredentialsRequestNamespace,
		Name:      "externaldns-credentials-request-" + strings.ToLower(string(externalDNS.Spec.Provider.Type))
	}
```

pkg/operator/controller/names.go
```diff
@@ -20,19 +20,63 @@ import (
...
-	ExternalDNSBaseName = "external-dns"
+	ExternalDNSBaseName         = "external-dns"
+	CredentialsRequestNamespace = "openshift-cloud-credential-operator"
+	kind                        = "OpenShiftAPIServer"
+	group                       = "operator.openshift.io"
+	version                     = "v1"
 )
 
+// Returns true if platform is OCP
+func IsOCP(kubeClient discovery.DiscoveryInterface) bool {
```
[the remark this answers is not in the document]
miheer: yes but do you know a common place where I can add it so that the webhook file also can access it?

api/v1alpha1/externaldns_webhook.go
```diff
@@ -129,3 +149,22 @@ func (r *ExternalDNS) validateProviderCredentials() error {
...
 	return nil
 }
+
+// Returns true if platform is OCP
+func isOCP(kubeClient discovery.DiscoveryInterface) bool {
+	// Since, CRD for OpenShift API Server was introduced in OCP v4.x we can verify if the current cluster is on OCP v4.x by
+	// ensuring that resource exists against Group(operator.openshift.io), Version(v1) and Kind(OpenShiftAPIServer)
+	// In case it doesn't exist we assume that external dns is running on non OCP 4.x environment
+	resources, err := kubeClient.ServerResourcesForGroupVersion(group + "/" + version)
+	fmt.Printf("\nresources %v", resources)
```
alebedev87, line 159: The logging should be done with the logger from the controller runtime.

pkg/operator/controller/names.go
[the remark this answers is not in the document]
arjunrn: Since the function in `externaldnscontroller` is already exported it can be used in the `api` package.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,179 @@
...
+
+// desiredCredentialsRequestName returns the desired credentials request definition for externalDNS
+func desiredCredentialsRequest(name types.NamespacedName, externalDNS *operatorv1alpha1.ExternalDNS) (*cco.CredentialsRequest, error) {
+	credentialsRequest := &cco.CredentialsRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+		},
+		Spec: cco.CredentialsRequestSpec{
+			ServiceAccountNames: []string{"external-dns-operator"},
```
alebedev87, line 120: I'm wondering why the operator's service account would need the operand's permissions.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return true, cr, nil
+}
+
+// createExternalDNSCredentialsRequest creates the given credentials request.
+func (r *reconciler) createExternalDNSCredentialsRequest(ctx context.Context, desired *cco.CredentialsRequest) error {
+	if err := r.client.Create(ctx, desired); err != nil {
+		return fmt.Errorf("failed to create externalDNS credentials request %s: %w", desired.Name, err)
+	}
+	r.log.Info("created externalDNS cluster role", "name", desired.Name)
```
alebedev87, line 177: It's still `cluster role`.

pkg/operator/controller/names.go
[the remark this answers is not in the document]
miheer: I did try that but gives a cyclic dependency error.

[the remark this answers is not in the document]
miheer: Ignore the above ^^ I got your point

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,179 @@
...
+	credentialsRequest := &cco.CredentialsRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+		},
+		Spec: cco.CredentialsRequestSpec{
+			ServiceAccountNames: []string{"external-dns-operator"},
+			SecretRef: k8sv1.ObjectReference{
+				Name:      "externaldns-cloud-credentials",
+				Namespace: "external-dns-operator",
```
alebedev87, line 123: There is a config field for the operator namespace: [link](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/config/config.go#L48).

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	exists, current, err := r.currentExternalDNSCredentialsRequest(ctx, name)
+
+	if err != nil {
+		return false, nil, err
+	}
+
+	if !exists {
+		if err := r.createExternalDNSCredentialsRequest(ctx, desired); err != nil {
```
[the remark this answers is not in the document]
miheer: I am creating the Credentials request only if platform is OCP and if Provider Types are AWS, GCP and Azure.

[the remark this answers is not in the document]
miheer: if someone specifies credentials in the CR when platform is OCP and if Provider Types are AWS, GCP and Azure then we get fetch the secret we got from CCO and those creds will be ignored.

pkg/operator/controller/credentials-secret/controller.go
```diff
@@ -143,6 +155,42 @@ func New(mgr manager.Manager, config Config) (controller.Controller, error) {
 		return requests
 	}
 
+	if err := mgr.GetFieldIndexer().IndexField(
+		context.Background(),
+		&operatorv1alpha1.ExternalDNS{},
+		credentialsSecretIndexFieldNameInOperand,
+		client.IndexerFunc(func(o client.Object) []string {
+			ed := o.(*operatorv1alpha1.ExternalDNS)
+			name := "external-dns-credentials-" + ed.Name
```
alebedev87, line 164: You can use the function from `controller` package to get the name of the secret from the target namespace: [link](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/names.go#L52) (operand namespace won't matter here).
If you want [the concatenation logic](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/names.go#L55) can even be put into a new dedicated function.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,179 @@
...
+			TypeMeta: metav1.TypeMeta{
+				Kind: "AWSProviderSpec",
+			},
+			StatementEntries: []cco.StatementEntry{
+				{
+					Effect: "Allow",
+					Action: []string{
+						"route53:ChangeResourceRecordSets",
+					},
+					Resource: "arn:aws:route53:::hostedzone/*",
```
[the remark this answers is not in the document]
miheer: actually i have just used the IAM which is mentioned in the external-dns doc for aws

pkg/operator/controller/names.go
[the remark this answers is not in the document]
miheer: go run sigs.k8s.io/controller-tools/cmd/controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" rbac:roleName=external-dns-operator webhook paths="./..." output:crd:artifacts:config=config/crd/bases
-: import cycle not allowed: import stack: [github.com/openshift/external-dns-operator/api/v1alpha1 github.com/openshift/external-dns-operator/pkg/operator/controller github.com/openshift/external-dns-operator/api/v1alpha1]
/home/miheer/.gvm/gos/go1.16.3/src/github.com/openshift/external-dns-operator/api/v1alpha1/externaldns_webhook.go:22:24: no such package located
Error: not all generators ran successfully

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,179 @@
...
+func desiredCredentialsRequest(name types.NamespacedName, externalDNS *operatorv1alpha1.ExternalDNS) (*cco.CredentialsRequest, error) {
+	credentialsRequest := &cco.CredentialsRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+		},
+		Spec: cco.CredentialsRequestSpec{
+			ServiceAccountNames: []string{"external-dns-operator"},
+			SecretRef: k8sv1.ObjectReference{
+				Name:      "externaldns-cloud-credentials",
```
alebedev87, line 122: You created a constant for this name: [here](https://github.com/openshift/external-dns-operator/blob/e1edc15be14caea60b976326ab22a829e16c0d3a/pkg/operator/controller/credentials-secret/controller.go#L50).
But the best would be to put this name to [controller package](https://github.com/openshift/external-dns-operator/blob/e1edc15be14caea60b976326ab22a829e16c0d3a/pkg/operator/controller/credentials-secret/controller.go#L50) like for the namespace below of for the destination secret name in the other comment.

pkg/operator/controller/names.go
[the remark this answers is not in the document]
miheer: I defined the function in the webhook so not it is accessible.

pkg/operator/controller/externaldns/credentials-request.go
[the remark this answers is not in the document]
alebedev87: Continuing Arjun's question about the changes.

I think that the check for the need of the update is missing in this `ensure` function, similar to what has been done for other resources: [cluster role example](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/externaldns/rbac.go#L58).
When the update function is there, `desired` variable would make sense to remain on its current place.

api/v1alpha1/externaldns_webhook.go
miheer: actually logging it is not needed. removed it.

pkg/operator/controller/externaldns/credentials-request.go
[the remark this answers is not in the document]
miheer: Update is not needed as we will creating secrets by making a call to CCO once the CR is created. Operator watches the src ns and adds that secret  to target ns if it is missing.

bundle/manifests/externaldns.olm.openshift.io_crd.yaml
```diff
@@ -77,7 +77,7 @@ spec:
                       - Exact
                       - Pattern
                       type: string
-                    names:
+                    name:
```
alebedev87, line 80: Hm, this one looks familiar :)
https://github.com/openshift/external-dns-operator/pull/55/files

pkg/operator/controller/externaldns/credentials-request.go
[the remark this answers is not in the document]
alebedev87: What would happen if I as cluster admin change `externaldns-credentials-request-aws` credentials request's provider spec to Azure one?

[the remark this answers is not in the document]
miheer: Azure credentials request will get created. I just removed the checks as @arjunrn mentioned to do so here  https://github.com/openshift/external-dns-operator/pull/50#discussion_r746615496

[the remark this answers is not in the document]
miheer: i think I shall revert that because we need checks, so if someone mistakenly says  create cred request Azure on AWS, the provider type will be checked and AWS credentials request will be created.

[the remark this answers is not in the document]
alebedev87: I think you are mixing up the provider type of ExternalDNS custome resource and the provider spec of `Credentials Request`. What I meant was the unsolicited changes in the provider spec of `CredentialsRequest`. For the moment your code only checks the presence of `CredentialsRequest` resource by name (`Get` API call), not its content. So if the name of the resource doesn't change but only its content (e.g. `ProviderSpec`) the operator won't notice that and the source (and then the destination) secret will contain undesired (modified by the third party) content (e.g.`ProviderSpec`).

pkg/operator/controller/names.go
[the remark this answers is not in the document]
arjunrn: Personally I would like to see this refactored completed. We don't need to verify if the cluster is an OCP cluster every time. Once at startup is sufficient and reasonable. So that bool could be passed as a parameter to both the webhook and the controller.

bundle/manifests/externaldns.olm.openshift.io_crd.yaml
miheer: yeah but then git shall not show this difference

pkg/operator/controller/names.go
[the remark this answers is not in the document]
miheer: done

pkg/operator/controller/externaldns/credentials-request.go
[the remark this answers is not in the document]
miheer: hmm OK i will add update function.

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,287 @@
...
+		panic(err)
+	}
+	if err := operatorv1alpha1.AddToScheme(testScheme); err != nil {
+		panic(err)
+	}
+
+	if err := cco.AddToScheme(testScheme); err != nil {
+		panic(err)
+	}
+}
```
alebedev87, line 42: `test/` package (in this same directory) already has the test scheme initialized.

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,287 @@
...
+			scheme: testScheme,
+			config: testConfig(),
+			log:    zap.New(zap.UseDevMode(true)),
+		}
+
+		t.Run(tc.name, func(t *testing.T) {
+
+			_, got, _ := r.ensureExternalCredentialsRequest(context.TODO(), tc.inputExtDNS)
+
+			if reflect.DeepEqual(got.Name, tc.expectedCredentialRequest.Name) == false {
```
alebedev87, line 97:
```suggestion
			if got.Name != tc.expectedCredentialRequest.Name {
```

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,287 @@
...
+				t.Errorf("Got name %v  but expected name is %v", got.Name, tc.expectedCredentialRequest.Name)
+			}
+
+			if reflect.DeepEqual(got.Namespace, tc.expectedCredentialRequest.Namespace) == false {
```
alebedev87, line 101:
```suggestion
			if got.Namespace != tc.expectedCredentialRequest.Namespace {
```

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,287 @@
...
+				t.Errorf("Got namespace %v  but expected namespace is %v", got.Namespace, tc.expectedCredentialRequest.Namespace)
+			}
+
+			if tc.inputExtDNS.Spec.Provider.Type == operatorv1alpha1.ProviderTypeAWS {
+				codec, _ := cco.NewCodec()
+				awsSpec := cco.AWSProviderSpec{}
+				err := codec.DecodeProviderSpec(got.Spec.ProviderSpec, &awsSpec)
+				if err != nil {
+					t.Errorf("error decoding provider v1 spec: %v", err)
+				}
```
alebedev87, line 111: Can you please factorize the decoding logic into a dedicated function? This would save a lot of space.

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,287 @@
...
+
+	switch providerType {
+
+	case operatorv1alpha1.ProviderTypeAWS:
+		codec, err := cco.NewCodec()
+		if err != nil {
+			fmt.Printf("error creating new codec: %v", err)
+			return nil
+		}
+		awsProviderSpec, err := awsProviderConfig(codec)
```
alebedev87, line 214: I don't quite get this. `ensureExternalCredentialsRequest` uses `<provider>ProviderConfig()` functions. Then you use the same function to get the expected object's provider spec. Obviously it will always be the same. What is tested then?

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,287 @@
...
+		r := &reconciler{
+			client: cl,
+			scheme: testScheme,
+			config: testConfig(),
+			log:    zap.New(zap.UseDevMode(true)),
+		}
+
+		t.Run(tc.name, func(t *testing.T) {
+
+			_, got, _ := r.ensureExternalCredentialsRequest(context.TODO(), tc.inputExtDNS)
```
alebedev87, line 95: Can you please check the existence and error too?

pkg/operator/controller/externaldns/credentials-request_test.go
miheer: we have to decode for comparison or else it tries to compares addresses and the cmp or deep reflect fails as in expected a "\n" char is getting added

alebedev87: What I meant was not about the encoding or decoding. I wanted to say that the spec you generate as the expected value will always be the same as the one generated by the function you are testing. This makes the test pointless.

pkg/operator/controller/names.go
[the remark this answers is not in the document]
alebedev87: I agree with Arjun on the point that the check is sufficient to be done once at startup.
This would also simplify the other parts of the code: for instance, Kube client's `DiscoveryInterface` won't need to be used in the controllers.

pkg/operator/controller/externaldns/credentials-request_test.go
miheer: no it does not make the test pointless as we are checking the contents of the provider spec if they match.

miheer: As mentioned to you in the expect a new line gets added not sure why. So after decoding we see human readable output and the reflect equal and go cmp return it as true.

miheer: you can try what I am saying...just try to compare the encoded the values

pkg/operator/controller/externaldns/credentials-request_test.go
miheer: I am getting this error after removing it

== RUN   TestEnsureCredentialsRequest
--- FAIL: TestEnsureCredentialsRequest (0.00s)
panic: failed to add runtime object &{{ } {test     999 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] {AWS 0xc0000c05b0 <nil> <nil> <nil> <nil>} {{OpenShiftRoute map[] <nil> <nil> <nil> <nil>}  []} [public-zone]} {[] 0 []}} to fake client: no kind is registered for the type v1alpha1.ExternalDNS in scheme "github.com/openshift/external-dns-operator/pkg/operator/controller/externaldns/credentials-request_test.go:25" [recovered]
	panic: failed to add runtime object &{{ } {test     999 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] {AWS 0xc0000c05b0 <nil> <nil> <nil> <nil>} {{OpenShiftRoute map[] <nil> <nil> <nil> <nil>}  []} [public-zone]} {[] 0 []}} to fake client: no kind is registered for the type v1alpha1.ExternalDNS in scheme "github.com/openshift/external-dns-operator/pkg/operator/controller/externaldns/credentials-request_test.go:25"

goroutine 48 [running]:
testing.tRunner.func1.2(0x17e5780, 0xc00063ba20)

pkg/operator/controller/externaldns/credentials-request_test.go
miheer: factorizing wont help here as  codec.DecodeProviderSpec expects provider specific Spec. it is very hard to explain here what the problem is. If you have some ideas then please explain. But this a very simple thing  like having a function with switch for provider types. Do you want a seperate functions for different providers ?

pkg/operator/controller/externaldns/credentials-request.go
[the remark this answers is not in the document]
miheer: done

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -95,14 +96,34 @@ func (r *reconciler) createCredentialsSecret(ctx context.Context, secret *corev1
...
+	var secret *corev1.Secret
+	var data map[string][]byte
+	secret = &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      destName.Name,
 			Namespace: destName.Namespace,
 		},
-		Data: sourceSecret.Data,
 	}
+	data = sourceSecret.Data
```
alebedev87, line 108: You know that `data` and `sourceSecret.Data` are now pointing to the same dataset?
When you change `data` this changes `sourceData` too.

I think this method needs to be refactored (to avoid the reference problem and the need to `delete` keys):
- `secret.Data` can be added with new values directly without this intermediate `data` variable
- for AWS case you need to do a `for` loop to get all key/values from `sourceSecret` to avoid the same reference problem as here

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -68,10 +73,28 @@ var sourceStringTable = map[operatorv1alpha1.ExternalDNSSourceType]string{
...
+		Group:   "config.openshift.io",
+		Version: "v1",
+		Kind:    "Infrastructure",
+	})*/
+	var platformStatus *configv1.PlatformStatus
+	var err error
+	if operatorv1alpha1.IsOpenShift {
+		infraConfig := &configv1.Infrastructure{}
+		if err = r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
+			return false, &appsv1.Deployment{}, fmt.Errorf("failed to get infrastructure 'config': %v", err)
```
alebedev87, line 86:
```suggestion
			return false, nil, fmt.Errorf("failed to get infrastructure 'config': %v", err)
```

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -68,10 +73,28 @@ var sourceStringTable = map[operatorv1alpha1.ExternalDNSSourceType]string{
...
+		}
+
+		platformStatus, err = util.GetPlatformStatus(r.client, infraConfig)
```
alebedev87, line 89: We don't need to get the platform status at each `ensureExternalDNSDeployment` call. This data doesn't change, same as `IsOpenShift` variable which you set once at the webhook setup.
This would avoid dragging this status through all the methods (and changing their definitions too) down to `fillXXXFields` methods. Also, the status is not used at all in case of not OpenShift deployment, so all the methods definition changed only for a specific case.

alebedev87 (changes requested):
I didn't look at the tests yet as I would like some errors to be fixed first. So, this is part1.

pkg/util/platform_status.go
```diff
@@ -0,0 +1,59 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetPlatformStatus provides a backwards-compatible way to look up platform status. AWS is the
+// special case. 4.1 clusters on AWS expose the region config only through install-config. New AWS clusters
```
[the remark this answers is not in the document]
alebedev87: @Miciah : do we need to drag the dependency on the cluster config map for the new components?

pkg/operator/controller/externaldns/pod.go
```diff
@@ -448,6 +455,7 @@ func (b *externalDNSVolumeBuilder) providerSpecificVolumes() []corev1.Volume {
 	case externalDNSProviderTypeAzure:
 		return b.azureVolumes()
 	case externalDNSProviderTypeGCP:
+		fmt.Printf("in case externalDNSProviderTypeGCP:")
```
alebedev87, line 458: debug

pkg/operator/controller/externaldns/pod.go
```diff
@@ -265,6 +267,7 @@ func (b *externalDNSContainerBuilder) fillProviderSpecificFields(container *core
 	case externalDNSProviderTypeAzure:
 		b.fillAzureFields(container)
 	case externalDNSProviderTypeGCP:
+		//b.fillGCPFields(container, platformStatus)
```
alebedev87, line 270: leftover

pkg/operator/controller/externaldns/pod.go
```diff
@@ -109,6 +110,7 @@ func (b *externalDNSContainerBuilder) buildSeq(seq int, zone string) (*corev1.Co
 	container := b.defaultContainer(controller.ExternalDNSContainerName(zone))
 	err := b.fillProviderAgnosticFields(seq, zone, container)
 	if err != nil {
+		fmt.Printf("\n if err != nil {")
```
alebedev87, line 113: debug

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -118,7 +122,9 @@ func (r *reconciler) currentExternalDNSDeployment(ctx context.Context, nsName ty
 }
 
 // desiredExternalDNSDeployment returns the desired deployment resource.
+//func desiredExternalDNSDeployment(namespace, image, secretName string, serviceAccount *corev1.ServiceAccount, externalDNS *operatorv1alpha1.ExternalDNS, platformStatus *configv1.PlatformStatus) (*appsv1.Deployment, error) {
```
alebedev87, line 125: leftover

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -18,9 +18,12 @@ package externaldnscontroller
 
 import (
 	"context"
+
 	"fmt"
 	"sort"
 
+	//configv1 "github.com/openshift/api/config/v1"
```
alebedev87, line 25: leftover

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -95,14 +98,44 @@ func (r *reconciler) createCredentialsSecret(ctx context.Context, secret *corev1
...
+		secret.Data = data
+	} else {
+		data := sourceSecret.Data
+		secret.Data = data
+	}
+
+	return secret
+}
+
+func Base64Decode(src []byte) ([]byte, error) {
```
alebedev87, line 137: is this function used somewhere?

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -95,14 +98,44 @@ func (r *reconciler) createCredentialsSecret(ctx context.Context, secret *corev1
...
+			"tenantId":        string(sourceSecret.Data["azure_tenant_id"]),
+		}
+		azureMarshalledJson, _ := json.Marshal(azure_map)
+		data := map[string][]byte{
+			"azure.json": azureMarshalledJson,
+		}
+		secret.Data = data
+	} else {
+		data := sourceSecret.Data
+		secret.Data = data
```
alebedev87, line 131:
```suggestion
		secret.Data = sourceSecret.Data
```

pkg/operator/controller/credentials-secret/controller.go
```diff
@@ -200,7 +261,10 @@ func isInNS(namespace string) func(o client.Object) bool {
 }
 
 // getExternalDNSCredentialsSecretName returns the name of the credentials secret retrieved from externalDNS resource
-func getExternalDNSCredentialsSecretName(externalDNS *operatorv1alpha1.ExternalDNS) string {
+func (r *reconciler) getExternalDNSCredentialsSecretName(externalDNS *operatorv1alpha1.ExternalDNS) string {
```
alebedev87, line 264: should this function still be a method of the reconciler? I think it's not needed anymore as `IsOpenShift` is global now, no?

pkg/util/platform_status.go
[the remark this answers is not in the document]
miheer: @alebedev87 I think there is no harm to have this.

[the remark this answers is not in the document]
miheer: @alebedev87 if I remove it I will again to a lot of testing :) I am using the way we use for cluster ingress operator.  I think we can do such type of refactoring once e2e tests are in place.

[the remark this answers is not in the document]
miheer: @alebedev87 I have removed it as per your expectation.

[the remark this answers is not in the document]
miheer: @alebedev87 PTAL . Lets merge this if you are happy.

pkg/operator/controller/externaldns/controller_test.go
```diff
@@ -128,12 +188,44 @@ func TestReconcile(t *testing.T) {
...
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cluster",
+					},
+					Status: configv1.InfrastructureStatus{
+						PlatformStatus: &configv1.PlatformStatus{
+							Type: configv1.GCPPlatformType,
+						},
+					},
+				}
```
alebedev87, line 219: why this is simply not added to the testcase's `existingObjects` field?

pkg/operator/controller/externaldns/controller_test.go
```diff
@@ -128,12 +188,44 @@ func TestReconcile(t *testing.T) {
...
 			r := &reconciler{
-				client: cl,
-				scheme: test.Scheme,
-				config: tc.inputConfig,
-				log:    zap.New(zap.UseDevMode(true)),
+				client:     cl,
+				scheme:     test.Scheme,
+				config:     tc.inputConfig,
+				log:        zap.New(zap.UseDevMode(true)),
+				kubeclient: kubeFakeClient.Discovery(),
```
alebedev87, line 198: By the way, why do we need kubeclient in the reconciler? Isn't it again the leftover from the times when `IsOpenShift` was not global?

alebedev87 (commented):
Part 2. Unit tests.

bundle/manifests/external-dns-operator-auth-proxy_v1_service.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   labels:
     name: external-dns-operator
   name: external-dns-operator-metrics-service
+  namespace: external-dns-operator
```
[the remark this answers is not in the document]
miheer: But the code will fail. I will remove it but not at my risk.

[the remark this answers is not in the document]
miheer: i think you have sent a PR right ? Then may be lets merge that after this one ?

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,435 @@
...
+			config: testConfig(),
+			log:    zap.New(zap.UseDevMode(true)),
+		}
+
+		t.Run(tc.name, func(t *testing.T) {
+
+			exists, got, err := r.ensureExternalCredentialsRequest(context.TODO(), tc.inputExtDNS)
+
+			if err != nil {
+				t.Errorf("Error while ensuring Credentials request")
```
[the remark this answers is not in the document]
miheer: @alebedev87  just for learning why t.Log?

bundle/manifests/external-dns-operator-auth-proxy_v1_service.yaml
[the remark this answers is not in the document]
miheer: anyway as per your wish I have removed it.

pkg/operator/controller/externaldns/controller_test.go
miheer: remove it as it is not needed

pkg/operator/controller/externaldns/controller_test.go
```diff
@@ -116,6 +122,60 @@ func TestReconcile(t *testing.T) {
...
 			},
 		},
+		{
+			name:            "Bootstrap when OCP",
+			existingObjects: []runtime.Object{testExtDNSInstance(), testSecret()},
+			inputConfig:     testConfig(),
+			inputRequest:    testRequest(),
+			ocpPlatform:     true,
+			expectedResult:  reconcile.Result{},
+			expectedEvents: []testEvent{
```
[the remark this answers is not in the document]
miheer: done

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,408 @@
...
+
+}
+
+func decodeGCPProviderSpec(gotCrededentialrequest, expectedCredentialRequest cco.CredentialsRequest) (gotDecodedGCPSpec, expectedDecodedGCPSpec cco.GCPProviderSpec) {
+
+	codec, _ := cco.NewCodec()
+	gotDecodedGCPSpec = cco.GCPProviderSpec{}
+	err := codec.DecodeProviderSpec(gotCrededentialrequest.Spec.ProviderSpec, &gotDecodedGCPSpec)
+	if err != nil {
+		return gotDecodedGCPSpec, cco.GCPProviderSpec{}
```
alebedev87, line 147: I'm still not quite convinced that these 2 cannot be the same. Return an error if the decode fails and check it in the test, this would avoid any possible comparison mistake.

bundle/manifests/external-dns-operator-auth-proxy_v1_service.yaml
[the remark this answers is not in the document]
alebedev87: The code won't fail, service packaged in OLM bundle will be created in the target namespace set by the user who does the installation. The bug was in the other files and yes it's fixed with this PR: https://github.com/openshift/external-dns-operator/pull/68 (thanks for LGTM!).

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,264 @@
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "CredentialsRequest",
+			APIVersion: "cloudcredential.openshift.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name.Name,
+			Namespace: name.Namespace,
+		},
+		Spec: cco.CredentialsRequestSpec{
+			ServiceAccountNames: []string{controller.ServiceAccountName, controller.ExternalDNSResourceName(externalDNS)},
```
alebedev87, line 223: Why the operand service account is here?

pkg/operator/controller/credentials-secret/controller_test.go
```diff
@@ -112,26 +142,39 @@ func TestReconcile(t *testing.T) {
...
+				config:        tc.inputConfig,
+				log:           zap.New(zap.UseDevMode(true)),
+				kubeDiscovery: kubeFakeClient.Discovery(),
+			}
+
+			if tc.ocpPlatform {
+				operatorv1alpha1.IsOpenShift = true
+			} else {
+				operatorv1alpha1.IsOpenShift = false
 			}
```
alebedev87, line 176:
```suggestion
			operatorv1alpha1.IsOpenShift = tc.ocpPlatform 
```

pkg/operator/controller/credentials-secret/controller_test.go
```diff
@@ -72,13 +63,51 @@ func TestReconcile(t *testing.T) {
...
+				},
+			},
+		},
 		{
 			name:            "Bootstrap",
 			existingObjects: []runtime.Object{testAWSExtDNSInstance(), testSrcSecret()},
 			inputConfig:     testConfig(),
 			inputRequest:    testRequest(),
 			expectedResult:  reconcile.Result{},
+			ocpPlatform:     false,
```
alebedev87, line 110: No need to set it for non OCP test cases, `false` is the default value of `bool`.

pkg/operator/controller/externaldns/controller_test.go
```diff
@@ -128,12 +196,20 @@ func TestReconcile(t *testing.T) {
...
 				config: tc.inputConfig,
 				log:    zap.New(zap.UseDevMode(true)),
 			}
+
+			if tc.ocpPlatform {
+				operatorv1alpha1.IsOpenShift = true
+			} else {
+				operatorv1alpha1.IsOpenShift = false
+			}
+
```
alebedev87, line 212:
```suggestion
			operatorv1alpha1.IsOpenShift = tc.ocpPlatform

```

pkg/operator/controller/externaldns/credentials-request_test.go
```diff
@@ -0,0 +1,417 @@
...
+			}
+
+			if got.Namespace != tc.expectedCredentialRequest.Namespace {
+				t.Errorf("Got namespace %v  but expected namespace is %v", got.Namespace, tc.expectedCredentialRequest.Namespace)
+			}
+
+			if tc.inputExtDNS.Spec.Provider.Type == operatorv1alpha1.ProviderTypeAWS {
+				gotDecodedAWSSpec, expectedAWSSpec, err := decodeAWSProviderSpec(*got, *tc.expectedCredentialRequest)
+				if err != nil {
+					t.Log("Not able to decode AWS Provider Spec because of", err)
```
alebedev87, line 114: It must be `t.Errorf` to `t.Log`.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,264 @@
...
+	} else if updated {
+		return r.currentExternalDNSCredentialsRequest(ctx, name)
+	}
+
+	return true, current, nil
+}
+
+// updateExternalDNSClusterRole updates the cluster role with the desired state if the rules differ
+func (r *reconciler) updateExternalDNSCredentialsRequest(ctx context.Context, current, desired *cco.CredentialsRequest, externalDNS *operatorv1alpha1.ExternalDNS) (bool, error) {
+	changed, reason := externalDNSCredentialsRequestChanged(current, desired, externalDNS)
```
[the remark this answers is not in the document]
alebedev87: Error is still not handled

pkg/operator/controller/externaldns/credentials-request.go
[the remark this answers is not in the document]
alebedev87: I agree with Arjun, this can be more granular, let's keep it for followup improvements.

pkg/operator/controller/externaldns/credentials-request.go
```diff
@@ -0,0 +1,264 @@
...
+		return r.currentExternalDNSCredentialsRequest(ctx, name)
+	}
+
+	return true, current, nil
+}
+
+// updateExternalDNSClusterRole updates the cluster role with the desired state if the rules differ
+func (r *reconciler) updateExternalDNSCredentialsRequest(ctx context.Context, current, desired *cco.CredentialsRequest, externalDNS *operatorv1alpha1.ExternalDNS) (bool, error) {
+	var updated *cco.CredentialsRequest
+	changed, err := externalDNSCredentialsRequestChanged(current, desired, externalDNS)
```
alebedev87, line 75: This error must be checked and returned if it's not `nil`.

alebedev87 (approved):

miheer:
@alebedev87 thanks a lot! Your reviews are valuable to me. I learnt a lot from them.

pkg/util/platform_status.go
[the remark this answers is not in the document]
Miciah: The isn't that the operator could be running on an OpenShift 4.1 cluster; the issue is that the operator could be running on a cluster that was installed at version 4.1 and upgraded to 4.10, and such a cluster might not have `PlatformStatus` set.  

That said, OpenShift-specific logic should be in in a separate package as much as possible to avoid having platform-specific logic in the main logic.

PR #51, opened 2021-11-01
Bug: 2019001  OLM operators shall not be a part of realease payload.

miheer:
OLM operator shall not be a part of Release Payload as the release payload is composed of components that are critical to the core platform and cannot be installed through OLM (e.g. control plane and kubelet).

This commit removes this operator from release payload as it was causing olm /bundle/manifests to get copied in /manifests which CVO was consuming and causing failures as they were meant for OLM.


Miciah:
/retitle Bug 2019001: OLM operators should not be included in the release payload

PR #55, opened 2021-11-05
This commit fixes the names plural field to singular name which is needed Match Type Exact.

miheer:
This commit fixes the names plural field to singular name which is needed Match Type Exact.


alebedev87:
Weird, this was already fixed in PR43: https://github.com/openshift/external-dns-operator/pull/43/files

miheer:
> Weird, this was already fixed in PR43: https://github.com/openshift/external-dns-operator/pull/43/files

Looks like you did not try/test via olm. Because bundle-image-build would have copied it to bundle/manifests folder

miheer:
It was giving me error while creating CR as the webhook does not allow plural name

PR #56, opened 2021-11-08
implement the recommendations of the security review report

sherine-k:

- Stop using github.com/form3tech-oss/jwt-go. The report recommends using github.com/golang-jwt/jwt instead. But since this library is not really used, we suggest to remove it for now.
- Update RBAC configuration, restricting access to resources such as secrets, serviceaccounts to the operator and operand namespaces only
  - Namespace permissions remain on the clusterRole
  - Permissions for secrets creation/update in the operator namespace are given through a new role (generated by kubebuilder annotations)
  - Permissions for secrets & serviceaccounts creation/update in the operand namespace are given through a role and rolebinding under the configure/operand_rbac folder. Makefile's targets deploy, undeploy and olm-manifest are updated to this effect


config/operand_rbac/operand_rbac.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: external-dns-operator-operand
+  namespace: external-dns
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: external-dns-operator-operand
+subjects:
+- kind: ServiceAccount
+  name:  external-dns-operator
+  namespace:  external-dns-operator
```
alebedev87, line 42: Can you please add the newline char at the end of the line?

sherine-k:
/assign @Miciah 
Hi @Miciah, I would like to have your opinion about this PR.
Thanks for your help
Sherine

Makefile
```diff
@@ -111,10 +111,13 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/operand_rbac | kubectl apply -f -
 
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build config/operand_rbac | kubectl delete -f -
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
```
Miciah, line 118: Does this work properly?  If I understand correctly, the first `kubectl delete` command would delete the namespace because the namespace definition is in `config/operand_rbac/operand_rbac.yaml`, but deleting the namespace would also delete the objects inside the namespace, so the objects that the second `kubectl delete` command is intended to delete would already deleted by the first `kubectl delete` command.  

Maybe just deleting the namespace would be sufficient for now.  On the other hand, we might want the operator to have the chance to do some cleanup before RBAC and the operator itself are deleted.  If that isn't currently a requirement though, we can just do the simplest thing (deleting the namespace) for now if that suffices.  

Should this rule or another rule clean up the operands that the operator has created?

sherine-k: Hi @Miciah , I put the operand namespace (+contents) delete first so that I can ensure that when `make undeploy` is run, all roles and namespace are also gone. Strategy being to remove operand first, operator later.
If we want to let the operator cleanup first, then remove operand namespace, I can simply reverse the order.

sherine-k:
Hi @Miciah, @alebedev87 
I'll reopen a new PR with the same content. My rebase does not look great here.
I also have a big doubt about removing `k8s.io/client-go/plugin/pkg/client/auth` from `cmd/external-dns-operator/main.go`
The comment right before it explains it's needed at runtime.
What do you think?

PR #57, opened 2021-11-09
Add OLM bundle and backport annotations

alebedev87:
`annotations.yaml` will be cloned by CPaaS midstream but the added annotations won't be customized by the CPaaS build.


lmzuccarelli (commented):
TBH I'm not familiar with the backport annotations. The change looks straight forward enough though

PR #59, opened 2021-11-15
implement the recommendations of the security review report (2)

sherine-k:

    Stop using github.com/form3tech-oss/jwt-go. The report recommends using github.com/golang-jwt/jwt instead. But since this library is not really used, we suggest to remove it for now.
    Update RBAC configuration, restricting access to resources such as secrets, serviceaccounts to the operator and operand namespaces only
        Namespace permissions remain on the clusterRole
        Permissions for secrets creation/update in the operator namespace are given through a new role (generated by kubebuilder annotations)
        Permissions for secrets & serviceaccounts creation/update in the operand namespace are given through a role and rolebinding under the configure/operand_rbac folder. Makefile's targets deploy, undeploy and olm-manifest are updated to this effect


Makefile
```diff
@@ -30,7 +30,7 @@ BIN_DIR=$(shell pwd)/bin
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-CONTAINER_ENGINE ?= podman
+CONTAINER_ENGINE ?= docker
```
frobware: Does the security review require this change?

sherine-k: Sorry, well spotted, I'll change it back

Makefile
```diff
@@ -121,9 +121,11 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/operand_rbac | kubectl apply -f -
 
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
+	$(KUSTOMIZE) build config/operand_rbac | kubectl delete -f -
```
Miciah, line 128: Sorry, I think told you the wrong thing on the previous PR.  We should delete the operand first so that the operator can clean things up, and then we should delete the operator namespace.

pkg/operator/operator.go
```diff
@@ -59,12 +59,13 @@ type Operator struct {
...
 // +kubebuilder:rbac:groups=externaldns.olm.openshift.io,resources=externaldnses/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=externaldns.olm.openshift.io,resources=externaldnses/finalizers,verbs=update
-// +kubebuilder:rbac:groups="",resources=namespaces;serviceaccounts;secrets,verbs=get;list;watch;delete;create;update
+// +kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch;delete;create;update
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;delete;create;update
 // +kubebuilder:rbac:groups="",resources=services;endpoints;pods;nodes,verbs=get;list;watch
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,verbs=get;list;watch
 // +kubebuilder:rbac:groups=route.openshift.io,resources=routes,verbs=get;watch;list
+// +kubebuilder:rbac:groups="",namespace=external-dns-operator,resources=secrets,verbs=get;list;watch;delete;create;update
```
Miciah, line 69: Do you also need to add a rule to allow the operator to create the operand serviceaccount, or was that permission superfluous all along?

Makefile
sherine-k: Sure, sounds more logical to me as well

PR #62, opened 2021-11-17
Renable e2e test with automated hosted zone provisioning

arjunrn:


test/e2e/operator_test.go
```diff
@@ -62,240 +49,185 @@ func init() {
...
-			os.Exit(1)
-		}
-		defaultParentZoneID = os.Getenv("EXTDNS_PARENT_ZONEID")
-		if defaultParentZoneID == "" {
-			fmt.Printf("No parent zone ID specified. Please set the environment variable EXTDNS_PARENT_ZONEID\n")
-			os.Exit(1)
-		}
-
-		domainName = generateDomainName()
+	hostedZoneID, nameServers, err = helper.ensureHostedZone(hostedZoneDomain)
```
alebedev87, line 122: If there are 2 PR running this e2e test, won't they race on the usage on the same hostedZone? Knowing also that the test service has the hard coded name, there can be false positive and negative tests.
This may need to be refactored in the followups to try to get the unique hostedzone name with the commit checksum in the name, for instance.

test/e2e/operator_test.go
```diff
@@ -62,240 +49,185 @@ func init() {
...
-		httpClient = &http.Client{
-			Transport: &http.Transport{
-				Dial:        dialer.Dial,
-				DialContext: dialer.DialContext,
-			},
-		}
-
-		os.Exit(m.Run())
-	*/
+	err = helper.deleteHostedZone(hostedZoneID)
```
alebedev87, line 130: Do we want to `defer` this deletion right after successful `ensureHostedZone`?

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.15.0
 	github.com/openshift/api v0.0.0-20211019100638-b2cbe79f2e4b
+	github.com/openshift/cloud-credential-operator v0.0.0-20211115195328-4c537fbd48e9
```
alebedev87, line 12: Do we still need this dependency knowing that we're getting the root credentials now?

test/e2e/operator_test.go
arjunrn: We cannot deferred calls because we have to call `os.Exit()` and that immediately exits the program. We can move all the logic into another function just to use defer but I think that might be overkill for now.

test/e2e/operator_test.go
arjunrn: Yes, this need to be done otherwise we might get too many flakes. Let's not merge this as-is.

test/e2e/operator_test.go
```diff
@@ -6,53 +6,40 @@ import (
...
 )
 
 const (
-	recordExistsTimeout  = 5 * time.Minute
-	podReadyTimeout      = 1 * time.Minute
-	httpGetTimeout       = 10 * time.Minute
-	recordDeletedTimeout = 5 * time.Minute
-	dnsLookupTimeout     = 5 * time.Second
-	dnsServer            = "8.8.8.8:53"
+	hostedZoneDomain   = "example-test.info"
```
arjunrn, line 30: The hosted zone need to be randomized so concurrent executions of the e2e tests do not interfere with each other.

test/e2e/operator_test.go
alebedev87: Yes, it's kind of a choice between:
- last step cleanup (current implementation) which fails the test if the cleanup broke but leave leftover zones of the test cases panic
- best effort cleanup (no `Exit` with `defer`) which won't fail the test if the clean broke but will try to do something the test cases panic

test/e2e/operator_test.go
```diff
@@ -64,238 +51,180 @@ func init() {
...
-		current := podConditionMap(pod.Status.Conditions...)
-		return conditionsMatchExpected(expected, current), nil
-	})
-	err = wait.PollImmediate(5*time.Second, httpGetTimeout, func() (bool, error) {
-		var resp *http.Response
-		resp, err := httpClient.Get(fmt.Sprintf("http://%s", expectedDomainName))
+
+		// get the IPs of the loadbalancer
+		lbHostname := service.Status.LoadBalancer.Ingress[0].Hostname
+		ips, err := net.LookupIP(lbHostname)
```
arjunrn, line 189: This uses the default resolver where `PreferGo` is false.

alebedev87:
Merging this one to iterate on the remark in later PRs.
/lgtm

PR #66, opened 2021-11-22
e2e test: use unique hosted zone and custom resolver

alebedev87:
Follow-up of https://github.com/openshift/external-dns-operator/pull/62. Aims at addressing the remarks about the conflicting hosted zone and go built-in resolver.


alebedev87:
Added the probing of all nameservers instead of just the first one.

PR #70, opened 2021-11-24
Ne 741/e2e azure

Connect2naga:
1. Adding e2e test case for Azure provider
2. Enchancing the hostedzone name creation on e2e test
3. if the hostname of the test service is empty then look for IP address.


test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	RESOURCE_GROUP = "azure_resourcegroup"
+	SUBSCIPTION_ID = "azure_subscription_id"
+	TANENT_ID      = "azure_tenant_id"
```
lmzuccarelli, line 29: nit: should this be TENANT_ID ?

test/e2e/operator_test.go
```diff
@@ -191,17 +198,24 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
 		}
 
 		// get the IPs of the loadbalancer
-		lbHostname := service.Status.LoadBalancer.Ingress[0].Hostname
-		// use built in Go resolver instead of the platform's one
-		ips, err := customResolver("").LookupIP(context.TODO(), "ip", lbHostname)
-		if err != nil {
-			t.Logf("waiting for loadbalancer IP for %s", lbHostname)
-			// if the hostname cannot be resolved currently then retry later
+		if service.Status.LoadBalancer.Ingress[0].Hostname != "" {
```
lmzuccarelli, line 201: Please check Andrey's previous changes 

```// try all nameservers and fail only if all failed
	for _, nameSrv := range nameServers {
		t.Logf("looking for DNS record in nameserver: %s", nameSrv)
		// create a DNS resolver which uses the nameservers of the test hosted zone
		customResolver := customResolver(nameSrv)

		// verify that the IPs of the record created by ExternalDNS match the IPs of loadbalancer obtained in the previous step.
		if err := wait.PollImmediate(dnsPollingInterval, dnsPollingTimeout, func() (done bool, err error) {
			ips, err := customResolver.LookupHost(context.TODO(), fmt.Sprintf("%s.%s", testServiceName, hostedZoneDomain))
			if err != nil {
				t.Log("waiting for dns record")
				return false, nil
			}
			for _, ip := range ips {
				if _, ok := serviceIPs[ip]; !ok {
					return false, nil
				}
			}
			return true, nil
		}); err != nil {
			t.Logf("failed to verify that DNS has been correctly set.")
		} else {
			return
		}
	}```

**Review by lmzuccarelli — CHANGES_REQUESTED** [review]
Great work Naga. Made some comments

File: test/e2e/azure.go (hunk 1 above)
Comment (Connect2naga) [reply test/e2e/azure.go]: Corrected the typo

> 
Thanks Luigi :)

File: test/e2e/operator_test.go (hunk 2 above)
Comment (Connect2naga) [reply test/e2e/operator_test.go:201]: sure, Will cross check with @alebedev87.

File: test/e2e/azure.go [hunk 3]
```diff
@@ -0,0 +1,223 @@
...
+
+const (
+	RESOURCE_GROUP = "azure_resourcegroup"
+	SUBSCIPTION_ID = "azure_subscription_id"
+	TENANT_ID      = "azure_tenant_id"
+	CLIENT_ID      = "azure_client_id"
+	CLIENT_SECRET  = "azure_client_secret"
+	REGION         = "azure_region"
+
+	KUBE_SYSTEM_SECRET_NAME = "azure-credentials"
```
alebedev87, line 34: This may be more visible to put all caps for consts but it's not compliant with the rest of the code in this (and other openshift repositories).

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+type cluserConfig struct {
+	Cloud                       string            `json:"cloud" yaml:"cloud"`
+	Environment                 azure.Environment `json:"-" yaml:"-"`
+	TenantID                    string            `json:"tenantId" yaml:"tenantId"`
+	SubscriptionID              string            `json:"subscriptionId" yaml:"subscriptionId"`
+	ResourceGroup               string            `json:"resourceGroup" yaml:"resourceGroup"`
+	Location                    string            `json:"location" yaml:"location"`
+	ClientID                    string            `json:"aadClientId" yaml:"aadClientId"`
+	ClientSecret                string            `json:"aadClientSecret" yaml:"aadClientSecret"`
+	UseManagedIdentityExtension bool              `json:"useManagedIdentityExtension" yaml:"useManagedIdentityExtension"`
```
alebedev87, line 47: Why do we need this variable? It's not used anywhere, seems like some leftover from previous versions of `getAccessToken`.

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+	return false, nil
+}
+
+// ref: https://github.com/kubernetes-sigs/external-dns/blob/master/provider/azure/azure.go
+// getAccessToken retrieves Azure API access token.
+func getAccessToken(cfg *cluserConfig) (*adal.ServicePrincipalToken, error) {
+	// Try to retrieve token with service principal credentials.
+	// Try to use service principal first, some AKS clusters are in an intermediate state that `UseManagedIdentityExtension` is `true`
```
alebedev87, line 202: Not sure this comment is still valid as `UseManagedIdentityExtension` seems to be a leftover.

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+			"azure.json": azureCreds,
+		},
+	}
+}
+
+func (a *azureTestHelper) platform() string {
+	return string(configv1.AzurePlatformType)
+}
+
+func (a *azureTestHelper) ensureHostedZone(rootDomain string) (string, []string, error) {
```
alebedev87, line 147: This method is inconsistent with other providers which check for existence of the zone and don't create if it's already there. Not sure it's the best way but let's keep all consistent.

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+	a.zoneName = rootDomain
+	return zoneID, nameservers, nil
+}
+
+func (a *azureTestHelper) deleteHostedZone(rootDomain string) error {
+	if a.zoneName == "" {
+		fmt.Printf("ZoneName is empty, nothing to be deleted")
+		return nil
+	}
+	if _, err := a.zoneClient.Delete(context.TODO(), a.config.ResourceGroup, a.zoneName, ""); err != nil {
```
alebedev87, line 166: Why setting `zoneName` to `rootDomain` in `ensure*` method just to use it here when you have `rootDomain` as argument of the method?

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+		return fmt.Errorf("unable to delete zone :%s, failed error: %v", a.zoneName, err)
+	}
+	// verify the zone is present
+	if _, err := a.isZoneNameExists(); err != nil {
```
alebedev87, line 170: Why should we not trust Azure SDK and verify the deletion?

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+	TENANT_ID      = "azure_tenant_id"
+	CLIENT_ID      = "azure_client_id"
+	CLIENT_SECRET  = "azure_client_secret"
+	REGION         = "azure_region"
+
+	KUBE_SYSTEM_SECRET_NAME = "azure-credentials"
+)
+
+// config represents common config items for Azure DNS and Azure Private DNS
+type cluserConfig struct {
```
alebedev87, line 38: Why this structure has json tags? I didn't see where it's marshaled/unmarshared in the code.

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+	SUBSCIPTION_ID = "azure_subscription_id"
+	TENANT_ID      = "azure_tenant_id"
+	CLIENT_ID      = "azure_client_id"
+	CLIENT_SECRET  = "azure_client_secret"
+	REGION         = "azure_region"
+
+	KUBE_SYSTEM_SECRET_NAME = "azure-credentials"
+)
+
+// config represents common config items for Azure DNS and Azure Private DNS
```
alebedev87, line 37:
```suggestion
// cluserConfig represents common config items for Azure DNS and Azure Private DNS
```

test/e2e/operator_test.go
```diff
@@ -121,6 +125,9 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	if version.SHORTCOMMIT != "" {
+		hostedZoneDomain = version.SHORTCOMMIT + "." + baseZoneDomain
+	}
```
alebedev87, line 130: Why this change? `SHORTCOMMIT` is supposed to be there, it's set at compile time [here](https://github.com/openshift/external-dns-operator/blob/main/Makefile#L87).

test/e2e/operator_test.go
```diff
@@ -191,17 +198,24 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
...
+			ips, err := customResolver("").LookupIP(context.TODO(), "ip", lbHostname)
+			if err != nil {
+				t.Logf("waiting for loadbalancer IP for %s", lbHostname)
+				// if the hostname cannot be resolved currently then retry later
+				return false, nil
+			}
+			for _, ip := range ips {
+				serviceIPs[ip.String()] = struct{}{}
+			}
+		} else if service.Status.LoadBalancer.Ingress[0].IP != "" {
```
alebedev87, line 213: I think it makes more sense to check the presence of IP first and only if it's not there check the hostname.

alebedev87 (commented):
Thanks for your first PR in ExternalDNS, @Connect2naga !
I left some comments though.

Also I really would like to see the CI job added first to avoid [the problem we got with GCP](https://github.com/openshift/release/pull/23923#issuecomment-978481128).

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+		TenantID:       a.config.TenantID,
+		SubscriptionID: a.config.SubscriptionID,
+		ResourceGroup:  a.config.ResourceGroup,
+		ClientID:       a.config.ClientID,
+		ClientSecret:   a.config.ClientSecret,
+	}
+	azureCreds, err := json.Marshal(credData)
+	if err != nil {
+
+	}
```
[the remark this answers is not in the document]
Connect2naga: done.

test/e2e/operator_test.go
```diff
@@ -1,4 +1,5 @@
-//+build e2e
+//go:build e2e
+// +build e2e
```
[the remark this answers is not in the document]
Connect2naga: yes these will be 1.17, i will remove this go build tag & verify
so as part of the build this `make verify` should fail ? am i right?

test/e2e/azure.go
```diff
@@ -0,0 +1,223 @@
...
+		SubscriptionID:              string(secret.Data[SUBSCIPTION_ID]),
+		ResourceGroup:               string(secret.Data[RESOURCE_GROUP]),
+		Location:                    string(secret.Data[REGION]),
+		ClientID:                    string(secret.Data[CLIENT_ID]),
+		ClientSecret:                string(secret.Data[CLIENT_SECRET]),
+		UseManagedIdentityExtension: true,
+	}
+
+	var environment azure.Environment
+	if a.config.Cloud == "" {
```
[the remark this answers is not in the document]
Connect2naga: yes, for local test had modified the code.
taken care.

test/e2e/azure.go
Connect2naga: except  `UseManagedIdentityExtension`, others we are using,   this will be takenout.

test/e2e/azure.go
Connect2naga: this is additional check kept for one enhancement. 
in case test case fails, it close the test without cleaning, so that time these zones will be left over.
this will not cause any issue since SDK support createOrUpdate().

as mentioned in comment, it will be verified apart of deletion.

test/e2e/azure.go
Connect2naga: this is taken from external-dns ref, initial aim is to read from file and and parse.
since we already having value & directly assigned.
will remove json tags.

test/e2e/azure.go
Connect2naga: taken care

test/e2e/operator_test.go
Connect2naga: yes, since it is compile time.
when we execute in local, this value will be empty so the `hostedZoneDomain` look like ".example-test.info"
which is invalid for azure to create zone.

test/e2e/operator_test.go
Connect2naga: actually did same, but changed :), will do the revert.

test/e2e/azure.go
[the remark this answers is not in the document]
Connect2naga: CI job changes committed  : https://github.com/openshift/release/pull/23940
also verified in local.

Connect2naga:
Verified in local to test e2e with Azure & checked `make verify`, results is below.
 

<pre><span style="background-color:#000000"> snagarju@snagarju </span><span style="background-color:#3464A3"><font color="#000000"> ~/go/src/external-dns-operator </font></span><span style="background-color:#C3A000"><font color="#3464A3"></font></span><span style="background-color:#C3A000"><font color="#000000">  NE-741/e2e_Azure ● </font></span><font color="#C3A000"> </font>make verify
/home/snagarju/go/src/external-dns-operator/bin/golangci-lint run -c .golangci.yaml --deadline=30m
hack/verify-gofmt.sh
hack/verify-deps.sh
hack/verify-generated.sh
<span style="background-color:#000000"> snagarju@snagarju </span><span style="background-color:#3464A3"><font color="#000000"> ~/go/src/external-dns-operator </font></span><span style="background-color:#C3A000"><font color="#3464A3"></font></span><span style="background-color:#C3A000"><font color="#000000">  NE-741/e2e_Azure ● </font></span><font color="#C3A000"> </font>make test-e2e 
go test \
-ldflags &quot;-X github.com/openshift/external-dns-operator/pkg/version.SHORTCOMMIT=3bde9a7 -X github.com/openshift/external-dns-operator/pkg/version.COMMIT=3bde9a7a8824611cc458c02a2f6b8bd3076f16be&quot; \
-timeout 1h \
-count 1 \
-v \
-tags e2e \
-run &quot;&quot; \
./test/e2e
I1125 11:32:10.929496 1867076 request.go:665] Waited for 1.04828778s due to client-side throttling, not priority and fairness, request: GET:https://api.ci-ln-nfrym32-1d09d.ci.azure.devcluster.openshift.com:6443/apis/machine.openshift.io/v1beta1?timeout=32s
ensuring hosted zone: 3bde9a7.example-test.info
=== RUN   TestOperatorAvailable
--- PASS: TestOperatorAvailable (0.25s)
=== RUN   TestExternalDNSRecordLifecycle
    operator_test.go:242: looking for DNS record in nameserver: ns1-02.azure-dns.com.
    operator_test.go:250: waiting for dns record
    operator_test.go:250: waiting for dns record
    operator_test.go:250: waiting for dns record
--- PASS: TestExternalDNSRecordLifecycle (81.60s)
PASS
ok  	github.com/openshift/external-dns-operator/test/e2e	90.844s
</pre>

test/e2e/operator_test.go
alebedev87: After the discussion: the symbol may be not there when the test is run from IDE or for debugging. Let's do this check the, thanks @Connect2naga!

test/e2e/operator_test.go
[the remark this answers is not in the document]
alebedev87: Yes, it would fail if you used go1.17 on local and added this "new" tag because CI would try to build/test with go1.16 which would remove this tag and create `git diff`. `make verify` will verify for any unsolicited (not in PR) diffs.

test/e2e/azure.go
Connect2naga: removed these entris

test/e2e/azure.go
Connect2naga: yes @alebedev87 , each provider having different SDK,
in aws we have to verify the zone exist or not if exist we have to use else create.
where are in Azure, it will be CreateOrUpdate, SDK iteslf will taken care the duplicates

test/e2e/azure.go
Connect2naga: again it is based on the provider,
aws expects the zoneID, and azure expects zonename.
in operator_test, we are passing zoneId. 

there is refactoring required, i am working on it, once this PR closed i will push the code.

test/e2e/azure.go
Connect2naga: ohh, didnt see the comment section, will take out now.

test/e2e/azure.go
alebedev87: Indeed, the zone ID is passed, the name of the argument of the method is wrong then, it should be something like `id` or `zoneID`.
Let's keep this intra-method context of `zoneName` and refactor in the follow ups.

test/e2e/azure.go
```diff
@@ -0,0 +1,171 @@
...
+	z, err := a.zoneClient.CreateOrUpdate(context.TODO(), a.config.ResourceGroup, rootDomain,
+		dns.Zone{Location: &location}, "", "")
+	if err != nil {
+		return "", []string{}, err
+	}
+	var zoneID string
+	zoneID = *z.ID
+	nameservers := append(*z.ZoneProperties.NameServers)
+	a.zoneName = rootDomain
+	return zoneID, nameservers, nil
```
alebedev87, line 134:
```suggestion
	a.zoneName = rootDomain
	return *z.ID, *z.ZoneProperties.NameServers, nil
```

PR #73, opened 2021-11-24
A sample CR for external dns operator with source route in OpenShift enviroment on GCP..

miheer:
A sample CR for external dns operator with source route.

### DIFF
--- a/config/samples/gcp/kustomization.yaml
+++ b/config/samples/gcp/kustomization.yaml
@@ -1,2 +1,3 @@
 resources:
+- operator_v1alpha1_externaldns_openshift.yaml
 - operator_v1alpha1_externaldns_detailed.yaml
--- a/config/samples/gcp/operator_v1alpha1_externaldns_openshift.yaml
+++ b/config/samples/gcp/operator_v1alpha1_externaldns_openshift.yaml
@@ -1,0 +1,23 @@
+apiVersion: externaldns.olm.openshift.io/v1alpha1
+kind: ExternalDNS
+metadata:
+  name: sample-gcp
+spec:
+  domains:
+  - filterType: Include
+    matchType: Exact
+    name: test-gcp1.qe.gcp.devcluster.openshift.com
+  provider:
+    type: GCP
+  source:
+    # In case you have multiple ingress controllers you must specify ingress controller name in the routerName
+    # so that the external dns will use the router canonical name correrponding to it to create a dns record. 
+    openshiftRouteOptions:  
+      routerName: default
+    # Source Type is route resource of OpenShift    
+    type: OpenShiftRoute
+    # It's not possible to not have FQDNTemplate if the hostname annotation policy is ignore.
+    fqdnTemplate: 
+    - "{{.Name}}.apps.test-gcp1.qe.gcp.devcluster.openshift.com"
+  zones:
+  - test-gcp1-q6m5v-private-zone


config/samples/gcp/gcp_openshift_route.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: sample-gcp
+spec:
+  domains:
+  - filterType: Include
+    matchType: Exact
+    name: misalunkgcp.qe.gcp.devcluster.openshift.com
+  provider:
+    type: GCP
+  source:
+    hostnameAnnotation: Allow
```
alebedev87, line 13: Hostname annotation is not really necessary for OpenShift routes and it's not the most secure option so it's better to know show it as a recommended sample.

config/samples/gcp/gcp_openshift_route.yaml
```diff
@@ -0,0 +1,18 @@
...
+    openshiftRouteOptions:
```
alebedev87, line 14: Can you please add a comment explaining why we need this field? This would be useful for the users.

config/samples/gcp/gcp_openshift_route.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: externaldns.olm.openshift.io/v1alpha1
+kind: ExternalDNS
+metadata:
+  name: sample-gcp
+spec:
+  domains:
+  - filterType: Include
+    matchType: Exact
+    name: misalunkgcp.qe.gcp.devcluster.openshift.com
```
alebedev87, line 9: Can you put a neutral domain name (not engineer name I mean)?

config/samples/gcp/gcp_openshift_route.yaml
```diff
@@ -0,0 +1,18 @@
...
+  provider:
+    type: GCP
+  source:
+    hostnameAnnotation: Allow
+    openshiftRouteOptions:
+      routerName: default
+    type: OpenShiftRoute
+  zones:
+  - <zone name>
```
alebedev87, line 18: If you put a domain name similar to a real one, let's put a zone ID (for GCP it's zone ID not name) similar to a real one too, not just a placeholder.
And can you add a new line at the end of the line please?

config/samples/gcp/gcp_openshift_route.yaml
```diff
@@ -0,0 +1,22 @@
...
+    type: GCP
+  source:
+    # Incase you have multiple ingress controllers you must specify ingress controller name in the routerName
+    # so that the external dns will use the router canonical name correrponding to it to create a dns record. 
+    openshiftRouteOptions:  
+      routerName: default
+    #Source Type is route resource of OpenShift    
+    type: OpenShiftRoute
+    fqdnTemplate: 
+    - "{{.Name}}.apps.test-gcp1.qe.gcp.devcluster.openshift.com"
```
alebedev87, line 20: Can this ever work in the domain filter is the exact match of `test-gcp1.qe.gcp.devcluster.openshift.com` ? I believe that `app` subdpmaon should be omitted.

config/samples/gcp/gcp_openshift_route.yaml
```diff
@@ -0,0 +1,22 @@
...
+  zones:
+  - test-gcp1-q6m5v-private-zone
```
alebedev87, line 22: Sorry for being a bit too pedantical but GCP managed zone ID looks more like this `3651032588905568971`.

config/samples/gcp/gcp_openshift_route.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: sample-gcp
+spec:
+  domains:
+  - filterType: Include
+    matchType: Exact
+    name: test-gcp1.qe.gcp.devcluster.openshift.com
+  provider:
+    type: GCP
+  source:
+    # Incase you have multiple ingress controllers you must specify ingress controller name in the routerName
```
alebedev87, line 13:
```suggestion
    # In case you have multiple ingress controllers you must specify ingress controller name in the routerName
```

config/samples/gcp/gcp_openshift_route.yaml
miheer: You can put the name which is in the output of ->
[miheer@localhost Downloads]$ gcloud dns managed-zones list | grep mykrbid.qe.gcp
mykrbid-xdl4p-private-zone            mykrbid.qe.gcp.devcluster.openshift.com.            Created By OpenShift Installer  private
[miheer@localhost Downloads]$

config/samples/gcp/gcp_openshift_route.yaml
miheer: Not sure what you mean. I have used CR from real env. test-gcp1.qe.gcp.devcluster.openshift.com  Exact means apps.test-gcp1.qe.gcp.devcluster.openshift.com also works. You can prepend that with apps or anything. Putting apps is important as the ingress controller creates a wildcard for *.apps .

miheer: Also in real case fqdnTemplate:  is useless as hostname will always be available in the route as we have a check openshift api server for it's existence.  I had to add it as webhook does not allow HostNameAnnotation Ignore with no fqdnTemplate.

alebedev87: Yes, you are right, `FQDNTemplate` not really needed for the OpenShift case. Let's remove it then to showcase this.

config/samples/gcp/gcp_openshift_route.yaml
alebedev87: Are you sure it'll work with the name of the managed zone? This value will be passed to `--zone-id-filter` flag.

config/samples/gcp/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
 - operator_v1alpha1_externaldns_detailed.yaml
+- gcp_openshift_route.yaml
```
alebedev87, line 3:
```suggestion
- operator_v1alpha1_externaldns_openshift.yaml
```

config/samples/gcp/gcp_openshift_route.yaml
alebedev87: Yes, you are right, the validation in the operator mirrors the same validation in external-dns: no matter the source - it's not possible to not have `FQDNTemplate` if the hostname annotatin policy is ignore.

config/samples/gcp/gcp_openshift_route.yaml
miheer: Yes from my testing

alebedev87: That's a bit confusing for the flag of this name but it's true: the filtering can be done using the name too: [link](https://github.com/kubernetes-sigs/external-dns/blob/master/provider/google/google.go#L180).

PR #74, opened 2021-11-24
Sample CR for external dns operator to work on Azure in Openshift.

miheer:
Sample CR for external dns operator to work on Azure in Openshift.

### DIFF
--- a/config/samples/azure/kustomization.yaml
+++ b/config/samples/azure/kustomization.yaml
@@ -1,2 +1,3 @@
 resources:
+- operator_v1alpha1_externaldns_openshift.yaml
 - operator_v1alpha1_externaldns_detailed.yaml
--- a/config/samples/azure/operator_v1alpha1_externaldns_openshift.yaml
+++ b/config/samples/azure/operator_v1alpha1_externaldns_openshift.yaml
@@ -1,0 +1,23 @@
+apiVersion: externaldns.olm.openshift.io/v1alpha1
+kind: ExternalDNS
+metadata:
+  name: sample-azure
+spec:
+  domains:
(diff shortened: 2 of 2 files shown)


config/samples/azure/azure_openshift_route.yaml
```diff
@@ -0,0 +1,23 @@
...
+    matchType: Exact
+    name: test-azure1.qe.azure.devcluster.openshift.com
+  provider:
+    type: Azure
+  source:
+    # Incase you have multiple ingress controllers you must specify ingress controller name in the routerName
+    # so that the external dns will use the router canonical name correrponding to it to create a dns record.
+    openshiftRouteOptions:
+      routerName: default
+    # Source Type is route resource of OpenShift
```
alebedev87, line 17: Where is `type` field?

config/samples/azure/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
+- azure_openshift_route.yaml
```
alebedev87, line 2: I'm fine with any name but I would really prefer the things to be consistent.
The other file in this directory is called `operator_v1alpha1_externaldns_detailed.yaml`. BTW it's the kubebuilder pattern for the samples.

config/samples/azure/azure_openshift_route.yaml
miheer: line 21

miheer: anyway I made the changes

miheer:
@alebedev87 PTAL. Thanks!

PR #82, opened 2021-11-29
CFE-192: Add ocp route test

Elbehery:
This PR add a test for OpenShift Route to be discovered by External-DNS


test/e2e/util.go
```diff
@@ -158,6 +174,30 @@ func defaultExternalDNS(name, zoneID, zoneDomain string) operatorv1alpha1.Extern
...
+					Type: operatorv1alpha1.SourceTypeRoute,
+					OpenShiftRoute: &operatorv1alpha1.ExternalDNSOpenShiftRouteOptions{
+						RouterName: routeName,
+					},
+					AnnotationFilter: map[string]string{
+						"external-dns.mydomain.org/publish": "yes",
+					},
+				},
+				HostnameAnnotationPolicy: "Ignore",
+				FQDNTemplate:             []string{fmt.Sprintf("{{.Name}}.%s", zoneDomain)},
```
alebedev87, line 195: Not really needed for `Route` source.

test/e2e/util.go
```diff
@@ -158,6 +174,30 @@ func defaultExternalDNS(name, zoneID, zoneDomain string) operatorv1alpha1.Extern
...
+			Name: name,
+		},
+		Spec: operatorv1alpha1.ExternalDNSSpec{
+			Zones: []string{zoneID},
+			Source: operatorv1alpha1.ExternalDNSSource{
+				ExternalDNSSourceUnion: operatorv1alpha1.ExternalDNSSourceUnion{
+					Type: operatorv1alpha1.SourceTypeRoute,
+					OpenShiftRoute: &operatorv1alpha1.ExternalDNSOpenShiftRouteOptions{
+						RouterName: routeName,
+					},
```
alebedev87, line 189: This can be removed for the moment as this functionality is not yet completed.

test/e2e/util.go
```diff
@@ -58,6 +61,19 @@ func clusterIPService(name, namespace string) *corev1.Service {
 	return testService(name, namespace, corev1.ServiceTypeClusterIP)
 }
 
+func testRoute(name, namespace string) *routev1.Route {
+	return &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:   namespace,
+			Name:        name,
+			Annotations: map[string]string{},
```
alebedev87, line 69: Annotation has to be the same as one from the annotation filter of ExternalDNS instance.

test/e2e/util.go
```diff
@@ -30,6 +32,7 @@ type providerTestHelper interface {
 	platform() string
 	makeCredentialsSecret(namespace string) *corev1.Secret
 	buildExternalDNS(name, zoneID, zoneDomain string, credsSecret *corev1.Secret) operatorv1alpha1.ExternalDNS
+	buildRouteExternalDNS(name, routeName, zoneID, zoneDomain string) operatorv1alpha1.ExternalDNS
```
alebedev87, line 35:
```suggestion
	buildOpenShiftExternalDNS(name, routeName, zoneID, zoneDomain string) operatorv1alpha1.ExternalDNS
```
As it's not only about the route source.

test/e2e/util.go
```diff
@@ -58,6 +61,19 @@ func clusterIPService(name, namespace string) *corev1.Service {
...
 
+func testRoute(name, namespace string) *routev1.Route {
+	return &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:   namespace,
+			Name:        name,
+			Annotations: map[string]string{},
+		},
+		Spec: routev1.RouteSpec{
+			Host: "my-domain.com",
```
alebedev87, line 72: This can be passed as parameter too. Typically the hosts of routes are of the format: `<given-hostname>.<base-cluster-domain>`. In the test you can pass the host which is made of any test name + the domain of just created zone.

alebedev87 (commented):
Just a first quick look. I agree with the direction, some comments left though.

Elbehery:
> Noticed that e2e-gcp-operator is failing - (from log)
> 
> `Failed to create external DNS "test-extdns": admission webhook "vexternaldns.kb.io" denied the request: "fqdnTemplate" must be specified when "hostnameAnnotation" is "Ignore"`
> 
> @Elbehery Can you have a look at why its failing ?

@lmzuccarelli  I have adjusted the webhook 👍🏽 

@alebedev87  thanks a lot for your help 🙏🏽

api/v1alpha1/externaldns_webhook.go
```diff
@@ -134,6 +134,9 @@ func (r *ExternalDNS) validateFilters() error {
 }
 
 func (r *ExternalDNS) validateHostnameAnnotationPolicy() error {
+	if r.Spec.Source.Type == SourceTypeRoute {
+		return nil
+	}
```
lmzuccarelli, line 139: This fixes the e2e-gcp tests ?

Elbehery: yes, but actually I should revert this and all the tests .. There is a validationWebhook tests for this, and I should have kept the FQDN template 

fix on the way

test/e2e/util.go
alebedev87: Actually it is :(

test/e2e/operator_test.go
```diff
@@ -152,6 +158,89 @@ func TestOperatorAvailable(t *testing.T) {
...
+
+	t.Log("Creating external dns instance with source type route")
+	extDNS := helper.buildOpenShiftExternalDNS(testExtDNSName, hostedZoneID, hostedZoneDomain)
+	if err := kubeClient.Create(context.TODO(), &extDNS); err != nil {
+		t.Fatalf("Failed to create external DNS %q: %v", testExtDNSName, err)
+	}
+	defer kubeClient.Delete(context.TODO(), &extDNS)
+
+	// create a service of type ClusterIP without annotation
+	t.Log("Creating test route service")
```
alebedev87, line 176: Do you really need the service? When I tried the route creation I put a service name which didn't exist and OCP admitted it. We'll not have the router configured properly without the service, so no e2e connectivity but we don't test this part, only the resolution of the router's IP.

test/e2e/operator_test.go
```diff
@@ -152,6 +158,89 @@ func TestOperatorAvailable(t *testing.T) {
...
+
+	// create a route with the annotation targeted by the ExternalDNS resource
+	t.Log("Creating source route")
+	route := testRoute(testRouteName, testNamespace, testRouteHost, testServiceName)
+	if err := kubeClient.Create(context.Background(), route); err != nil {
+		t.Fatalf("Failed to create test route %s/%s: %v", testNamespace, testRouteName, err)
+	}
+	defer kubeClient.Delete(context.TODO(), route)
+
+	// get the route canonical name
```
alebedev87, line 191:
```suggestion
	// get the router canonical name
```

test/e2e/operator_test.go
Elbehery: ok, it was failing on the schema registration, i thought we need the service, will remove it now 👍🏽

test/e2e/operator_test.go
```diff
@@ -152,6 +158,89 @@ func TestOperatorAvailable(t *testing.T) {
...
+	var targetRoute routev1.Route
+	err = kubeClient.Get(context.TODO(), types.NamespacedName{
+		Namespace: testNamespace,
+		Name:      testRouteName,
+	}, &targetRoute)
+	if err != nil {
+		t.Fatalf("Failed to retrieve the created route %s/%s: %v", testNamespace, testRouteName, err)
+	}
+	targetRouteCName := targetRoute.Status.Ingress[0].RouterCanonicalHostname
```
alebedev87, line 200:
```suggestion
	targetRouterCName := targetRoute.Status.Ingress[0].RouterCanonicalHostname
```

test/e2e/operator_test.go
```diff
@@ -152,6 +158,89 @@ func TestOperatorAvailable(t *testing.T) {
...
+	if targetRouteCName == "" {
+		t.Fatalf("Created route's canonical name is empty %v", err)
+	}
+	// get the route IPs by looking up the route canonical name
+	routeIPs := make(map[string]struct{})
```
alebedev87, line 205:
```suggestion
	routerExpectedIPs := make(map[string]struct{})
```

test/e2e/operator_test.go
```diff
@@ -152,6 +158,89 @@ func TestOperatorAvailable(t *testing.T) {
...
+	routeExpectedIPs, err := customResolver("").LookupIP(context.TODO(), "ip", targetRouteCName)
```
alebedev87, line 206:
```suggestion
	routerIPs, err := customResolver("").LookupIP(context.TODO(), "ip", targetRouteCName)
```

test/e2e/operator_test.go
```diff
@@ -152,6 +158,89 @@ func TestOperatorAvailable(t *testing.T) {
...
+					return false, nil
+				}
+			}
+			return true, nil
+		}); err != nil {
+			t.Logf("Failed to verify that DNS has been correctly set.")
+		} else {
+			return
+		}
+	}
```
alebedev87, line 240: This part seems to be repeated 1-1 from the service test case. Maybe it's worth putting this into `utils.go`.

test/e2e/operator_test.go
```diff
@@ -32,6 +33,8 @@ const (
 	baseZoneDomain     = "example-test.info"
 	testNamespace      = "external-dns-test"
 	testServiceName    = "test-service"
+	testRouteName      = "test-route"
+	testRouteHost      = "myroute.example.com"
```
alebedev87, line 37: I think that using a public domain for the test may be not good. Even if `myroute.example.com` doesn't seem to exist but someone can own it at some point and then it'll be always resolvable, so the test will always pass.
I'd prefer the route's host to be from the test domain you create for your test: `myroute.<commit-id>.example-test.info`.

alebedev87:
Taken over by https://github.com/openshift/external-dns-operator/pull/88

/close

PR #83, opened 2021-11-29
Avoid using global variables for configuration

alebedev87:
**Current situation**
The setup of the webhook checks for OpenShift API to set a global `api/v1alpha1` package variable which indicates whether it's openshift or not. In the same place the infrastructure config is retrieved and and another global variable for `PlatformStatus` is set in the same package. The rest of the code uses this to switchcase between pure Kubernetes and OpenShift installations.

_Problem 1_: platform detection and status have no logical adherence to the API wehbook. This is something which has to be done once but the webhook is not optimal place for this. 
_Problem 2_: testing with global variables is more difficult as the functions depend on external variants which are not easy to set and even more difficult to be reset to the previous values.

**Proposal**
Move the platform specific things into the operator's config: `pkg/operator/config#Config`. The platform gets detected in main (only Kube client is needed), the platforms status gets retrieved at the operator initialization (we already know the platform and can use the specific client). Then these data is propagated to the places where it's needed.

_Result_: the things which have to be done once are not bound to the webhook setup, fewer dependencies on the global variables. Slightly (a couple of %) better code coverage.
 
 
 **Bonus**
 Discovered by e2e testing of the proposal: credentials secret specified in `ExternalDNS` CR is not used when the platform is OpenShift. Fix this to give the priority to explicitly specified secret.

### DIFF
--- a/api/v1alpha1/externaldns_webhook.go
+++ b/api/v1alpha1/externaldns_webhook.go
@@ -17,19 +17,11 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"context"
 	"errors"
 	"fmt"
-
-	configv1 "github.com/openshift/api/config/v1"
-
-	"k8s.io/apimachinery/pkg/types"
-
 	"regexp"
 
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/kubernetes"
 
 	utilErrors "k8s.io/apimachinery/pkg/util/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -37,40 +29,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
-const (
-	kind    = "OpenShiftAPIServer"
-	group   = "operator.openshift.io"
-	version = "v1"
-)
-
 // webhookLog is for logging in this package.
 var webhookLog = logf.Log.WithName("validating-webhook")
 
-var IsOpenShift bool
+var isOpenShift bool
 
-var PlatformStatus *configv1.PlatformStatus
-
-func (r *ExternalDNS) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	kubeClient, err := kubernetes.NewForConfig(mgr.GetConfig())
-	if err != nil {
-		return err
-	}
-	IsOpenShift = IsOCP(kubeClient)
-
-	if IsOpenShift {
-
-		var err error
-		infraConfig := &configv1.Infrastructure{}
-		if err = mgr.GetClient().Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
-			return fmt.Errorf("failed to get infrastructure 'config': %v", err)
-		}
-
-		PlatformStatus = infraConfig.Status.PlatformStatus
-	}
-
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(r).
-		Complete()
+func (r *ExternalDNS) SetupWebhookWithManager(mgr ctrl.Manager, openshift bool) error {
+	isOpenShift = openshift
+	return ctrl.NewWebhookManagedBy(mgr).For(r).Complete()
 }
 
 //+kubebuilder:webhook:path=/validate-externaldns-olm-openshift-io-v1alpha1-externaldns,mutating=false,failurePolicy=fail,sideEffects=None,groups=externaldns.olm.openshift.io,resources=externaldnses,verbs=create;update,versions=v1alpha1,name=vexternaldns.kb.io,admissionReviewVersions={v1,v1beta1}
@@ -141,7 +107,7 @@ func (r *ExternalDNS) validateHostnameAnnotationPolicy() error {
 }
 
 func (r *ExternalDNS) validateProviderCredentials() error {
-	if IsOpenShift && (r.Spec.Provider.Type == ProviderTypeAWS || r.Spec.Provider.Type == ProviderTypeGCP || r.Spec.Provider.Type == ProviderTypeAzure) {
+	if isOpenShift && (r.Spec.Provider.Type == ProviderTypeAWS || r.Spec.Provider.Type == ProviderTypeGCP || r.Spec.Provider.Type == ProviderTypeAzure) {
 		return nil
 	}
 	provider := r.Spec.Provider
@@ -168,22 +134,4 @@ func (r *ExternalDNS) validateProviderCredentials() error {
 		}
 	}
 	return nil
-}
-
-// Returns true if platform is OCP
-func IsOCP(kubeClient discovery.DiscoveryInterface) bool {
-	// Since, CRD for OpenShift API Server was introduced in OCP v4.x we can verify if the current cluster is on OCP v4.x by
-	// ensuring that resource exists against Group(operator.openshift.io), Version(v1) and Kind(OpenShiftAPIServer)
-	// In case it doesn't exist we assume that external dns is running on non OCP 4.x environment
-	resources, err := kubeClient.ServerResourcesForGroupVersion(group + "/" + version)
-	if err != nil {
-		return false
-	}
-
-	for _, apiResource := range resources.APIResources {
-		if apiResource.Kind == kind {
-			return true
-		}
-	}
-	return false
 }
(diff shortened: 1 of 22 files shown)


pkg/operator/controller/externaldns/pod.go
```diff
@@ -345,8 +335,9 @@ func (b *externalDNSContainerBuilder) fillGCPFields(container *corev1.Container)
 			container.Args = append(container.Args, fmt.Sprintf("--google-project=%s", *b.externalDNS.Spec.Provider.GCP.Project))
 		}
 	} else {
-		container.Args = append(container.Args, fmt.Sprintf("--google-project=%s", operatorv1alpha1.PlatformStatus.GCP.ProjectID))
+		container.Args = append(container.Args, fmt.Sprintf("--google-project=%s", b.platformStatus.GCP.ProjectID))
```
Miciah, line 338: This _should_ work because `platformStatus` _should_ always be set on a GCP cluster (`platformStatus` wasn't set on 4.1 clusters, but 4.1 couldn't be installed on GCP, so it should be impossible to have a GCP cluster that was upgraded from 4.1).  However, it is a pointer type, and the value is technically optional, so I'd still recommend having a nil check to be safe.

api/v1alpha1/externaldns_utils.go
```diff
@@ -0,0 +1,26 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+// PublicProvider returns true if the dns provider is one of supported public providers
+func (e *ExternalDNS) PublicProvider() bool {
```
Miciah, line 20: Is "Public" significant here, or would it be reasonable to name this method something like `ValidProvider` or `SupportedProvider`?

Miciah (commented):
Very nice cleanup!  A couple of minor suggestions, but overall
/approve

api/v1alpha1/externaldns_utils.go
alebedev87: No, not significant, I just didn't come up with a better name. The idea of this method is to tell that "if you are in the OpenShift case you need to do something only if it's one of these providers". Because we don't leverage OpenShift features for `BlueCat` and `InfoBlox`.

`Supported`, I don't know, this may give an impression that `BlueCat` and `InfoBlox` are not supported. The same can be thought about `Valid`. WDYT? Do you have any other ideas?

pkg/operator/controller/externaldns/pod.go
alebedev87: Done

alebedev87:
aws e2e test failed: multiple nameservers failed to find the record. Intermittent?

/retest

alebedev87:
AWS e2e failure is due to the operand POD which didn't find a key in the credentials secret:
```yaml
                "containerStatuses": [
                    {
                        "image": "k8s.gcr.io/external-dns/external-dns:v0.8.0",
                        "imageID": "",
                        "lastState": {},
                        "name": "external-dns-n587hc8h58fh55cq",
                        "ready": false,
                        "restartCount": 0,
                        "started": false,
                        "state": {
                            "waiting": {
                                "message": "couldn't find key aws_access_key_id in Secret external-dns/external-dns-credentials-test-extdns",
                                "reason": "CreateContainerConfigError"
                            }
                        }
                    }
                ],
```

miheer:
@alebedev87 please make sure you test this on Azure, AWS and GCP.

miheer:
> AWS e2e failure is due to the operand POD which didn't find a key in the credentials secret:
> 
> ```yaml
>                 "containerStatuses": [
>                     {
>                         "image": "k8s.gcr.io/external-dns/external-dns:v0.8.0",
>                         "imageID": "",
>                         "lastState": {},
>                         "name": "external-dns-n587hc8h58fh55cq",
>                         "ready": false,
>                         "restartCount": 0,
>                         "started": false,
>                         "state": {
>                             "waiting": {
>                                 "message": "couldn't find key aws_access_key_id in Secret external-dns/external-dns-credentials-test-extdns",
>                                 "reason": "CreateContainerConfigError"
>                             }
>                         }
>                     }
>                 ],
> ```

Yes please test the code manually on AWS.

miheer:
You need to make changes here

https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/externaldns/controller.go#L108

miheer:
Also here

https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/credentials-secret/controller.go#L256

miheer:
I think you have made the above changes but it looks somewhere the bool is not getting set properly which is not creating the secret properly.

miheer:
I will really recommend to test this manually on AWS, Azure and GCP.

api/v1alpha1/externaldns_utils.go
arjunrn: Maybe `AutoCredentialConfigProvider()` if the cloud provider is supported for automatic credential configuration.

api/v1alpha1/externaldns_webhook.go
```diff
@@ -17,60 +17,26 @@ limitations under the License.
...
-			return fmt.Errorf("failed to get infrastructure 'config': %v", err)
-		}
-
-		PlatformStatus = infraConfig.Status.PlatformStatus
-	}
-
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(r).
-		Complete()
+func (r *ExternalDNS) SetupWebhookWithManager(mgr ctrl.Manager, openshift bool) error {
```
arjunrn, line 37: It may be time to end this hacky way in which kubebuilder insists on registering the webhook. In a PR once this is merged I think I would like to implement a struct which embeds `ExternalDNS` and has the `isOpenshift` field. This way we can avoid the package level variable here as well.

api/v1alpha1/externaldns_utils.go
alebedev87: @arjunrn: sorry, didn't see your comment. I called it `ManagedCredentialsProvider()`. Also I moved it outside the api package to remove the unnecessary method from the "public" API.

alebedev87:
@miheer : there is no need to test the code manually on any provider. CCO logic is already tested by the e2e tests. That's actually why I got the error which I fixed in the commit that followed the comment: because I modified `desiredCredentialsSecret` function which was used in the CCO related flow.

alebedev87:
Added the change to not use the CCO credentials if they are explicitly specified in CR.

Use case when credentials are taken from CCO will be covered in https://github.com/openshift/external-dns-operator/pull/82.

pkg/operator/config/config.go
```diff
@@ -55,16 +72,70 @@ type Config struct {
...
 	// EnableWebhook is the flag indicating if the webhook server should be started.
 	EnableWebhook bool
+
+	// EnablePlatformDetection is the flag indicating if the operator needs to detect on which platform it runs
+	EnablePlatformDetection bool
+
+	// IsOpenShift is the flag indicating that the operator runs in OpenShift cluster
+	IsOpenShift bool
+
+	// PlatformStatus is the details about the undelying platform
```
Miciah, line 82:
```suggestion
	// PlatformStatus is the details about the underlying platform.
```

pkg/operator/config/config.go
```diff
@@ -17,17 +17,34 @@ limitations under the License.
...
-	DefaultOperatorNamespace = "external-dns-operator"
-	DefaultOperandNamespace  = "external-dns"
-	DefaultEnableWebhook     = true
+	DefaultExternalDNSImage        = "k8s.gcr.io/external-dns/external-dns:v0.8.0"
+	DefaultMetricsAddr             = "127.0.0.1:8080"
+	DefaultOperatorNamespace       = "external-dns-operator"
+	DefaultOperandNamespace        = "external-dns"
+	DefaultEnableWebhook           = true
+	DefaultEnablePlatformDetection = true
+	DefaultIsOpenShift             = false
```
Miciah, line 42: `DefaultIsOpenShift` can be removed too along with the `operatorconfig.New` function.

pkg/utils/utils.go
```diff
@@ -0,0 +1,26 @@
+/*
+You may obtain a copy of the License at
```
Miciah, line 2: Looks like the licence text was truncated.  
```suggestion
/*
Copyright 2021.

Licensed under the Apache License, Version 2.0 (the "License");
you may not use this file except in compliance with the License.
You may obtain a copy of the License at
```

PR #85, opened 2021-12-02
Use ubi for container images

dbaker-rh:
Leverage UBI to build and run the container image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,7 @@ COPY . .
 # Build
 RUN make build-operator
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
+# Use minimal base image to package the manager binary
 FROM registry.access.redhat.com/ubi8/ubi-micro:latest
 WORKDIR /
 COPY --from=builder /opt/app-root/src/bin/external-dns-operator .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Build the manager binary
-FROM golang:1.16 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset:latest as builder
 
-WORKDIR /workspace
+WORKDIR /opt/app-root/src
 COPY . .
 
 # Build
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,17 +1,17 @@
 # Build the manager binary
-FROM golang:1.16 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset:latest as builder
```
frobware, line 2: Shouldn't this be similar (if not identical) to the way we build (for example) the cluster-ingress-operator?

```
FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.10 AS builder
```

Taken from: https://github.com/openshift/cluster-ingress-operator/commit/c22568aac5be683cd0dbf1eba43522a9645455ca

dbaker-rh: This Dockerfile is for end-users building manually (e.g. in the README.md manual steps for "make image").

In this repo, there is Dockerfile.openshift which tracks that change for the 4.9 ... 4.10 changes, and which uses the ART images.

frobware:
/lgtm

based on the understanding that `Dockerfile.openshift` is for OpenShift CI and this addition doesn't affect CI builds. Also seems like a ~200MB image size saving. Makes me wonder why `Dockerfile.openshift` cannot produce a similar size image.

Dockerfile
```diff
@@ -1,17 +1,17 @@
...
 
-WORKDIR /workspace
+WORKDIR /opt/app-root/src
 COPY . .
 
 # Build
 RUN make build-operator
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
```
dbaker-rh: I just re-reviewed, and see this -- I need to update my commit to remove this comment now we're using UBI instead.

dbaker-rh:
@frobware - can you re-lgtm?  I pushed a new commit to fix the comments.

frobware:
> @frobware - can you re-lgtm? I pushed a new commit to fix the comments.

Sure, but you may as well squash first.

PR #89, opened 2021-12-03
Bug 2027281: Update to ensure private zone provider is selected for Azure Cloud

lmzuccarelli:
This PR addresses the selection of  the private zone provider when the user creates an ExternalDNS CR in Azure for private zone deployment

### DIFF
--- a/pkg/operator/controller/externaldns/deployment.go
+++ b/pkg/operator/controller/externaldns/deployment.go
@@ -18,7 +18,6 @@ package externaldnscontroller
 
 import (
 	"context"
-	"strings"
 
 	"fmt"
 	"sort"
@@ -50,7 +49,7 @@ const (
 	masterNodeRoleLabel                 = "node-role.kubernetes.io/master"
 	osLabel                             = "kubernetes.io/os"
 	linuxOS                             = "linux"
-	privateDNSZones                     = "privatednszones"
+	azurePrivateDNSZonesResourceSubStr  = "privatednszones"
 )
 
 // providerStringTable maps ExternalDNSProviderType values from the
@@ -185,7 +184,7 @@ func desiredExternalDNSDeployment(namespace, image, secretName string, serviceAc
 
 	cbld := &externalDNSContainerBuilder{
 		image:          image,
-		Provider:       provider,
+		provider:       provider,
 		source:         source,
 		secretName:     secretName,
 		volumes:        volumes,
@@ -193,22 +192,25 @@ func desiredExternalDNSDeployment(namespace, image, secretName string, serviceAc
 		isOpenShift:    isOpenShift,
 		platformStatus: platformStatus,
 	}
+
 	if len(externalDNS.Spec.Zones) == 0 {
-		container, err := cbld.build("")
-		if err != nil {
-			return nil, fmt.Errorf("failed to build container: %w", err)
+		// an empty list means publish to all zones
+		// this is a special case for Azure
+		// both public and private zones will need to be published to
+		providerList := []string{provider}
+		if cbld.provider == externalDNSProviderTypeAzure {
+			providerList = append(providerList, externalDNSProviderTypeAzurePrivate)
 		}
-		depl.Spec.Template.Spec.Containers = []corev1.Container{*container}
+		for _, p := range providerList {
+			cbld.provider = p
+			container, err := cbld.build("")
+			if err != nil {
+				return nil, fmt.Errorf("failed to build container: %w", err)
+			}
+			depl.Spec.Template.Spec.Containers = append(depl.Spec.Template.Spec.Containers, *container)
+		}
 	} else {
 		for _, zone := range externalDNS.Spec.Zones {
-			// check each zone for the keyword 'privatednszones', if the provider is Azure
-			// this ensures that the provider 'azure-private-dns' is passed to the container
-			// to set the operand provider correctly
-			if cbld.Provider == externalDNSProviderTypeAzure {
-				if strings.Contains(strings.ToLower(zone), privateDNSZones) {
-					cbld.Provider = externalDNSProviderTypeAzurePrivate
-				}
-			}
 			container, err := cbld.build(zone)
 			if err != nil {
 				return nil, fmt.Errorf("failed to build container for zone %s: %w", zone, err)
--- a/pkg/operator/controller/externaldns/deployment.go
+++ b/pkg/operator/controller/externaldns/deployment.go
@@ -18,6 +18,7 @@ package externaldnscontroller
 
 import (
 	"context"
+	"strings"
 
 	"fmt"
 	"sort"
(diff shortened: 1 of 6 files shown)


miheer:
/title Bug 2027281:  Update to ensure private zone provider is selected for Azure Cloud

miheer:
https://bugzilla.redhat.com/show_bug.cgi?id=2027281  Please attach this bug in the title

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -198,6 +201,14 @@ func desiredExternalDNSDeployment(namespace, image, secretName string, serviceAc
 		depl.Spec.Template.Spec.Containers = []corev1.Container{*container}
 	} else {
```
alebedev87, line 202: ZoneID is an optional field of the API. So we are not guaranteed to have it in the CR which disables the guess of the Azure provider based on the resource path pattern.

CC @Miciah

Miciah: If an empty zones list means "publish in all zones", then we need two external-dns instances on Azure when the zones list is empty: one for public zones and one for private zones.  

We should also make sure that the godoc in the API is clear about the semantics of an empty zones list, namely that the operator configures external-dns to publish to all zones if the zones list is empty unless the operator is running on a platform on which the operator can look up a default set of zones, viz. OpenShift with its cluster DNS config.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -38,16 +39,18 @@ import (
...
+	externalDNSProviderTypeAzure        = "azure"
+	externalDNSProviderTypeAzurePrivate = "azure-private-dns"
+	externalDNSProviderTypeBlueCat      = "bluecat"
+	externalDNSProviderTypeInfoblox     = "infoblox"
+	appNameLabel                        = "app.kubernetes.io/name"
+	appInstanceLabel                    = "app.kubernetes.io/instance"
+	masterNodeRoleLabel                 = "node-role.kubernetes.io/master"
+	osLabel                             = "kubernetes.io/os"
+	linuxOS                             = "linux"
+	privateDNSZones                     = "privatednszones"
```
alebedev87, line 53:
```suggestion
	azurePrivateDNSZonesResourceSubStr                     = "privatednszones"
```
Just to make it more precise, WDYT?

pkg/operator/controller/externaldns/deployment.go
alebedev87: After the discussion with everybody: 

- If the zones list is empty, create one external-dns container for public and one for private.
- Else, create one external-dns container per zone, configured as public or private in accordance with the zone.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -78,7 +78,7 @@ const (
 // externalDNSContainerBuilder builds the definition of the containers for ExternalDNS POD
 type externalDNSContainerBuilder struct {
 	image          string
-	provider       string
+	Provider       string
```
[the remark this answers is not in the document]
lmzuccarelli: Yep fixed this. Left over from some experimentation

pkg/operator/controller/externaldns/deployment.go
lmzuccarelli: Done

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -190,14 +193,38 @@ func desiredExternalDNSDeployment(namespace, image, secretName string, serviceAc
...
 	} else {
 		for _, zone := range externalDNS.Spec.Zones {
+			// check each zone for the keyword 'privatednszones', if the provider is Azure
+			// this ensures that the provider 'azure-private-dns' is passed to the container
+			// to set the operand provider correctly
+			if cbld.provider == externalDNSProviderTypeAzure {
+				if strings.Contains(strings.ToLower(zone), azurePrivateDNSZonesResourceSubStr) {
+					cbld.provider = externalDNSProviderTypeAzurePrivate
+				}
+			}
```
alebedev87, line 227: This can be checked inside the container builder as the externalDNS instance is passed to it.
This way we would keep all the platform specific logic in the container builder.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -190,14 +193,38 @@ func desiredExternalDNSDeployment(namespace, image, secretName string, serviceAc
...
+				}
+				depl.Spec.Template.Spec.Containers = append(depl.Spec.Template.Spec.Containers, *container)
+			}
+		} else {
+			container, err := cbld.build("")
+			if err != nil {
+				return nil, fmt.Errorf("failed to build container: %w", err)
+			}
+			depl.Spec.Template.Spec.Containers = []corev1.Container{*container}
 		}
```
alebedev87, line 217:
```suggestion
                providerList := []string{provider}
		if cbld.provider == externalDNSProviderTypeAzure {
			providerList = append(providerList, externalDNSProviderTypeAzurePrivate)
		}
		for _, p := range providerList {
			cbld.provider = p
			container, err := cbld.build("")
			if err != nil {
				return nil, fmt.Errorf("failed to build container: %w", err)
			}
			depl.Spec.Template.Spec.Containers = append(depl.Spec.Template.Spec.Containers, *container)
		}
```
WDYT?

lmzuccarelli: Yes cleaner code - good suggestion

pkg/operator/controller/externaldns/deployment.go
lmzuccarelli: Added the check in cbld.build()

pkg/operator/controller/externaldns/pod.go
```diff
@@ -92,6 +92,14 @@ type externalDNSContainerBuilder struct {
...
 	seq := b.counter
 	b.counter++
+	// check each zone for the keyword 'privatednszones', if the provider is Azure
+	// this ensures that the provider 'azure-private-dns' is passed to the container
+	// to set the operand provider correctly
+	if b.provider == externalDNSProviderTypeAzure {
+		if strings.Contains(strings.ToLower(zone), azurePrivateDNSZonesResourceSubStr) {
+			b.provider = externalDNSProviderTypeAzurePrivate
+		}
+	}
```
alebedev87, line 102: Sorry, I didn't precise the proposal I made. What I meant by moving this part in the container builder is that in the container builder we have methods for each individual provider. And this Azure specific code would feel like at home in `fillAzureFields()`, unlike here in `build()` where the code is platform agnostic.
To make this happen I think that `--provider` flag would have to be moved to each provider.

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -1150,6 +1316,68 @@ func TestDesiredExternalDNSDeployment(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:             "Private Zone Azure",
```
alebedev87, line 1320: Does it make sense to repeat the same test for `Route` source? Will this go via different code path?

lmzuccarelli: I think this can be removed as we already have tested the code in the `Service` section.

pkg/operator/controller/externaldns/pod.go
lmzuccarelli: Ok got it - this makes sense

lmzuccarelli: Ok makes total sens - I have moved the code to fillAzureFields() now

alebedev87:
/lgtm
Azure test case failed though, worth checking what went wrong. Ping me if this needs a new commit, @lmzuccarelli.

PR #90, opened 2021-12-06
test utils: lookupCNAME explained and safe

alebedev87:


test/e2e/util.go
```diff
@@ -214,20 +214,34 @@ func rootCredentials(kubeClient client.Client, name string) (map[string][]byte,
...
+	if host[len(host)-1] != '.' {
+		host += "."
+	}
+	m.SetQuestion(host, dns.TypeCNAME)
 	r, _, err := c.Exchange(&m, server+":53")
 	if err != nil {
-		return nil, err
+		return "", err
+	}
+	if len(r.Answer) == 0 {
```
arjunrn, line 237: Isn't this an error in our case?

alebedev87: Yes, I think it's better to an error case.

alebedev87:
`Error fetching blob: invalid status code from registry 500 (Internal Server Error)`

PR #91, opened 2021-12-07
[WIP] test e2e: handle the abrupt stop

alebedev87:


alebedev87:
After discussion with Arjun, putting this on hold to look for a better replacement of `go test -timeout` for the timeout management.
/hold

alebedev87:
`testing.T.Cleanup()` cleanups for each test instance, so for each test (as expected). So it can't be used as-is for the zone cleanup.

alebedev87:
`go test -timeout`'s panic cannot be handled in any of the test not in main.

alebedev87:
Let's postpone this one until better times
/close

PR #92, opened 2021-12-07
 CFE-193 : E2e test cases to verify multi ingress support in externalDNS

Connect2naga:


test/e2e/operator_test.go
```diff
@@ -338,17 +426,61 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
...
+func verifyOpenShiftRouteSource(t *testing.T, canonicalName, host string) {
+	//var canonicalName = "router-external-dns.external-dns.example-naga.info"
+	// try all nameservers and fail only if all failed
+	recordExist := false
+	for _, nameSrv := range nameServers {
+		t.Logf("Looking for cname record in nameserver: %s", nameSrv)
+		if err := wait.PollImmediate(dnsPollingInterval, dnsPollingTimeout, func() (done bool, err error) {
+			cnames, err := lookupCNAME(host, nameSrv)
+			if err != nil {
+				t.Logf("cname lookup failed for nameserver : %s , error : %v", nameSrv, err)
```
alebedev87, line 438:
```suggestion
				t.Logf("CNAME lookup failed, error : %v", err)
```
As the nameserver was already mentioned at the beginning of the polling.

test/e2e/operator_test.go
```diff
@@ -338,17 +426,61 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
...
+				if strings.Contains(cname, canonicalName) {
+					recordExist = true
+					return true, nil
+				}
 			}
-			if len(nameserver) > 0 {
-				return d.DialContext(ctx, network, fmt.Sprintf("%s:53", nameserver))
+			return false, nil
+		}); err != nil {
+			t.Logf("Failed to verify host record with CNAME Record")
```
alebedev87, line 449:
```suggestion
			t.Logf("Failed polling for CNAME Record")
```

test/e2e/operator_test.go
```diff
@@ -338,17 +426,61 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
...
+		}
+	}
+
+	if !recordExist {
+		t.Logf("Cname record not found in any nameService, heance test failed")
+		t.Fail()
+		return
```
alebedev87, line 456:
```suggestion
		t.Fatalf("CNAME record not found in any nameserver")
```

test/e2e/operator_test.go
```diff
@@ -338,17 +426,61 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
...
+			Namespace: route1Name.Namespace,
+			Name:      route1Name.Name,
+		}, &route1); err != nil {
+			return false, err
+		}
+		if len(route1.Status.Ingress) != 2 {
+			return false, nil
+		}
+
+		for _, ingress := range route1.Status.Ingress {
```
alebedev87, line 473: Could not add a suggestion due to deleted lines:
```go
		for _, ingress := range route1.Status.Ingress {
			if strings.Contains(ingress.RouterCanonicalHostname, hostedZoneDomain) {
				canonicalName = ingress.RouterCanonicalHostname
                                return true, nil
			}
		}
		return false, nil
```

test/e2e/operator_test.go
```diff
@@ -338,17 +426,61 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
...
+
+	if !recordExist {
+		t.Logf("Cname record not found in any nameService, heance test failed")
+		t.Fail()
+		return
+	}
+}
+
+func fetchRouterCanonicalHostname(route1Name types.NamespacedName) (canonicalName string, err error) {
+	route1 := routev1.Route{}
```
alebedev87, line 461: Route number 1! numero uno! :)

test/e2e/util.go
```diff
@@ -214,20 +234,114 @@ func rootCredentials(kubeClient client.Client, name string) (map[string][]byte,
...
+			}
+			return d.DialContext(ctx, network, address)
+		},
+	}
+}
+
+func assertIngressControllerDeleted(t *testing.T, cl client.Client, ing *operatorv1.IngressController) {
+	t.Helper()
+	if err := deleteIngressController(t, cl, ing, 2*time.Minute); err != nil {
+		t.Fatalf("WARNING: cloud resources may have been leaked! failed to delete ingresscontroller %s: %v", ing.Name, err)
```
alebedev87, line 255: Which cluster resources you refer to?

test/e2e/util.go
```diff
@@ -154,51 +160,65 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
...
+		resource.Spec.Source = getServiceTypeSource(zoneDomain)
+	case "OpenShiftRoute":
+		resource.Spec.Source = getOpenshiftRouteTypeSource(routerName, zoneDomain)
+		// TODO : test with default as routername for default router
+		if routerName == "" || routerName == "default" {
+			resource.Spec.Source.OpenShiftRoute = nil
+		}
+	default:
+		//TODO : we should return error and fail the test on Source type
+		fmt.Errorf("source type not support : %s", sourceType)
```
alebedev87, line 184: Why not making 2 functions: `testExternalDNSServiceSource` and `testExternalDNSRouteSource` and not bother about passing the right source type (in the right case too)?
This would make the function call lighter too: 1 parameter fewer.

test/e2e/util.go
```diff
@@ -214,20 +234,114 @@ func rootCredentials(kubeClient client.Client, name string) (map[string][]byte,
...
+				return true, nil
+			}
+			t.Logf("failed to delete ingress controller %s/%s: %v", ic.Namespace, ic.Name, err)
+			return false, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("timed out waiting for ingresscontroller to be deleted: %v", err)
+	}
```
alebedev87, line 307: You don't trust Kube API?

test/e2e/util.go
```diff
@@ -154,51 +160,65 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
...
-					},
+		},
+	}
+
+	switch sourceType {
+	case "Service":
+		resource.Spec.Source = getServiceTypeSource(zoneDomain)
+	case "OpenShiftRoute":
+		resource.Spec.Source = getOpenshiftRouteTypeSource(routerName, zoneDomain)
+		// TODO : test with default as routername for default router
```
alebedev87, line 178: This TODO kinda confirms that my question about merging both OpenShift test cases into one may need to be thought of.

alebedev87 (commented):
Thanks for the PR, @Connect2naga! I think it needs some changes though.

test/e2e/aws.go
```diff
@@ -27,8 +27,6 @@ type awsTestHelper struct {
 	secretKey string
 }
 
-var _ providerTestHelper = &awsTestHelper{}
```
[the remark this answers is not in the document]
Connect2naga: this is not required, if the interface not implemented then it will fail in build, explicit check not required.

test/e2e/operator_test.go
```diff
@@ -123,7 +125,7 @@ func TestMain(m *testing.M) {
 	}
 
 	if version.SHORTCOMMIT != "" {
-		hostedZoneDomain = strconv.FormatInt(time.Now().Unix(), 10) + "." + version.SHORTCOMMIT + "." + baseZoneDomain
+		hostedZoneDomain = version.SHORTCOMMIT + "." + baseZoneDomain
```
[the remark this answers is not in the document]
Connect2naga: i think this is not required, bcz when we create any zone in all the providers we are checking is exist or not, if not exist then creation,
either we have to use time stamp or Short commit ID, both are not required.

test/e2e/operator_test.go
```diff
@@ -138,6 +140,19 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	fmt.Printf("create the provider credentails ready for external-dns-operator")
+	resourceSecret = helper.makeCredentialsSecret("external-dns-operator")
```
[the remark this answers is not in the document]
Connect2naga: ok.
if this is local to the test then we should move the credential also inside the test itself, isn't it ?

test/e2e/operator_test.go
```diff
@@ -158,26 +173,56 @@ func TestOperatorAvailable(t *testing.T) {
...
+	t.Log("Creating source service")
+	service := defaultService(testServiceName, testNamespace)
+	if err := kubeClient.Create(context.Background(), service); err != nil {
+		t.Fatalf("Failed to create test service %s/%s: %v", testNamespace, testServiceName, err)
+	}
+	defer kubeClient.Delete(context.TODO(), service)
+	verifySourceServiceDNSRecords(t, testNamespace, testServiceName)
+}
+
+func TestExternalDNSRecordLifecycleWithSourceAs_Default_Route(t *testing.T) {
```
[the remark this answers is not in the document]
Connect2naga: Done

test/e2e/operator_test.go
```diff
@@ -158,26 +173,56 @@ func TestOperatorAvailable(t *testing.T) {
...
+	testRouteNamespace := "test-extdns-openshift-route-default"
```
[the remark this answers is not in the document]
Connect2naga: ok :)

test/e2e/util.go
Connect2naga: means which ingress Controller are we deleting?
so here we are passing the ing from the caller, so this will be removed.

test/e2e/util.go
Connect2naga: there is no issue, we are write two functions,
but the we loose the code re-usability.

test/e2e/util.go
Connect2naga: kube APi will create the instance but in few instance , the conditions were not updated so , test is failed sporadically.
to avoid the sporadically this check added.

test/e2e/util.go
Connect2naga: yes, we can keep only one test case, no need of two

test/e2e/operator_test.go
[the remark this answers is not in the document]
Connect2naga: Reverted these changes,

test/e2e/util.go
Connect2naga: these function call will be replaced by inline. actual calling will be optimized by go itself.

test/e2e/operator_test.go
Connect2naga: :) corrected

test/e2e/operator_test.go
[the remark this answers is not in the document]
alebedev87: The problem with the ensuring of the hosted zone is that [we check by the domain name](https://github.com/openshift/external-dns-operator/blob/main/test/e2e/aws.go#L92), not by zone ID.
So, when you test on local and you managed to have some leftover zones you risk to get into old zone as the commit often doesn't change.

test/e2e/util.go
alebedev87: I don't see the link with the deletion which this method is for. I don't think we need this code, if API server told us the object is deleted we don't have any reason to not trust it.
If we would follow this logic, we have to check the deletion everywhere: when we delete `ExternalDNS` instance, `Route`, `Service`, etc.

test/e2e/util.go
alebedev87: The usability will not be harmed, you will still use the same common base and will just add the source `Service` for `testExternalDNSServiceSource` and the source `Route` for `testExternalDNSRouteSource`. Similar to how we build `ExternalDNS` instance in the helpers.
This will avoid the delema about  if the source given is not `Source` or `OpenShiftRoute`.

test/e2e/util.go
alebedev87: The ingress controller resource cleanup is up to the `cluster-ingress-controller`, if there is any cleanup to do a finalize is used so that when the delete API call is done we are sure the controller is taken care by the operator. Also, the ingress controller you create is a `HostNetwork` one, so it doesn't even allocate a `LoadBalancer`.
So, I don't think we need `assure` method here, we can remove the ingress controller instance like any other by just calling `defer <delete-ingress-controller-api-call>` like for other resources (externalDNS, route, etc.).

test/e2e/operator_test.go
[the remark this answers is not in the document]
Connect2naga: reverted the code. original kept back.

test/e2e/util.go
Connect2naga: sure, done.

test/e2e/util.go
Connect2naga: Done.

test/e2e/util.go
Connect2naga: sure , removed this code and used API alone.

test/e2e/util.go
alebedev87: Yes, let's make your case work, then the other can be deleted because basically they are testing the same code path.

test/e2e/operator_test.go
```diff
@@ -158,26 +173,56 @@ func TestOperatorAvailable(t *testing.T) {
...
+	t.Log("Ensuring test namespace")
+	err := kubeClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testRouteNamespace}})
+	if err != nil && !errors.IsAlreadyExists(err) {
+		t.Fatalf("Failed to ensure namespace %s: %v", testRouteNamespace, err)
+	}
+
+	externalDnsServiceName := fmt.Sprintf("%s-source-as-default-route", testExtDNSName)
+	t.Logf("Creating external dns instance :%s", externalDnsServiceName)
+	extDNS := helper.buildExternalDNS(externalDnsServiceName, hostedZoneID, hostedZoneDomain,
```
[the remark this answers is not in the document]
alebedev87: The code was reverted to Mustafa's version.

test/e2e/operator_test.go
```diff
@@ -237,35 +282,78 @@ func TestExternalDNSWithRoute(t *testing.T) {
...
 	t.Log("Ensuring test namespace")
-	err := kubeClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNamespace}})
+	err := kubeClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testIngressNamespace}})
 	if err != nil && !errors.IsAlreadyExists(err) {
-		t.Fatalf("Failed to ensure namespace %s: %v", testNamespace, err)
+		t.Fatalf("Failed to ensure namespace %s: %v", testIngressNamespace, err)
 	}
+	openshiftRouterName := "external-dns"
+	name := types.NamespacedName{Namespace: testIngressNamespace, Name: openshiftRouterName}
+	t.Logf("Create custome ingress controller %s/%s", name.Namespace, name.Name)
```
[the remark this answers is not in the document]
alebedev87: This one is still here.

test/e2e/operator_test.go
```diff
@@ -237,35 +282,78 @@ func TestExternalDNSWithRoute(t *testing.T) {
...
+		t.Fatalf("failed to create ingresscontroller: %v", err)
+	}
+	defer assertIngressControllerDeleted(t, kubeClient, ing)
 
-	t.Log("Creating credentials secret")
-	resourceSecret := helper.makeCredentialsSecret(testCredSecretName)
-	err = kubeClient.Create(context.TODO(), resourceSecret)
-	if err != nil {
-		t.Fatalf("Failed to create credentials secret %s/%s for resource: %v", resourceSecret.Namespace, resourceSecret.Name, err)
+	if err = waitForIngressControllerCondition(t, kubeClient, 15*time.Minute, name); err != nil {
```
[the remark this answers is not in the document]
alebedev87: Why you removed the waiting for the ingress readiness? i think it was the right thing to do.

test/e2e/operator_test.go
```diff
@@ -237,35 +282,78 @@ func TestExternalDNSWithRoute(t *testing.T) {
...
 	if err != nil && !errors.IsAlreadyExists(err) {
-		t.Fatalf("Failed to ensure namespace %s: %v", testNamespace, err)
+		t.Fatalf("Failed to ensure namespace %s: %v", testIngressNamespace, err)
 	}
+	openshiftRouterName := "external-dns"
+	name := types.NamespacedName{Namespace: testIngressNamespace, Name: openshiftRouterName}
+	t.Logf("Create custome ingress controller %s/%s", name.Namespace, name.Name)
+	ing := newHostNetworkController(name, name.Name+"."+hostedZoneDomain)
+	if err = kubeClient.Create(context.TODO(), ing); err != nil && !errors.IsAlreadyExists(err) {
+		t.Fatalf("failed to create ingresscontroller: %v", err)
```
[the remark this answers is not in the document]
alebedev87: Still here.

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+	externalDnsServiceName := fmt.Sprintf("%s-source-as-openshift-route", testExtDNSName)
+	t.Logf("Creating external dns instance : %s", externalDnsServiceName)
+	extDNS := helper.buildOpenShiftExternalDNS(externalDnsServiceName, hostedZoneID, hostedZoneDomain, openshiftRouterName)
+	if err = kubeClient.Create(context.TODO(), &extDNS); err != nil && !errors.IsAlreadyExists(err) {
+		t.Fatalf("Failed to create external DNS %q: %v", testExtDNSName, err)
+	}
+	defer kubeClient.Delete(context.TODO(), &extDNS)
+
+	// Create conflicting routes in the namespaces
+	makeRoute := func(name types.NamespacedName, host, path string) *routev1.Route {
```
alebedev87, line 384: `testRoute` function from `util.go` can be used here instead.

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+	// The first route should be admitted
+	if err = kubeClient.Create(context.TODO(), route); err != nil && !errors.IsAlreadyExists(err) {
+		t.Fatalf("failed to create route: %v", err)
+	}
+	defer kubeClient.Delete(context.TODO(), route)
+	canonicalName := ""
+	t.Logf("Getting canonicalName for the route :%s ", routeName.Name)
+	if canonicalName, err = fetchRouterCanonicalHostname(t, routeName); err != nil {
+		t.Fatalf("Failed to get RouterCanonicalHostname for route %s/%s: %v", routeName.Namespace, routeName.Name, err)
+	}
```
alebedev87, line 419:
```suggestion
	canonicalName, err := fetchRouterCanonicalHostname(t, routeName)
	if err != nil {
		t.Fatalf("Failed to get RouterCanonicalHostname for route %s/%s: %v", routeName.Namespace, routeName.Name, err)
	}
```

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+	t.Logf("canonicalName  : %s for the route :%s ", routeName.Name, canonicalName)
+	verifyOpenShiftRouteSource(t, canonicalName, fmt.Sprintf("app.%s", hostedZoneDomain))
```
alebedev87, line 421: can you please define a variable for the route's host, so that it's guaranteed to be the same when you create the route and when you verify?

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+}
+
+func verifyOpenShiftRouteSource(t *testing.T, canonicalName, host string) {
```
alebedev87, line 424: What this function is verifying is that the given host is resolved to the given CNAME, not the source of type `Route` in ExternalDNS custom resource. Something like `verifyCNAMERecord` would better reflect the contents.
BTW do we really need this function? Why not adding its body directly to the test code?

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+
+	if !recordExist {
+		t.Logf("Cname record not found in any nameService, heance test failed")
+		t.Fail()
+		return
+	}
+}
+
+func fetchRouterCanonicalHostname(t *testing.T, route1Name types.NamespacedName) (string, error) {
+	customeRoute := routev1.Route{}
```
alebedev87, line 454:
```suggestion
	route := routev1.Route{}
```

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+	}
+
+	if !recordExist {
+		t.Logf("Cname record not found in any nameService, heance test failed")
+		t.Fail()
+		return
+	}
+}
+
+func fetchRouterCanonicalHostname(t *testing.T, route1Name types.NamespacedName) (string, error) {
```
alebedev87, line 453:
```suggestion
func fetchRouterCanonicalHostname(t *testing.T, routeName types.NamespacedName) (string, error) {
```

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+			return false, err
+		}
+		if len(customeRoute.Status.Ingress) < 1 {
+			t.Logf("No ingress found in route, retrying..")
+			return false, nil
+		}
+
+		for _, ingress := range customeRoute.Status.Ingress {
+			if strings.Contains(ingress.RouterCanonicalHostname, hostedZoneDomain) {
+				canonicalName = ingress.RouterCanonicalHostname
```
alebedev87, line 471:
```suggestion
				canonicalName = ingress.RouterCanonicalHostname
				return true,nil
```

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+			}
+		}
+		if canonicalName == "" {
+			t.Logf("unable to fetch the canonicalHostname, retrying..")
+			return false, nil
+		}
+		return true, nil
```
alebedev87, line 478:
```suggestion
	        t.Logf("No matching canonicalHostname, retrying..")
		return false, nil
```

test/e2e/operator_test.go
Connect2naga: taken care already.

test/e2e/operator_test.go
Connect2naga: Done

test/e2e/operator_test.go
Connect2naga: name Change done
copying directly is not ok, it looks clumsy  :( 
i prefer more granular(which will speaks more about the tasks like setup , actions, verification ...) but it will leads to lot more changes.

test/e2e/operator_test.go
Connect2naga: done

test/e2e/operator_test.go
Connect2naga: already takencare.

test/e2e/operator_test.go
Connect2naga: taken care already :)

test/e2e/util.go
```diff
@@ -255,3 +260,20 @@ func equalFQDN(name1, name2 string) bool {
...
+func newHostNetworkController(name types.NamespacedName, domain string) *operatorv1.IngressController {
+	repl := int32(1)
+	return &operatorv1.IngressController{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: name.Namespace,
+			Name:      name.Name,
+		},
+		Spec: operatorv1.IngressControllerSpec{
+			Domain:   domain,
+			Replicas: &repl,
```
arjunrn, line 273: You can use the [pointer helper function](https://pkg.go.dev/k8s.io/utils/pointer#Int32) here.

arjunrn (commented):
Some minor nits but looks good overall.

config/manager/kustomization.yaml
```diff
@@ -8,5 +8,5 @@ generatorOptions:
   disableNameSuffixHash: true
 images:
 - name: controller
-  newName: controller
+  newName: quay.io/snagarju/external-dns-operator
```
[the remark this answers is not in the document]
Connect2naga: Yes Arjun, we are waiting for the externalDNS latest image

test/e2e/operator_test.go
```diff
@@ -350,3 +355,106 @@ func customResolver(nameserver string) *net.Resolver {
 		},
 	}
 }
+
+func TestExternalDNSRecordLifecycleWithSourceAs_OpenShiftRoute(t *testing.T) {
```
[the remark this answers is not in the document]
Connect2naga: done.

test/e2e/operator_test.go
```diff
@@ -350,3 +355,129 @@ func customResolver(nameserver string) *net.Resolver {
...
+	testIngressNamespace := "test-extdns-openshift-route"
+	t.Log("Ensuring test namespace")
+	err := kubeClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testIngressNamespace}})
+	if err != nil && !errors.IsAlreadyExists(err) {
+		t.Fatalf("Failed to ensure namespace %s: %v", testIngressNamespace, err)
+	}
+	openshiftRouterName := "external-dns"
+	name := types.NamespacedName{Namespace: testIngressNamespace, Name: openshiftRouterName}
+	t.Logf("Create custome ingress controller %s/%s", name.Namespace, name.Name)
```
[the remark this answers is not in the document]
Connect2naga: done

test/e2e/util.go
Connect2naga: done

test/e2e/operator_test.go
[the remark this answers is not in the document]
Connect2naga: Done

test/e2e/operator_test.go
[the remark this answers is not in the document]
Connect2naga: done

test/e2e/operator_test.go
alebedev87: No, I still can see the check for the empty canonicalName. It's not needed anymore as we return the name is found.

Connect2naga: it should be like this.
```
	for _, ingress := range route.Status.Ingress {
			if strings.Contains(ingress.RouterCanonicalHostname, hostedZoneDomain) {
				canonicalName = ingress.RouterCanonicalHostname
				return true, nil
			}
		}
	return false, nil
```

test/e2e/util.go
```diff
@@ -235,7 +244,7 @@ func lookupCNAMEMiekg(host, server string) (string, error) {
 		return "", err
 	}
 	if len(r.Answer) == 0 {
-		return "", fmt.Errorf("not found")
+		return "", fmt.Errorf("No results for the host:%s in nameServer: %s ", host, server)
```
alebedev87, line 247: I think this is a bit of an overkill as error. Here is what we are getting if this error is used:
```
 operator_test.go:221: Looking for DNS record in nameserver: ns-596.awsdns-10.net
 operator_test.go:227: Waiting for DNS record: myroute.1639073290.3c8d057.example-test.info, error: No results for the host :myroute.1639073290.3c8d057.example-test.info. in nameServer : ns-596.awsdns-10.net
```
Note the redundant hostname and nameserver. They are passed to the function, so the caller always knows them and can log them if there is a need, that's what happened here.

Compare the previous example to this:
```
operator_test.go:221: Looking for DNS record in nameserver: ns-596.awsdns-10.net
operator_test.go:227: Waiting for DNS record: myroute.1639073290.3c8d057.example-test.info, error: not found
```

test/e2e/operator_test.go
```diff
@@ -350,3 +354,106 @@ func customResolver(nameserver string) *net.Resolver {
...
+	canonicalName, err := fetchRouterCanonicalHostname(t, routeName)
+	if err != nil {
+		t.Fatalf("Failed to get RouterCanonicalHostname for route %s/%s: %v", routeName.Namespace, routeName.Name, err)
+	}
+	t.Logf("CanonicalName: %s for the route: %s ", routeName.Name, canonicalName)
+	verifyCNAMERecordForOpenshiftRoute(t, canonicalName, host)
+}
+
+func verifyCNAMERecordForOpenshiftRoute(t *testing.T, canonicalName, host string) {
+	//var canonicalName = "router-external-dns.external-dns.example-naga.info"
```
alebedev87, line 403: Leftover

test/e2e/operator_test.go
```diff
@@ -350,3 +354,106 @@ func customResolver(nameserver string) *net.Resolver {
 		},
 	}
 }
+
+// Test to verify the ExternalDNS should create the CNAME and txt record for the OpenshiftRoute
```
alebedev87, line 358: This test doesn't test the creation of TXT record.

test/e2e/operator_test.go
```diff
@@ -350,3 +354,106 @@ func customResolver(nameserver string) *net.Resolver {
...
+// with multiple ingress controller deployed in Openshift.
+// the canonical name of route should resolve to the CNAME of Host,
```
alebedev87, line 360:
```suggestion
// Route's host should resolve to the canonical name of the specified ingress controller.
```

Connect2naga: Done

test/e2e/operator_test.go
Connect2naga: taken care

test/e2e/operator_test.go
Connect2naga: taken care

test/e2e/util.go
```diff
@@ -181,8 +183,8 @@ func defaultExternalDNS(name, zoneID, zoneDomain string) operatorv1alpha1.Extern
 	}
 }
 
-func routeExternalDNS(name, zoneID, zoneDomain string) operatorv1alpha1.ExternalDNS {
-	return operatorv1alpha1.ExternalDNS{
+func routeExternalDNS(name, zoneID, zoneDomain, routerName string) operatorv1alpha1.ExternalDNS {
+	resoure := operatorv1alpha1.ExternalDNS{
```
alebedev87, line 187: `resoure`?! You wanted to say `resource` I think. Something like `extDNS` can be even more meaningful.

Connect2naga: this resource itself not needed, we will use direct return.
was mis-directed with previous test failure

test/e2e/operator_test.go
```diff
@@ -217,12 +221,12 @@ func TestExternalDNSWithRoute(t *testing.T) {
...
 		// verify dns records has been created for the route host.
 		if err := wait.PollImmediate(dnsPollingInterval, dnsPollingTimeout, func() (done bool, err error) {
-			cname, err := lookupCNAMEMiekg(testRouteHost, nameSrv)
+			cname, err := lookupCNAME(testRouteHost, nameSrv)
 			if err != nil {
 				t.Logf("Waiting for DNS record: %s, error: %v", testRouteHost, err)
 				return false, nil
 			}
-			if equalFQDN(cname, targetRouterCName) {
+			if strings.Contains(cname, targetRouterCName) {
```
alebedev87, line 229: This should be reverted, we want the exact match, not substring.

test/e2e/util.go
```diff
@@ -194,6 +196,9 @@ func routeExternalDNS(name, zoneID, zoneDomain string) operatorv1alpha1.External
 					AnnotationFilter: map[string]string{
 						"external-dns.mydomain.org/publish": "yes",
 					},
+					OpenShiftRoute: &operatorv1alpha1.ExternalDNSOpenShiftRouteOptions{
+						RouterName: routerName,
+					},
```
alebedev87, line 201: This will make the existing Mustafa's test not working as ExteranlDNS > 0.10.1 will be needed for his case too.
Either you make the route addition optional if `routerName` is empty, or both cases need to be merged.

test/e2e/operator_test.go
```diff
@@ -350,3 +354,106 @@ func customResolver(nameserver string) *net.Resolver {
 		},
 	}
 }
+
+// Test to verify the ExternalDNS should create the CNAME record for the OpenshiftRoute
+// with multiple ingress controller deployed in Openshift.
+// Route's host should resolve to the canonical name of the specified ingress controller.
+func TestExternalDNSCustomIngress(t *testing.T) {
+	t.Skip("Ths test need to enable once latest external-dns image available.")
```
alebedev87, line 362: Can you please mention the exact upstream version we need? `> v0.10.1`

test/e2e/operator_test.go
Connect2naga: ok

test/e2e/operator_test.go
```diff
@@ -350,3 +354,106 @@ func customResolver(nameserver string) *net.Resolver {
 		},
 	}
 }
+
+// Test to verify the ExternalDNS should create the CNAME record for the OpenshiftRoute
+// with multiple ingress controller deployed in Openshift.
+// Route's host should resolve to the canonical name of the specified ingress controller.
+func TestExternalDNSCustomIngress(t *testing.T) {
+	t.Skip("Ths test need to enable once latest external-dns image available(>v0.10.1).")
```
alebedev87, line 362:
```suggestion
	t.Skip("The test needs to be enabled once latest external-dns image available (>v0.10.1).")
```

test/e2e/operator_test.go
alebedev87: Then it can be used in `TestExternalDNSWithRoute` test case too. But fine, let's keep it for the followup PRs.

test/e2e/operator_test.go
Connect2naga: Done

test/e2e/operator_test.go
```diff
@@ -358,3 +363,112 @@ func customResolver(nameserver string) *net.Resolver {
...
+				return true, nil
+			}
+			return false, nil
+		}); err != nil {
+			t.Logf("Failed to verify host record with CNAME Record")
+		}
+	}
+
+	if !recordExist {
+		t.Fatalf("Cname record not found in any nameService, heance test failed")
```
arjunrn, line 442:
```suggestion
		t.Fatalf("CNAME record not found in any name server")
```

.golangci.yaml
```diff
@@ -14,6 +14,9 @@ run:
 
   # include test files or not, default is true
   tests: true
+  # list of build tags, all linters use it. Default is empty list.
+  build-tags:
```
arjunrn, line 18: Nice, I didn't realize I could simply include the build tags here.

test/e2e/operator_test.go
Connect2naga: Done.

.golangci.yaml
Connect2naga: Thanks :)

PR #93, opened 2021-12-07
Use miekg dns library for A record lookups

arjunrn:
Instead of using two different libraries for DNS record lookups the test will use only miekg/dns library for both A and CNAME record lookups.

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -9,11 +9,9 @@ import (
 	"math/rand"
 	"os"
 	"reflect"
-	"strings"
 	"testing"
 	"time"
 
-	"github.com/miekg/dns"
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	routev1 "github.com/openshift/api/route/v1"
@@ -24,6 +22,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/pointer"
+
+	"github.com/miekg/dns"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	operatorv1alpha1 "github.com/openshift/external-dns-operator/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


test/e2e/util.go
```diff
@@ -249,16 +246,28 @@ func lookupCNAME(host, server string) (string, error) {
...
+	for _, ans := range response.Answer {
+		if aRec, ok := ans.(*dns.A); ok {
+			ips = append(ips, aRec.A.String())
+		}
+	}
+	return ips, nil
+}
+
+func equalFQDN(name1, name2 string) bool {
+	return dns.Fqdn(name1) == dns.Fqdn(name2)
```
alebedev87, line 270: Nice!

test/e2e/util.go
```diff
@@ -229,18 +229,15 @@ func rootCredentials(kubeClient client.Client, name string) (map[string][]byte,
...
+	dnsClient := dns.Client{}
+	message := &dns.Msg{}
+	message.SetQuestion(dns.Fqdn(host), dns.TypeCNAME)
+	r, _, err := dnsClient.Exchange(message, fmt.Sprintf("%s:53", server))
 	if err != nil {
 		return "", err
 	}
 	if len(r.Answer) == 0 {
-		return "", fmt.Errorf("not found")
+		return "", fmt.Errorf("no CNAME records found for host %s on %s", host, server)
```
alebedev87, line 240: https://github.com/openshift/external-dns-operator/pull/92#discussion_r766123793

test/e2e/util.go
```diff
@@ -249,16 +246,28 @@ func lookupCNAME(host, server string) (string, error) {
...
-		name2 += "."
+	if len(response.Answer) == 0 {
+		return nil, fmt.Errorf("no IPs found for host %s on %s", host, server)
 	}
-	return name1 == name2
+	var ips []string
+	for _, ans := range response.Answer {
+		if aRec, ok := ans.(*dns.A); ok {
+			ips = append(ips, aRec.A.String())
+		}
```
alebedev87, line 264: Can we match the behavior of `lookupCNAME` function here? For the case when the record is not `dns.A`.

test/e2e/util.go
alebedev87: Not sure whether the anchor to a resolved comment works well, let me quote it:

> I think this is a bit of an overkill as error. Here is what we are getting if this error is used:
> ```
> operator_test.go:221: Looking for DNS record in nameserver: ns-596.awsdns-10.net
> operator_test.go:227: Waiting for DNS record: myroute.1639073290.3c8d057.example-test.info, error: No results for the > host :myroute.1639073290.3c8d057.example-test.info. in nameServer : ns-596.awsdns-10.net
>```
>
>Note the redundant hostname and nameserver. They are passed to the function, so the caller always knows them and can log them if there is a need, that's what happened here.
>
>Compare the previous example to this:
>```
>operator_test.go:221: Looking for DNS record in nameserver: ns-596.awsdns-10.net
>operator_test.go:227: Waiting for DNS record: myroute.1639073290.3c8d057.example-test.info, error: not found
>```

test/e2e/util.go
arjunrn: Looks like there are some records in the response which are not A records and that causes the whole lookup to fail. Instead I have added a check if number of IPs is 0.

PR #94, opened 2021-12-08
Replaced annotation filtering with label filtering

arjunrn:
Changed the annotation filter field in the ExternalDNS CR to a label filter where the filter is specified as a label selector. Added validation for field and corresponding tests.


api/v1alpha1/externaldns_types.go
```diff
@@ -427,14 +427,11 @@ type ExternalDNSSourceUnion struct {
 	// +required
 	Type ExternalDNSSourceType `json:"type"`
 
-	// AnnotationFilter describes an annotation filter
-	// used to filter which source instance resources
-	// ExternalDNS publishes records for.
-	// The annotation filter uses label selector semantics
-	// against source resource annotations.
+	// LabelFilter describes an annotation filter used to filter which source instance resources ExternalDNS publishes
+	// records for. The label filter uses label selector semantics against source resource annotations.
```
alebedev87, line 431:
```suggestion
	// LabelFilter describes a label filter used to filter which source instance resources ExternalDNS publishes
	// records for. The label filter uses label selector semantics against source resource labels.
```

api/v1alpha1/externaldns_types.go
```diff
@@ -427,14 +427,11 @@ type ExternalDNSSourceUnion struct {
...
 	//
 	// +optional
-	AnnotationFilter map[string]string `json:"annotationFilter,omitempty"`
+	LabelFilter *string `json:"labelFilter,omitempty"`
```
alebedev87, line 434: I think I'll stop here the review as the rest is pretty much dependent on this one.

I would like to challenge the type change. I think `map[string]string` is more appropriate as type. It brings UX similar to `metadata.labels` or any label selector.

@Miciah : WDYT?

arjunrn: @alebedev87 @Miciah It is in fact implemented as `metav1.LabelSelector`. Any arbitrary selector can be passed to the controller.

arjunrn: @Miciah just a observation. I see that both the `string` and `*metav1.LabelSelector` types are used in places. [Example 1](https://github.com/kubernetes/api/blob/872e73e89113b552c7677bb8e112b9ef7a61d826/apps/v1/types.go#L344) and [Example 2](https://github.com/kubernetes/apimachinery/blob/32df71429c452d9ef5e0290ee53932b39a9e4e36/pkg/apis/meta/v1/types.go#L342). As a user I would prefer to write the selector as a string because specifying it as a structed object is cumbersome. In fact the [docs](https://github.com/kubernetes/apimachinery/blob/1d209c836ad2fed6332e0aa08aefc669a649ba8e/pkg/apis/meta/v1/types.go#L1165) on `metav1.LabelSector` says _LabelSelector is an internal representation for the latter style._

pkg/operator/controller/externaldns/pod.go
```diff
@@ -139,12 +139,8 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
...
 	}
 
-	if len(b.externalDNS.Spec.Source.AnnotationFilter) > 0 {
-		annotationFilter := ""
-		for key, value := range b.externalDNS.Spec.Source.AnnotationFilter {
-			annotationFilter += fmt.Sprintf("%s=%s,", key, value)
-		}
-		args = append(args, fmt.Sprintf("--annotation-filter=%s", annotationFilter[0:len(annotationFilter)-1]))
+	if b.externalDNS.Spec.Source.LabelFilter != nil {
+		args = append(args, fmt.Sprintf("--label-filter=%s", metav1.FormatLabelSelector(b.externalDNS.Spec.Source.LabelFilter)))
```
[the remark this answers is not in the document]
arjunrn: Actually no, since they will be array values in the manifest adding a single quote into the string means it will be read by the controller as a value with quotes. These arguments are not processed by a shell before being passed to the command in the image.

alebedev87:
LGTM

cc: @Miciah for the final approval

alebedev87:
@Miciah : I'm giving a LGTM just to have some time for the API review before the feature freeze. fell free to come back to us if you think something needs a change.
/lgtm

PR #96, opened 2021-12-09
CFE-204: Adding test code to verify external-dns-operator on Azure private DNS

Elbehery:
This PR adds testing for Azure private DNS zone.

This PR is built atop of previous work done by @Connect2naga


test/e2e/azure.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
```
arjunrn, line 1: This build tag was added because you're using a newer version of go than currently used to build this operator.

test/e2e/azure.go
```diff
@@ -6,12 +7,15 @@ import (
...
 	"fmt"
+	"github.com/Azure/azure-sdk-for-go/services/privatedns/mgmt/2018-09-01/privatedns"
+	"os"
 	"strings"
 
 	operatorv1alpha1 "github.com/openshift/external-dns-operator/api/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2018-05-01/dns"
+	_ "github.com/Azure/azure-sdk-for-go/services/privatedns/mgmt/2018-09-01/privatedns"
```
arjunrn, line 18: I don't think this is a good idea particularly because this package has already been imported above.

test/e2e/azure.go
```diff
@@ -34,11 +38,14 @@ type clusterConfig struct {
...
-	zoneClient dns.ZonesClient
+	config                   *clusterConfig
+	zoneClient               dns.ZonesClient
+	pvtZoneClient            privatedns.PrivateZonesClient
+	virtualNetworkLinkClient privatedns.VirtualNetworkLinksClient
+	isPrivateZone            bool
 }
 
-var _ providerTestHelper = &azureTestHelper{}
+//var _ providerTestHelper = &azureTestHelper{}
```
arjunrn, line 48: So we don't need this anymore?

test/e2e/azure.go
```diff
@@ -176,3 +270,10 @@ func getAccessToken(cfg *clusterConfig) (*adal.ServicePrincipalToken, error) {
 	}
 	return nil, fmt.Errorf("no credentials provided for Azure API")
 }
+
+func getPrivateZoneFlag() bool {
```
arjunrn, line 274: So are we going to run the Azure test twice, once with this environment variable set and once without?

test/e2e/azure.go
Elbehery: i use `go1.16.4`  , but i will remove it now, thx for catching this 🙏🏽

test/e2e/azure.go
Elbehery: +1

test/e2e/azure.go
Elbehery: +1

test/e2e/azure.go
Elbehery: yes, we need to test for Azure ( public and private ) zones, each of them has different client package and setup .. so we use this flag to run the test one time for the public and another one for private

test/e2e/azure.go
```diff
@@ -176,3 +266,10 @@ func getAccessToken(cfg *clusterConfig) (*adal.ServicePrincipalToken, error) {
...
+	if os.Getenv("AZURE_PRIVATE_DNS") != "" {
+		return true
+	}
+	return false
```
arjunrn, line 274:
```suggestion
	return os.Getenv("AZURE_PRIVATE_DNS") != ""
```

test/e2e/operator_test.go
```diff
@@ -307,6 +307,32 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
 		t.Fatalf("Failed to get loadbalancer IPs for service %s/%s: %v", testNamespace, testServiceName, err)
 	}
 
+	// test for azure private zone
+	a, ok := helper.(*azureTestHelper)
+	if ok && a.isPrivateZone {
```
arjunrn, line 312: I don't understand this section. So the DNS record for the loadbalancer in the Azure private zones will be published publicly? And then we have to veriify that they are also published on the Azure's own DNS servers?

test/e2e/azure.go
Elbehery: updated 👍🏽

test/e2e/operator_test.go
Elbehery: afaik, the difference here is in the resolver. Azure PrivateDNS does not return name servers, we use the `virtual network link` to connect directly to the private dns zone, and lookup the test server record. 

Please correct me if i am wrong

arjunrn: IIUC the approach you are taking won't work because the link is being created between the test cluster and the private DNS zone. However the test does not run in the same test cluster. It could be running in another cluster which could be in another cloud environment.
Also I recommend you first stub out the test for Azure private DNS and enable the test in CI and then gradually enable the test. You will get faster feedback while implementing the tests.

Elbehery:
closing this

PR #98, opened 2021-12-10
Add the bundle generation check

alebedev87:
Add a new script to `make verify` target which:
- checks that the OLM bundle was updated with the latest files from `config/`
- runs `operator-sdk` bundle validation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,8 @@ INDEX_IMG ?= olm-bundle-index:latest
 OPM_VERSION ?= v1.17.4
 
 GOLANGCI_LINT_BIN=$(BIN_DIR)/golangci-lint
-GOLANGCI_LINT_VERSION=v1.43.0
+
+OPERATOR_SDK_BIN=$(BIN_DIR)/operator-sdk
 
 COMMIT ?= $(shell git rev-parse HEAD)
 SHORTCOMMIT ?= $(shell git rev-parse --short HEAD)
@@ -99,6 +100,7 @@ verify: lint
 	hack/verify-gofmt.sh
 	hack/verify-deps.sh
 	hack/verify-generated.sh
+	hack/verify-olm.sh
 
 ##@ Build
 GO=GO111MODULE=on GOFLAGS=-mod=vendor CGO_ENABLED=0 go
@@ -132,6 +134,7 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
+.SILENT: olm-manifests
 .PHONY: olm-manifests
 # TODO: try to replace this rule with 'operator-sdk generate bundle'
 # A little helper command to generate the manifests of OLM bundle from the files in config/.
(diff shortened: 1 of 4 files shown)


hack/verify-generated.sh
```diff
@@ -2,16 +2,22 @@
...
   make generate
   make manifests
+  echo "> generating the OLM bundle"
+  make olm-manifests
+  test -z "$(git status --porcelain | \grep -v '^??')" || print_failure
 
-  test -z "$(git status --porcelain)" || print_failure
-  echo "verified generated manifests and deep copy"
+  echo "> verifying the OLM bundle"
+  make validate-bundle || print_failure
```
arjunrn, line 20: Instead of calling validate-bundle in here can you add it as a prereq for the `verify` target in the _Makefile_.

Makefile
```diff
@@ -140,18 +140,18 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
...
-	cp -f config/rbac/auth_proxy_role.yaml $(BUNDLE_MANIFEST_DIR)/external-dns-operator-auth-proxy_rbac.authorization.k8s.io_v1_clusterrole.yaml
-	cp -f config/rbac/auth_proxy_role_binding.yaml $(BUNDLE_MANIFEST_DIR)/external-dns-operator-auth-proxy_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
-	cp -f config/rbac/auth_proxy_service.yaml $(BUNDLE_MANIFEST_DIR)/external-dns-operator-auth-proxy_v1_service.yaml
-	# opm is unable to find CRD if the standard yaml --- is at the top
-	sed -i -e '/^---$$/d' -e '/^$$/d' $(BUNDLE_MANIFEST_DIR)/*.yaml
-	# as per the recommendation of 'operator-sdk bundle validate' command
-	# doing the best (yaml regexp is far from perfect) to strip the metadata.namespace from the bundle manifests
-	# trying to not touch cluster level resources as they don't have the namespace
-	for f in $$(\grep -l 'kind: *\(Service\|ConfigMap\|Secret\|Role\) *$$' $(BUNDLE_MANIFEST_DIR)/*.yaml); do sed -i '/namespace:/d' $${f};done
+	@cp -f config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml $(BUNDLE_MANIFEST_DIR)/externaldns.olm.openshift.io_crd.yaml
```
arjunrn, line 143: I just found out about the `.SILENT` [fake target](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html#Special-Targets) which I think is what you want to use here.

hack/golangci-lint.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/sh
 set -e
 
-GOLANGCI_VERSION="1.43.0"
-
 OUTPUT_PATH=${1:-./bin/golangci-lint}
+GOLANGCI_VERSION=${2:-1.43.0}
```
arjunrn, line 5: Actually we don't want to do this, because we verify the SHA256 of the downloaded tar files and they will fail for any other version. We should drop the version in the `Makefile`.

alebedev87: Removed

hack/verify-generated.sh
alebedev87: Moved the OLM part into a dedicated file and added it to `verify` target.

alebedev87:
/retest

Cluster installation failed.

alebedev87:
The bot forgot about this PR?
/joke

PR #99, opened 2021-12-10
Reduce cluster wide privileges of operator

sherine-k:

* Stop using github.com/form3tech-oss/jwt-go. The report recommends using github.com/golang-jwt/jwt instead. But since this library is not really used, we suggest to remove it for now.
Update RBAC configuration, restricting access to resources such as secrets, serviceaccounts, namespaces, deployments to the operator and operand namespaces only

    Permissions for secrets, serviceaccounts & deployments creation/update in the operator namespace are given through a new role 
    Permissions for secrets & serviceaccounts & deployments creation/update in the operand namespace are given through a role and rolebinding that should be created manually by the clusterAdmin . See config/samples/extra_roles.yaml


Miciah:
This looks good to me.  I just have a couple of observations, but I'm not necessarily asking you to make changes in this PR:

* `ensureExternalDNSNamespace` still has logic to create the operand namespace if it doesn't exist.  I am fine with keeping this logic because (a) the user could choose to give the operator broader permission than what's in the "extra-roles" sample, and (b) we want eventually to find a way to give the operator this permission in a standard OLM-based installation even if we cannot do it in the 4.10 timeframe.  However, as it stands, if the user deletes the operand namespace, which includes the rolebinding that would give the operator permission to recreate the namespace, then the operator will try and fail to recreate the namespace: https://github.com/openshift/external-dns-operator/blob/8d36f1af0849503099666b34032d474ada567719/pkg/operator/controller/externaldns/namespace.go#L40-L42
* The instructions in `README.md` are formatted inconsistently, sometimes with one step for building and pushing an image, sometimes with one step to build and one step to push; similarly, sometimes one step creates both the operator namespace and the operand namespaces, and sometimes there is one step for each namespace.  The inconsistency absent an explanation is confusing.  However, this is an issue that already existed before your PR, so it should be addressed separately.  

/lgtm
/hold
to give @alebedev87, @arjunrn, @lmzuccarelli, or others the opportunity to have their say.

bundle/manifests/external-dns-operator_rbac.authorization.k8s.io_v1_rolebinding.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: RoleBinding
+metadata:
+  name: external-dns-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: external-dns-operator
+subjects:
+  - kind: ServiceAccount
+    name: external-dns-operator
```
alebedev87, line 11: Can you please add a newline here?

README.md
```diff
@@ -14,11 +14,17 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
    $ make image-build
    $ make image-push
    ```
-2. Run the following command to deploy the `ExternalDNS` Operator:
+2. Prepare the operator and operand namespaces:
+   ```
+   oc create ns external-dns-operator
```
alebedev87, line 19: In case of `make deploy` deployment, the operator namespace is created in `config/`: https://github.com/openshift/external-dns-operator/blob/main/config/manager/manager.yaml#L6.

config/rbac/operator_rolebinding.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: RoleBinding
+metadata:
+  name: external-dns-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: external-dns-operator
+subjects:
+  - kind: ServiceAccount
+    name: external-dns-operator
```
alebedev87, line 11: Same remark about the newline.

config/samples/extra-roles.yaml
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+  name: external-dns-operator
+  namespace: external-dns
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - serviceaccounts
+      - secrets
+      - namespaces
```
alebedev87, line 12: Namespaces too? In the local role?

test/e2e/operator_test.go
```diff
@@ -143,6 +144,21 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
+	err = ensureOperandNamespace()
+	if err != nil {
+		fmt.Printf("Failed to create external-dns namespace: %v\n", err)
+	}
```
alebedev87, line 150:
```suggestion
	if err = ensureOperandNamespace(); err != nil {
		fmt.Printf("Failed to create external-dns namespace: %v\n", err)
	}
```
A little more go idiomatic.

test/e2e/operator_test.go
```diff
@@ -364,6 +380,59 @@ func customResolver(nameserver string) *net.Resolver {
 	}
 }
 
+func ensureOperandNamespace() error {
+	err := kubeClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "external-dns"}})
+	return err
```
alebedev87, line 385:
```suggestion
	return kubeClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "external-dns"}})
	return err
```
Can you also make a constant for the operand namespace?

test/e2e/operator_test.go
```diff
@@ -364,6 +380,59 @@ func customResolver(nameserver string) *net.Resolver {
...
+
+	role := rbacv1.Role{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "external-dns-operator",
+			Namespace: "external-dns",
+		},
+		Rules: rules,
+	}
+	err := kubeClient.Create(context.TODO(), &role)
+	return err
```
alebedev87, line 410:
```suggestion
	return kubeClient.Create(context.TODO(), &role)
```

test/e2e/operator_test.go
```diff
@@ -364,6 +380,59 @@ func customResolver(nameserver string) *net.Resolver {
...
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "Role",
+			Name:     "external-dns-operator",
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      rbacv1.ServiceAccountKind,
+				Name:      "external-dns-operator",
+				Namespace: "external-dns-operator",
```
alebedev87, line 428: Would you mind making those constants too?

test/e2e/operator_test.go
```diff
@@ -364,6 +380,59 @@ func customResolver(nameserver string) *net.Resolver {
...
+			},
+		},
+	}
+	err := kubeClient.Create(context.TODO(), &rb)
+	return err
```
alebedev87, line 433:
```suggestion
	return kubeClient.Create(context.TODO(), &rb)
```

test/e2e/operator_test.go
```diff
@@ -143,6 +144,21 @@ func TestMain(m *testing.M) {
...
 	}
 
+	err = ensureOperandNamespace()
+	if err != nil {
+		fmt.Printf("Failed to create external-dns namespace: %v\n", err)
+	}
+
+	err = ensureOperandRole()
+	if err != nil {
+		fmt.Printf("Failed to create role external-dns-operator in ns external-dns: %v\n", err)
```
alebedev87, line 154: Don't we want to fail the test? What's the point of continuing?

sherine-k: Basically mostly this fails when the role already exists,  and we just want to go on...
Otherwise, we will need to take care of cleanups, we can also do that

README.md
sherine-k: Changed. thx

config/samples/extra-roles.yaml
sherine-k: Yes! otherwise the operator logs show errors on get namespace external-dns, so I had to keep it

test/e2e/operator_test.go
sherine-k: Done. Thx

test/e2e/operator_test.go
sherine-k: Done. Thx

test/e2e/operator_test.go
sherine-k: Done. Thx

test/e2e/operator_test.go
sherine-k: Done. Thx

test/e2e/operator_test.go
sherine-k: Done!

config/rbac/operator_rolebinding.yaml
sherine-k: Done. Thx!

bundle/manifests/external-dns-operator_rbac.authorization.k8s.io_v1_rolebinding.yaml
sherine-k: yes done!

test/e2e/operator_test.go
alebedev87: `ensure*` functions can do a `Get` call before `Create` to check if it's already there and return `error` only in case of a real failure (creation).

alebedev87:
@sherine-k : can you please keep `pkg/operator/controller/externaldns/namespace.go` file? It's not its fault that OLM doesn't

alebedev87:
/reopen
I'm really sorry, accidental click.

pkg/operator/controller/externaldns/controller.go
```diff
@@ -110,10 +110,6 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS cluster role: %w", err)
 	}
 
-	if _, _, err := r.ensureExternalDNSNamespace(ctx, r.config.Namespace); err != nil {
-		return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS namespace: %w", err)
-	}
```
alebedev87, line 115: I'm not sure about this one. It still makes sense for non OLM deployments.

We can do:
- Miciah proposed: just keep it as-is 
OR
- don't ensure the namespace in case the platform is OpenShift

cmd/external-dns-operator/main.go
```diff
@@ -20,10 +20,6 @@ import (
 	"flag"
 	"os"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
```
arjunrn, line 23: Why has this been dropped?

config/samples/extra-roles.yaml
dbaker-rh: I think this needs to re-open.  Seems fair to "get", "watch" on the namespace to squelch the error message, but you do not want to grant permissions to update it.

README.md
```diff
@@ -14,11 +14,16 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
    $ make image-build
    $ make image-push
    ```
-2. Run the following command to deploy the `ExternalDNS` Operator:
+2. Prepare the operator and operand namespaces:
```
Miciah, line 17:
```suggestion
2. Prepare the operand namespace:
```

README.md
```diff
@@ -97,11 +102,12 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
    $ make index-image-push
    ```
 
-5. Create the `external-dns-operator` namespace:
-   ```sh
-   $ oc create ns external-dns-operator
+5. Prepare the operator and operand namespaces:
```
Miciah, line 105:
```suggestion
5. Prepare the operand namespace:
```

config/rbac/operator_role.yaml
```diff
@@ -0,0 +1,29 @@
...
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: external-dns-operator
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+      - serviceaccounts
```
Miciah, line 11: Is this actually needed?  Granted, it isn't a significant security concern if the operator can manage serviceaccounts in its own namespace, but it isn't obvious to me why this is needed.

pkg/operator/controller/externaldns/controller.go
Miciah: We can remove `ensureExternalDNSNamespace` for now if @dbaker-rh prefers we do so.  The code is still there in Git history for such a time as we get the RBAC issues sorted out.

Miciah (commented):
I had a couple very minor comments, but nothing that should necessarily block this PR from merging.
/lgtm

config/rbac/operator_role.yaml
Miciah: By "this", I mean delete/get/list/update/watch access to serviceaccounts.

sherine-k:
Hi @Miciah , 
Concerning [serviceaccount privileges](https://github.com/openshift/external-dns-operator/pull/99#discussion_r768298070), we still need them as the operator creates a serviceaccount in the operand namespace per externaldns, that will be used to run the deployment.

README.md
```diff
@@ -97,11 +102,12 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
    $ make index-image-push
    ```
 
-5. Create the `external-dns-operator` namespace:
-   ```sh
-   $ oc create ns external-dns-operator
+5. Prepare the operand namespace:
```
arjunrn, line 105:
```suggestion
5. Prepare the operator namespace:
```

dbaker-rh:
I've gone through and re-reviewed the PR.

I notice bundle/manifests still shows clusterrole and clusterrolebinding permissions that I think is now superfluous with the recent changes, and should be reviewed.

I made a few other comments in-line.

The worst of the (over) permissioning looks to have been taken care of.  This looks good to me.

README.md
Miciah: "external-dns" is the operand namespace, and `config/samples/extra-roles.yaml` gives the operator permission to access resources inside the operand namespace, so isn't "Prepare the operand namespace" accurate?

Miciah:
> Hi @Miciah , Concerning [serviceaccount privileges](https://github.com/openshift/external-dns-operator/pull/99#discussion_r768298070), we still need them as the operator creates a serviceaccount in the operand namespace per externaldns, that will be used to run the deployment.

If I understand correctly, `config/samples/extra-roles.yaml` gives the operator permission to manage serviceaccounts in the "external-dns" operand namespace whereas `config/rbac/operator_role.yaml` give the operator permission to manage serviceaccounts in the "external-dns-operator" operator namespace, and we really need the former but not the latter, right?

sherine-k:
> > Hi @Miciah , Concerning [serviceaccount privileges](https://github.com/openshift/external-dns-operator/pull/99#discussion_r768298070), we still need them as the operator creates a serviceaccount in the operand namespace per externaldns, that will be used to run the deployment.
> 
> If I understand correctly, `config/samples/extra-roles.yaml` gives the operator permission to manage serviceaccounts in the "external-dns" operand namespace whereas `config/rbac/operator_role.yaml` give the operator permission to manage serviceaccounts in the "external-dns-operator" operator namespace, and we really need the former but not the latter, right?

We actually need both. we need to be able to create a secret and a serviceaccount in the operator namespace, and we need to create a secret and serviceaccount for each externaldns CR

sherine-k:
> I've gone through and re-reviewed the PR.
> 
> I notice bundle/manifests still shows clusterrole and clusterrolebinding permissions that I think is now superfluous with the recent changes, and should be reviewed.
> 
> I made a few other comments in-line.
> 
> The worst of the (over) permissioning looks to have been taken care of. This looks good to me.

Hi @dbaker-rh
For the comments in-line, did you add anything? I missed it I think.
When you say that "this looks good" do you mean that we can deliver it like this and work on reducing the clusterrole/clusterrolebinding privileges later? Or still try to make that change before the freeze in order to get the green flag for the release? I got confused...
Thanks a lot for all your help

PR #101, opened 2021-12-13
Remove namespace filtering and unused CRD source

alebedev87:
On the way to v1beta1:
- Remove namespace filtering
- Remove unused CRD source


test/e2e/operator_test.go
```diff
@@ -377,19 +377,20 @@ func TestExternalDNSCustomIngress(t *testing.T) {
 	if err != nil && !errors.IsAlreadyExists(err) {
 		t.Fatalf("Failed to ensure namespace %s: %v", testIngressNamespace, err)
 	}
+
 	openshiftRouterName := "external-dns"
 	name := types.NamespacedName{Namespace: testIngressNamespace, Name: openshiftRouterName}
-	t.Logf("Create custom ingress controller %s/%s", name.Namespace, name.Name)
+	t.Log("Create custom ingress controller")
```
arjunrn, line 383: The namespace/name information is still useful when the test fails later after this step, not just in this step.

alebedev87: Just wanted to be consistent with the log messages from the other test. They usually say that a resource is about to be created without mentioning its name, I suppose it's fine as most of them are constants.

PR #102, opened 2021-12-14
BZ2032179: propagate proxy settings to operand deployment

alebedev87:
The following providers support `HTTP_PROXY` variables for sure:
- [Azure](https://github.com/Azure/azure-sdk-for-go/blob/84308db09c9d24417ef18de3c5000a6865005806/sdk/azcore/runtime/transport_default_http_client.go#L20)
- [Infoblox](https://github.com/infobloxopen/infoblox-go-client/blob/bf4fe6fb0414d46427237d3500a71a0b091ab267/connector.go#L139)

These ones I think they do but didn't find the direct evidence yet:
- AWS
- GCP

And BlueCat [doesn't seem to](https://github.com/kubernetes-sigs/external-dns/blob/f4850c4746db4ab4b8993f88a006017d2c0f4901/provider/bluecat/bluecat.go#L625-L627).


alebedev87:
As per the discussion with Arjun, filter the definitely unsupported DNS providers.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -178,6 +186,26 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
 
 	container.Args = append(container.Args, filterArgs...)
 	container.Args = append(container.Args, args...)
+
+	//
+	// ENV
+	//
+	env := []corev1.EnvVar{}
```
Connect2naga, line 193: this local variable not required, instead we can use directly `container.Env` in subsequent places 
like :
    ```
if utils.EnvProxySupportedProvider(b.externalDNS) {
		if val := os.Getenv(httpProxyEnvVar); val != "" {
			container.Env = append(container.Env, corev1.EnvVar{Name: httpProxyEnvVar, Value: val})
		}
		if val := os.Getenv(httpsProxyEnvVar); val != "" {
			container.Env = append(container.Env, corev1.EnvVar{Name: httpsProxyEnvVar, Value: val})
		}
		if val := os.Getenv(noProxyEnvVar); val != "" {
			container.Env = append(container.Env, corev1.EnvVar{Name: noProxyEnvVar, Value: val})
		}
	}

```

File: pkg/utils/utils.go [hunk 2]
```diff
@@ -28,3 +28,12 @@ func ManagedCredentialsProvider(e *operatorv1alpha1.ExternalDNS) bool {
 	}
 	return false
 }
+
+// EnvProxySupportedProvider returns true if the ExternalDNS provider supports the proxy settings via environment variables HTTP(S)_PROXY, NO_PROXY
+func EnvProxySupportedProvider(e *operatorv1alpha1.ExternalDNS) bool {
```
Connect2naga, line 33: isProxyEnvSupportedByProvider() will be good, because we are using this API for decision making.

pkg/operator/controller/externaldns/pod.go
alebedev87: Yeah, just kept the same pattern as for the args. Not really needed here, will change.

pkg/utils/utils.go
alebedev87: Yeah, your name is good. I just kept the same pattern as for the other function in this file: `ManagedCredentialsProvider`.

alebedev87:
Didn't manage to test on GCP as cluster-bot doesn't support `proxy` setting for GCP. But it's still a valid configuration according to @lihongan. Testing `external-dns` on local gives different result when HTTP_PROXY variables are set and not. Also google-api-go-client (used by GCP provider) uses the default golang `http.Client` which is aware of the proxy variables. So I think that there is a chance GCP provider supports them. Therefore re-adding GCP.

alebedev87:
As discussed with @Miciah.
A custom trusted CA [can be provided](https://docs.openshift.com/container-platform/4.9/networking/configuring-a-custom-pki.html#configuring-a-custom-pki) for the cluster configured with a global proxy. In that case the custom CA needs to be injected into the operand containers as a trusted one. Placing the CA into the "standard" `/etc/pki/ca-trust/extracted/` should be enough. 
Here is an example of how this can done for the operator (not operand) POD: [link](https://docs.openshift.com/container-platform/4.9/operators/admin/olm-configuring-proxy-support.html#olm-inject-custom-ca_olm-configuring-proxy-support).

For the sake of keeping this PR nice and small, I propose the custom CA injection to be addressed in a follow-up PR.

PR #103, opened 2021-12-14
Set dummy FQDNTemplate for Route source

alebedev87:


api/v1alpha1/externaldns_webhook.go
```diff
@@ -100,6 +100,11 @@ func (r *ExternalDNS) validateFilters() error {
 }
 
 func (r *ExternalDNS) validateHostnameAnnotationPolicy() error {
+	if r.Spec.Source.Type == SourceTypeRoute {
```
arjunrn, line 103: Should we reject the CR when source type is `Route` and the hostname annotation policy is specified?

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -1163,6 +1163,7 @@ func TestDesiredExternalDNSDeployment(t *testing.T) {
 							"--registry=txt",
 							"--log-level=debug",
 							"--ignore-hostname-annotation",
+							`--fqdn-template={{""}}`,
```
arjunrn, line 1166: This seems kind of wrong. Like what does this mean? Why pass it at all?

pkg/operator/controller/externaldns/pod.go
```diff
@@ -165,6 +165,14 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
 
 	if len(b.externalDNS.Spec.Source.FQDNTemplate) > 0 {
 		args = append(args, fmt.Sprintf("--fqdn-template=%s", strings.Join(b.externalDNS.Spec.Source.FQDNTemplate, ",")))
+	} else {
+		// ExternalDNS needs FQDNTemplate if the hostname annotation is ignored even for Route source.
```
arjunrn, line 169: Ok, I see why now.

api/v1alpha1/externaldns_webhook.go
alebedev87: You mean when the hostname annotation policy is `Allow`? The route source [can use the hostname annotation](https://github.com/kubernetes-sigs/external-dns/blob/a3a9d4f27c1a02411585223059a5da5dcd837c31/source/source.go#L142).

arjunrn: In that case when the policy is `Allow` we won't validate that the the `fqdnTemplate` is set altogether. Is that what we want?

alebedev87: Well, yes, if the policy is `Allow`, the user can specify the custom hostname in the route object, so `FQDNTemplate` is not really needed.
But I believe it's still the valid option to have both. We already had a similar discussion [here](https://github.com/openshift/external-dns-operator/pull/26#discussion_r723310084).

alebedev87:
Rebased to fix the merge conflict.

PR #104, opened 2021-12-15
Don't create credentials request if the secret is given in CR

alebedev87:


alebedev87:
AWS test timed out, the set of received IPs was a bigger than the expected:
```
    operator_test.go:300: Getting IPs of service's load balancer
    operator_test.go:333: Loadbalancer's IP(s): map[34.215.81.125:{}]
    operator_test.go:341: Looking for DNS record in nameserver: ns-1583.awsdns-05.co.uk
    operator_test.go:351: Got IPs: [34.215.81.125 54.187.28.21]
```

alebedev87:
/assign @sherine-k 

This PR is what you noticed in one of your tests.

test/e2e/operator_test.go
```diff
@@ -348,13 +348,20 @@ func TestExternalDNSRecordLifecycle(t *testing.T) {
...
 			}
-			for _, ip := range ips {
-				if _, ok := serviceIPs[ip]; !ok {
+			// all expected IPs should be in the received IPs
+			// but these 2 sets are not necessary equal
+			for ip := range serviceIPs {
+				if _, found := gotIPs[ip]; !found {
 					return false, nil
 				}
 			}
```
alebedev87, line 367: @arjunrn: this is to fix [this issue](https://github.com/openshift/external-dns-operator/pull/104#issuecomment-995174939). It doesn't happen all the time but when it does, the test times out.

sherine-k (approved):
Thanks @alebedev87 !

test/e2e/operator_test.go
arjunrn: I would like to see this change in a separate PR if possible. Also `gotIPs` could be called `loadbalancerIPs`.

PR #105, opened 2021-12-15
Dummy PR for API review

alebedev87:


alebedev87:
/hold

Dummy just for the review of the API changes we made since [the last one](https://github.com/openshift/external-dns-operator/pull/12).

api/v1alpha1/externaldns_types.go
```diff
@@ -149,7 +155,7 @@ type ExternalDNSDomainUnion struct {
 	// foo.my-app.my-cluster-domain.com
 	//
 	// +optional
-	Name *string `json:"names,omitempty"`
+	Name *string `json:"name,omitempty"`
```
deads2k, line 158: you never shipped this API, right?

api/v1alpha1/externaldns_types.go
```diff
@@ -355,15 +357,9 @@ type ExternalDNSInfobloxProviderOptions struct {
...
-// desired secret in a specific namespace.
-type NamespacedSecretReference struct {
-	// Namespace is the namespace of the secret.
-	//
-	// +kubebuilder:validation:Required
-	// +required
-	Namespace string `json:"namespace"`
-
+// SecretReference contains the information to let you locate the desired secret.
+// Secret is expected to be in the operator namespace.
```
deads2k, line 361: Secret is required to be in the operator namespace

deads2k: This is weird actually.  You have a cluster scoped resource that refers to a secret in a particular namespace that can change depending on where the operator is installed?

Can you ever have more than one operator installed?  Do you know what namespace this is?  operator or operand?

api/v1alpha1/externaldns_types.go
```diff
@@ -405,8 +401,23 @@ type ExternalDNSSource struct {
...
 	// +kubebuilder:default:=Ignore
-	// +required
+	// +optional
 	HostnameAnnotationPolicy HostnameAnnotationPolicy `json:"hostnameAnnotation"`
+
+	// FQDNTemplate sets a templated string that's used to generate DNS names
+	// from sources that don't define a hostname themselves.
+	// Multiple global FQDN templates are possible.
+	//
+	// Should not be empty when HostnameAnnotationPolicy is set to Ignore.
```
deads2k, line 411: should not or must not?

deads2k: Is this honored when HostnameAnnotationPolicy is not Ignore?

api/v1alpha1/externaldns_types.go
```diff
@@ -501,6 +497,13 @@ type ExternalDNSServiceSourceOptions struct {
 	ServiceType []corev1.ServiceType `json:"serviceType,omitempty"`
 }
 
+type ExternalDNSOpenShiftRouteOptions struct {
+	// If source is openshift-route then you can pass the ingress controller name. Based on this name external-dns will select the respective router.
+	// +kubebuilder:validation:Required
+	// +required
+	RouterName string `json:"routerName"`
```
deads2k, line 504: How do you know the router name from a route instance?

api/v1alpha1/externaldns_types.go
alebedev87: Right.

api/v1alpha1/externaldns_types.go
alebedev87: Should not. Because it _must be_ not empty for the source of type `Service` and _can be_ empty for the source type `OpenShiftRoute`. We have a corresponding [validating webhook check](https://github.com/openshift/external-dns-operator/blob/main/api/v1alpha1/externaldns_webhook.go#L102).

When `HostnameAnnotationPolicy` is not `Ignore` this field can be empty regardless of the source type.

api/v1alpha1/externaldns_types.go
alebedev87: It's in the route's status: [route api](https://github.com/openshift/api/blob/be1be0e89115702f8b508d351c4f5c9a16e5ae95/route/v1/types.go#L166).

api/v1alpha1/externaldns_types.go
alebedev87: Yes, the namespace can change depending on where the operator is installed. The cluster admin will have to give this namespace to OLM at the installation. The secret will have to be there before the CR is created.

> Can you ever have more than one operator installed?

Technically, it's possible. But the idea is 1 operator for the whole cluster.

> Do you know what namespace this is? operator or operand?

The secret which was created by the cluster admin in the operator namespace will be copied into the operand namespace by the operator.

Also, an important note: if the secret is not referenced in the CR, the operator will request one from the cloud credentials operator (if the underlying platform is OpenShift).

deads2k: > Yes, the namespace can change depending on where the operator is installed. The cluster admin will have to give this namespace to OLM at the installation. The secret will have to be there before the CR is created.

This level of flexibility seems unnecessary and undesireable.  Why should the namespace be configurable?

api/v1alpha1/externaldns_types.go
deads2k: > Should not. Because it must be not empty for the source of type Service and can be empty for the source type OpenShiftRoute. We have a corresponding validating webhook check.

You should be crisp about the cases.

api/v1alpha1/externaldns_types.go
```diff
@@ -421,37 +432,22 @@ type ExternalDNSSourceUnion struct {
 	// +required
 	Type ExternalDNSSourceType `json:"type"`
 
-	// AnnotationFilter describes an annotation filter
-	// used to filter which source instance resources
-	// ExternalDNS publishes records for.
-	// The annotation filter uses label selector semantics
-	// against source resource annotations.
+	// LabelFilter specifies a label selector used to filter which source instance resources ExternalDNS publishes
+	// records for. The filter uses label selector semantics against source resource labels.
```
deads2k, line 436: Does nil mean match all or match none.

api/v1alpha1/externaldns_types.go
```diff
@@ -421,37 +432,22 @@ type ExternalDNSSourceUnion struct {
...
-	// https://github.com/kubernetes-sigs/external-dns/blob/master/docs/contributing/crd-source.md
-	// for more information about the requirements for ExternalDNS Source CRD requirements.
-	//
+	// OpenShiftRoute source configuration options for specifying ingress controller names.
+	// +kubebuilder:validation:Optional
 	// +optional
-	CRD *ExternalDNSCRDSourceOptions `json:"crd,omitempty"`
-
-	// OpenShiftRoute source currently has no unique configuration options.
+	OpenShiftRoute *ExternalDNSOpenShiftRouteOptions `json:"openshiftRouteOptions,omitempty"`
```
deads2k, line 450: Since this an openshift repo, this seems ok, but do you have a sense for how you will handle requests for extending this to other types?

deads2k:
I suggest clarifying a few bits in the doc, but the API looks ok. The namespace thing rubs me the wrong way, but I can't exactly how it will burn you in the future.

api/v1alpha1/externaldns_types.go
alebedev87: > Why should the namespace be configurable?

Technically the operator (its manifests) is agnostic to the namespace in which it will be installed. But I believe this is an OLM requirement (`OperatorGroup.spec.targetNamespaces`).

api/v1alpha1/externaldns_types.go
alebedev87: `nil` label filter means "filter nothing", so "get all source objects of the cluster".

Miciah:
This PR incorporates David's feedback:  https://github.com/openshift/external-dns-operator/pull/109

Miciah:
#109 merged, and I believe it addressed all the comments in this PR, so I'll close this PR.
/close

PR #108, opened 2021-12-16
Take away operator privileges on clusterroles and clusterrolebindings…

sherine-k:
…, replacing them by a manual procedure by clusteradmin


README.md
```diff
@@ -19,6 +19,8 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
    oc create ns external-dns
    oc apply -f config/samples/extra-roles.yaml
    ```
+   **Note**:  For OpenShift clusters use `config/samples/extra-roles-openshift.yaml` instead.
```
alebedev87, line 22: Why the extra roles are in `config/samples/` not `config/rbac/` ?

sherine-k: Oh I was just afraid they would get confused with ones that go under the bundle. But I can move them there.

config/samples/extra-roles-openshift.yaml
```diff
@@ -0,0 +1,98 @@
...
+metadata:
+  name: external-dns
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: external-dns
+subjects:
+  - kind: Group
+    name: system:serviceaccounts:external-dns
+    namespace: external-dns
```
alebedev87, line 98: Can you please add the new line at the end of the file?

pkg/operator/controller/externaldns/rbac.go
```diff
@@ -1,287 +0,0 @@
...
-			for _, resource := range rule.Resources {
-				sortedVerbs := make([]string, len(rule.Verbs))
-				copy(sortedVerbs, rule.Verbs)
-				sort.Strings(sortedVerbs)
-				m[apiGroup+"/"+resource] = sortedVerbs
-			}
-		}
-	}
-	return m
-}
```
alebedev87, line 287: :sob:

README.md
alebedev87: We have full control over what's going in the bundle thanks to `olm-manifests` Makefile target. We can simply not copy the extra role files in in the bundle.

alebedev87: Sorry, my bad, you are right there was an `if` for the route source. Well, yes, may be it makes sense to separate them. But just for a note: keeping them in one file won't harm.

pkg/operator/controller/externaldns/rbac.go
sherine-k: :cry:

sherine-k:
/assign @Miciah 
@lmzuccarelli @miheer, if you think we cannot nail this by 12EST cause @Miciah would be busy, would you mind taking over to review and lgtm?

lmzuccarelli:
> /assign @Miciah @lmzuccarelli @miheer, if you think we cannot nail this by 12EST cause @Miciah would be busy, would you mind taking over to review and lgtm?

Yes absolutely !!!

test/e2e/operator_test.go
```diff
@@ -159,6 +160,14 @@ func TestMain(m *testing.M) {
 		fmt.Printf("Failed to create rolebinding external-dns-operator in ns %s: %v\n", operandNamespace, err)
 	}
 
+	if err = ensureOperandClusterRoleBinding(); err != nil {
+		fmt.Printf("Failed to create clusterrolebinding external-dns : %v\n", err)
```
miheer, line 164: I think you can pass t *testing.T in the function and then add t.Errorf rather than using fmt.Printf

test/e2e/operator_test.go
```diff
@@ -159,6 +160,14 @@ func TestMain(m *testing.M) {
...
+	}
+
+	if err = ensureOperandClusterRole(); err != nil {
+		fmt.Printf("Failed to create clusterrole external-dns in : %v\n", err)
```
miheer, line 168: I think you can pass t *testing.T in the function and then add t.Errorf rather than using fmt.Printf

pkg/operator/operator.go
```diff
@@ -61,7 +61,6 @@ type Operator struct {
 // +kubebuilder:rbac:groups=externaldns.olm.openshift.io,resources=externaldnses/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=externaldns.olm.openshift.io,resources=externaldnses/finalizers,verbs=update
 // +kubebuilder:rbac:groups="",resources=services;endpoints;pods;nodes,verbs=get;list;watch
-// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
```
miheer, line 64: Did you run make manifests and make generate ?

bundle/manifests/external-dns-operator_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -71,19 +71,6 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
```
miheer, line 74: Was this reflected after running make bundle-image-build ?

test/e2e/operator_test.go
```diff
@@ -430,6 +467,27 @@ func ensureOperandRoleBinding() error {
...
+			Name: operandRbacRsrcName,
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     operandRbacRsrcName,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      rbacv1.GroupKind,
```
miheer, line 482: Can't we use kind: ServiceAccount  rather than using Group ?

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-binding-examples

config/rbac/extra-roles.yaml
```diff
@@ -50,3 +50,49 @@ subjects:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: external-dns
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: external-dns
+subjects:
+  - kind: Group
```
miheer, line 96: Can't we use kind: ServiceAccount rather than using Group ?

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-binding-examples

miheer (changes requested):

miheer:
@sherine-k  as discussed please check the suggestions in a followup PR.

pkg/operator/operator.go
sherine-k: Yes

config/rbac/extra-roles.yaml
Miciah: Miheer, in the examples you linked, I see the following:
> For all service accounts in the "dev" group in the "development" namespace:
> 
> ```yaml
> subjects:
> - kind: Group
>   name: system:serviceaccounts:dev
>   apiGroup: rbac.authorization.k8s.io
>   namespace: development

The only example that uses `kind: ServiceAccount` is for specifying a single ServiceAccount:

> For the default service account in the "kube-system" namespace:
> 
> ```yaml
> subjects:
> - kind: ServiceAccount
>   name: default
>   namespace: kube-system

test/e2e/operator_test.go
Miciah: Passing in `t` and to the test helper and then using `t.Helper()`, `t.Errorf()` etc. inside the helper is generally good practice.  However, in this case, the helper is used by the test runner which calls `m.Run()`, so I don't think there is a `t` that could be passed in to the helpers.

bundle/manifests/external-dns-operator_rbac.authorization.k8s.io_v1_clusterrole.yaml
sherine-k: this was reflected from running `make olm-manifests` really, after removing kubebuilder rbac annotations on `operator.go`

test/e2e/operator_test.go
sherine-k: Basically using this groups allows us to cover all existing and future serviceaccounts in external-dns namespace, without having to guess their names in advance. 
The alternative, using `kind: ServiceAccount` would force us to update the list of subjects in the clusterrolebinding, on every event occurring on the externalDNS instances

test/e2e/operator_test.go
sherine-k: Thanks @Miciah

miheer: @Miciah  @sherine-k   OK I am not sure what you mean but I can here it is added https://github.com/sherine-k/external-dns-operator/blob/7df0b56243b1fdf04e3bf6121a24d2f4eae76819/test/e2e/operator_test.go#L191

here Fatalf is used https://github.com/sherine-k/external-dns-operator/blob/7df0b56243b1fdf04e3bf6121a24d2f4eae76819/test/e2e/operator_test.go#L237

I think we shall exit and not just Printf and error

test/e2e/operator_test.go
miheer: Hmm, OK so basically you want all service accounts in external-dns ns to get that cluster role? Does it sound right ?

config/rbac/extra-roles.yaml
miheer: @Miciah 

subjects:
- kind: ServiceAccount
  name: system:serviceaccounts:external-dns
  namespace: external-dns

@Miciah  I was thinking something like this. Why do we want all the service accounts in operands ns to get bind to a cluster role ?

test/e2e/operator_test.go
Miciah: Regarding `t`, here's a simple question:  What `t`?  

Calling `os.Exit(1)` if the ensure function returns a non-nil error value might make sense.  We can also introduce a polling loop around the create operation if we need to make the test more resilient.  I wouldn't worry too much though unless we see the ensure steps fail.

test/e2e/operator_test.go
Miciah: Yes, that is what we agreed on yesterday.

config/rbac/extra-roles.yaml
Miciah: We want all serviceaccounts in the operand's namespace to be bound to the clusterrole because that is the only way we can avoid giving the operator broad permissions to create and delete clusterrolebindings.

test/e2e/operator_test.go
miheer: Hmm OK whatever you say.

miheer: Oh OK BTW I am very sorry I just saw that this function is present in TestMain function which has *testing.M

miheer:
@sherine-k   there is no need for a followup PR

PR #109, opened 2021-12-17
CRD fixes from #105 API review

Miciah:
#### CRD: Add missing +'s in kubebuilder tags

* `api/v1alpha1/externaldns_types.go` (`ExternalDNSInfobloxProviderOptions`): Change "kubebuilder:validation:Required" to "+kubebuilder:validation:Required".
* `bundle/manifests/externaldns.olm.openshift.io_crd.yaml`:
* `config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml`: Regenerate.

#### CRD: Fix `HostnameAnnotationPolicy` kubebuilder tag

The `HostnameAnnotationPolicy` field is optional in some cases, and it has the `+optional` tag, so its kubebuilder tag should also indicate that the field is optional.

* `api/v1alpha1/externaldns_types.go` (`ExternalDNSSource`): Replace `+kubebuilder:validation:Required` with `+kubebuilder:validation:Optional`.

#### CRD: Add `+kubebuilder:validation:Optional` tags

Add `+kubebuilder:validation:Optional` tags to all optional fields for consistency.

Before this change, some optional fields had both the `+optional` tag and the `+kubebuilder:validation:Optional` tag, and some fields had only the `+optional` tag.

* `api/v1alpha1/externaldns_types.go` (`ExternalDNSSpec`, `ExternalDNSDomainUnion`, `ExternalDNSProvider`, `ExternalDNSGCPProviderOptions`, `ExternalDNSSourceUnion`, `ExternalDNSCRDSourceOptions`: Add `+kubebuilder:validation:Optional` tags to optional fields.

#### CRD: Improve godoc based on API review

* `api/v1alpha1/externaldns_types.go` (`SecretReference`): Change "expected" to "required".
(`ExternalDNSSource`): Clarify exactly when `FQDNTemplate` must be specified and when it may be omitted.
(`ExternalDNSSourceUnion`): Describe the behavior for a null or empty label selector.
(`ExternalDNSOpenShiftRouteOptions`): Clarify that the router name is reported in route status.
* `bundle/manifests/externaldns.olm.openshift.io_crd.yaml`:
* `config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml`: Regenerate.

#### CRD: Fix grammar error in `OpenShiftRoute` godoc

* `api/v1alpha1/externaldns_types.go` (`ExternalDNSSourceUnion`): Rewrite the godoc as a complete sentence that describes the field in general terms.
* `bundle/manifests/externaldns.olm.openshift.io_crd.yaml`:
* `config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml`: Regenerate.

### DIFF
--- a/api/v1alpha1/externaldns_types.go
+++ b/api/v1alpha1/externaldns_types.go
@@ -368,7 +368,7 @@ type ExternalDNSInfobloxProviderOptions struct {
 }
 
 // SecretReference contains the information to let you locate the desired secret.
-// Secret is expected to be in the operator namespace.
+// Secret is required to be in the operator namespace.
 type SecretReference struct {
 	// Name is the name of the secret.
 	//
@@ -418,7 +418,10 @@ type ExternalDNSSource struct {
 	// from sources that don't define a hostname themselves.
 	// Multiple global FQDN templates are possible.
 	//
-	// Should not be empty when HostnameAnnotationPolicy is set to Ignore.
+	// This field must be specified with a nonempty value if the source type
+	// is Service or CRD and HostnameAnnotationPolicy is set to Ignore.  The
+	// field value may be omitted or empty if HostnameAnnotationPolicy is
+	// set to Allow or if the source type is OpenShiftRoute.
 	//
 	// Provided templates should follow the syntax defined for text/template Go package,
 	// see https://pkg.go.dev/text/template.
@@ -442,8 +445,11 @@ type ExternalDNSSourceUnion struct {
 	// +required
 	Type ExternalDNSSourceType `json:"type"`
 
-	// LabelFilter specifies a label selector used to filter which source instance resources ExternalDNS publishes
-	// records for. The filter uses label selector semantics against source resource labels.
+	// LabelFilter specifies a label selector for filtering the objects for
+	// which ExternalDNS publishes records. The filter uses label selector
+	// semantics against object labels.  Specifying a null or empty label
+	// selector causes ExternalDNS to publish records for all objects of the
+	// source type resource.
 	//
 	// +kubebuilder:validation:Optional
 	// +optional
@@ -510,7 +516,11 @@ type ExternalDNSServiceSourceOptions struct {
 }
 
 type ExternalDNSOpenShiftRouteOptions struct {
-	// If source is openshift-route then you can pass the ingress controller name. Based on this name external-dns will select the respective router.
+	// RouterName is the name of a router (AKA ingress controller) as
+	// reported in Route.status.ingress[].routerName.  External-dns will use
+	// the canonical hostname of the router identified by this name when
+	// publishing records for a given route.
+	//
 	// +kubebuilder:validation:Required
 	// +required
 	RouterName string `json:"routerName"`
--- a/api/v1alpha1/externaldns_types.go
+++ b/api/v1alpha1/externaldns_types.go
@@ -59,6 +59,7 @@ type ExternalDNSSpec struct {
 	// Excluding DNS records that were previous included via a resource update
 	// will *not* result in the original DNS records being deleted.
 	//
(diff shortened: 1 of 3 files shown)


api/v1alpha1/externaldns_types.go
```diff
@@ -418,7 +418,10 @@ type ExternalDNSSource struct {
 	// from sources that don't define a hostname themselves.
 	// Multiple global FQDN templates are possible.
 	//
-	// Should not be empty when HostnameAnnotationPolicy is set to Ignore.
+	// This field must be specified with a nonempty value if the source type
+	// is Service or CRD and HostnameAnnotationPolicy is set to Ignore.  The
+	// field value may be omitted or empty if HostnameAnnotationPolicy is
+	// set to Allow or if the source type is OpenShiftRoute.
```
Miciah, line 424: @deads2k, is this good an crispy?

api/v1alpha1/externaldns_types.go
```diff
@@ -442,8 +445,10 @@ type ExternalDNSSourceUnion struct {
 	// +required
 	Type ExternalDNSSourceType `json:"type"`
 
-	// LabelFilter specifies a label selector used to filter which source instance resources ExternalDNS publishes
-	// records for. The filter uses label selector semantics against source resource labels.
+	// LabelFilter specifies a label selector used to filter which source
+	// instance resources ExternalDNS publishes records for. The filter uses
+	// label selector semantics against source resource labels.  A null or
+	// empty label selector selects all objects with matching kind.
```
Miciah, line 451: Can someone double-check that this text is accurate?

api/v1alpha1/externaldns_types.go
```diff
@@ -510,7 +515,11 @@ type ExternalDNSServiceSourceOptions struct {
 }
 
 type ExternalDNSOpenShiftRouteOptions struct {
-	// If source is openshift-route then you can pass the ingress controller name. Based on this name external-dns will select the respective router.
+	// RouterName is the name of a router (AKA ingress controller) as
+	// reported in Route.status.ingress[].routerName.  External-dns will use
+	// the canonical hostname of the router identified by this name when
+	// publishing records for a given route.
```
Miciah, line 523: @deads2k, I tried to clarify this per your review.

api/v1alpha1/externaldns_types.go
deads2k: > @deads2k, is this good an crispy?

yes, thank you.

api/v1alpha1/externaldns_types.go
deads2k: > @deads2k, I tried to clarify this per your review.

looks good.

deads2k: actually, are they guaranteed to all be the same?

Miciah: I don't understand your question.  When external-dns publishes a record for a route, it publishes a CNAME record with the router's canonical hostname as the target.  This field in the externaldns CRD specifies the name of the router.  The canonical hostname should be the same for all routes exposed by a given external-dns instance.  (If two routes are served by the same router but have different canonical hostnames in their status, then something weird has happened.)

api/v1alpha1/externaldns_types.go
sherine-k: Not sure this is better? just some ideas
```suggestion
	//  LabelFilter specifies a label selector used to filter which instance(s) 
	// of the specified source type ExternalDNS publishes records for.
	// The filter uses label selector semantics matched against labels defined on 
	// instances of the specified source type.  A null or empty label selector 
	// selects all objects with matching source type.
```

sherine-k (approved):

PR #113, opened 2022-01-18
Bug 2041999: inject trusted CA to operand deployment

alebedev87:
This PR allows the operator to inject trusted CA(s) into ExternalDNS (operand) containers.

**What's inside**

- New flag to pass the name of the configmap containing the CA (`cmd/external-dns-operator/main.go`)
- New controller to copy this configmap from the operator into the operand namespace (`pkg/operator/controller/ca-configmap/`)
- Updated desired ExternalDNS deployment. Mounts the trusted CA file now (`pkg/operator/controller/externaldns/pod.go`)
- Updated RBAC files (`configmaps` resources)
- New `pkg/operator/controller/utils` package for things used in all the controllers, mainly for `EventCollector` used in all the controllers' tests (`pkg/operator/controller/utils/`, `pkg/operator/controller/utils/test/`) 
- New doc for the proxy support (`docs/proxy.md`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
 
 7. Create the `Catalogsource` object:
 
-   ```yaml
+   ```bash
    $ cat <<EOF | oc apply -f -
    apiVersion: operators.coreos.com/v1alpha1
    kind: CatalogSource
@@ -142,9 +142,13 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
    EOF
    ```
 
-8. Create a subscription object to install the Operator:
-   
-    ```yaml
+8. Create the operator namespace:
+    ```bash
+    oc create namespace external-dns-operator
+    ```
+
+9. Create the `Subscription` object:
+    ```bash
     cat <<EOF | oc apply -f -
     apiVersion: operators.coreos.com/v1alpha1
     kind: Subscription
@@ -158,7 +162,22 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
       sourceNamespace: openshift-marketplace
     EOF
     ```
-    **Note**: You can install the `ExternalDNS` Operator through the web console: Navigate to  `Operators` -> `OperatorHub`, search for the `ExternalDNS operator`,  and install it in the `external-dns-operator` namespace.
+
+10. Create the `OperatorGroup` object:
+    ```bash
+    cat <<EOF | oc apply -f -
+    apiVersion: operators.coreos.com/v1
+    kind: OperatorGroup
+    metadata:
+      name: external-dns-operator
+      namespace: external-dns-operator
+    spec:
+      targetNamespaces:
+      - external-dns-operator
+    EOF
+    ```
+
+**Note**: The steps starting from the 8th can be replaced with the following actions in the web console: Navigate to  `Operators` -> `OperatorHub`, search for the `ExternalDNS Operator`,  and install it in the `external-dns-operator` namespace.
 
 ### Running end-to-end tests manually
 
@@ -179,3 +198,7 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
    ```sh
    $ make test-e2e
    ```
+
+### Proxy support
+
+[Configuring proxy support for ExternalDNS Operator](./docs/proxy.md)
--- a/bundle/manifests/external-dns-operator_clusterserviceversion.yaml
+++ b/bundle/manifests/external-dns-operator_clusterserviceversion.yaml
@@ -32,6 +32,7 @@ metadata:
     createdAt: 2021/09/28
     containerImage: quay.io/openshift/origin-external-dns-operator:latest
     operatorframework.io/suggested-namespace: external-dns-operator
+    operators.openshift.io/infrastructure-features: '["disconnected","proxy-aware"]'
   name: external-dns-operator.v0.0.1
   namespace: external-dns-operator
 spec:
@@ -87,6 +88,7 @@ spec:
               - args:
                 - --metrics-bind-address=127.0.0.1:8080
                 - --externaldns-image=$(RELATED_IMAGE_EXTERNAL_DNS)
+                - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
                 env:
                 - name: OPERATOR_NAMESPACE
                   valueFrom:
@@ -94,6 +96,8 @@ spec:
                       fieldPath: metadata.namespace
                 - name: RELATED_IMAGE_EXTERNAL_DNS
                   value: k8s.gcr.io/external-dns/external-dns:v0.10.2
+                - name: TRUSTED_CA_CONFIGMAP_NAME
+                  value: ""
                 image: quay.io/openshift/origin-external-dns-operator:latest
                 name: operator
                 ports:
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -1,0 +1,86 @@
+# Configuring proxy support for ExternalDNS Operator
+
+The ExternalDNS Operator can work in an environment with a cluster-wide egress proxy set up. There is some configuration to be done to make the operator aware of the proxy:
+- operator container's environment has to be populated with one (or all) of the following variables: `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`.
+- if the proxy uses some custom TLS certificate authority (CA), it has to be put into a configmap and passed to the operator via the `--trusted-ca-configmap` flag.
+
+The ExternalDNS Operator doesn't need the proxy settings on its own because it doesn't interact with DNS providers. However, it has to propagate the proxy settings and CA certificate down to its operand: the ExternalDNS instance.
+
+## Contents
+
+- [Instructions](#instructions)
+- [OpenShift instructions](#openshift-instructions)
+
+## Instructions
+
+### Proxy settings
+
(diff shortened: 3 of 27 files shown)


docs/proxy.md
```diff
@@ -0,0 +1,94 @@
...
+    ```
+
+2. Patch the operator's deployment to pass the before created configmap:
+    ```bash
+    kubectl -n external-dns-operator patch deployment external-dns-operator -p '{"spec":{"template":{"spec":{"containers":[{"name":"external-dns-operator","args":["--trusted-ca-configmap=trusted-ca"]}]}}}}'
+    ```
+
+## OpenShift instructions
+
+If a global proxy is configured on the OpenShift cluster, OLM automatically configures Operators with cluster-wide proxy settings. `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` environment variables are added to the ExternalDNS Operator's deployment.
```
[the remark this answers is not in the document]
alebedev87: Then this CA has to be passed to the operator using the flag developed in this PR: `--trusted-ca-configmap`. The configmap is up to the end user to create.
OpenShift provides a [nice label](https://docs.openshift.com/container-platform/4.9/operators/admin/olm-configuring-proxy-support.html#olm-inject-custom-ca_olm-configuring-proxy-support ) which brings the trusted CA into the configmap of your choice.

Miciah (commented):
Do we need to worry about cleaning up the configmap if the user unsets `TRUSTED_CA_CONFIGMAP_NAME` after the operator has created the target configmap?

README.md
```diff
@@ -158,7 +162,22 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
...
+    metadata:
+      name: external-dns-operator
+      namespace: external-dns-operator
+    spec:
+      targetNamespaces:
+      - external-dns-operator
+    EOF
+    ```
+
+**Note**: The steps starting from the 8th can be replaced with the following actions in the web console: Navigate to  `Operators` -> `OperatorHub`, search for the `ExternalDNS Operator`,  and install it in the `external-dns-operator` namespace.
```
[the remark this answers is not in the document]
alebedev87: Yes, it's unrelated. I noticed that if you follow the readme instructions the operator doesn't get deployed cause the `OperatorGroup` has to tell in which namespace it should go. Seems like most of the developers used the webconsole for the deployment which creates the scbscription and operatorgroup for you.

pkg/operator/controller/ca-configmap/controller_test.go
```diff
@@ -0,0 +1,213 @@
...
+			cl := fake.NewClientBuilder().WithScheme(test.Scheme).WithRuntimeObjects(tc.existingObjects...).Build()
+
+			r := &reconciler{
+				client: cl,
+				scheme: test.Scheme,
+				config: tc.inputConfig,
+				log:    zap.New(zap.UseDevMode(true)),
+			}
+
+			c := test.NewEventCollector(t, cl, managedTypesList)
```
miheer, line 123: managedTypesList  so we everytime we pass managedTypesList [&ConfigMapList{ListMeta:{   <nil>},Items:[]ConfigMap{},}]  But I am not getting then how iteration happens on this list.

alebedev87: `managedTypesList` is a list of types to be watched, we need only types without any items as they will be passed to `Watch()` function later in `EventCollector.Start()`: [link](https://github.com/openshift/external-dns-operator/blob/96aebfb5f7ac49e1eb849b48d82112e9c9e51885/pkg/operator/controller/utils/test/events.go#L114).

bundle/manifests/external-dns-operator_clusterserviceversion.yaml
```diff
@@ -32,6 +32,7 @@ metadata:
     createdAt: 2021/09/28
     containerImage: quay.io/openshift/origin-external-dns-operator:latest
     operatorframework.io/suggested-namespace: external-dns-operator
+    operators.openshift.io/infrastructure-features: '["disconnected","proxy-aware"]'
```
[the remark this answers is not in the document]
alebedev87: Yeah, this seems to be left for the operator developer discretion: to call the operator ready for the proxy or not.
BTW, talking about the automation, you comment made me reread the [documentation](https://docs.openshift.com/container-platform/4.9/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-manual-annotations_osdk-generating-csvs) and it said:

> Required environment variables are passed down to Operands for managed workloads.

And I was happy see this as I had to do it myself in the previous PR but the people from the portfolio enablement channel [brought me down to earth](https://coreos.slack.com/archives/C01GJKL46A2/p1642680775021300?thread_ts=1642680300.021200&cid=C01GJKL46A2).

docs/proxy.md
```diff
@@ -0,0 +1,94 @@
...
+#### Operator already running
+
+1. Create a configmap containing PEM encoded CA ceritificate in `external-dns-operator` namespace:
+    ```bash
+    kubectl -n external-dns-operator create configmap trusted-ca --from-file=ca-bundle.crt=/path/to/ca/certificate.pem
+    ```
+
+2. Patch the operator's deployment to pass the before created configmap:
+    ```bash
+    kubectl -n external-dns-operator patch deployment external-dns-operator -p '{"spec":{"template":{"spec":{"containers":[{"name":"external-dns-operator","args":["--trusted-ca-configmap=trusted-ca"]}]}}}}'
```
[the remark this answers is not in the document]
alebedev87: Good catch! Didn't know about this, changed to `json` patch.

pkg/operator/controller/ca-configmap/controller_test.go
alebedev87: Clarified this in Slack. The confusion was in the point that the watch type is a list object which is not used a container for items.

docs/proxy.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### For running operator
+
+1. Create a configmap with CA ceritificate in `external-dns-operator` namespace:
+```bash
+oc -n external-dns-operator create configmap trusted-ca
+oc -n external-dns-operator label cm trusted-ca config.openshift.io/inject-trusted-cabundle=true
+```
+
+2. Add `spec.config.env` with the name of the before created configmap to your subscription created by OperatoHub:
```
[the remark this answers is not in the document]
alebedev87: Patch command added instead.

pkg/operator/controller/ca-configmap/controller_test.go
```diff
@@ -0,0 +1,213 @@
...
+		Data: map[string]string{
+			"ca-bundle.crt": "---pem encoded certificate---",
+		},
+	}
+}
+
+func testTargetConfigMap() *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testSrcConfigMapName,
```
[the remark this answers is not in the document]
alebedev87: Good point! Indeed it doesn't. Modified the collector and tests to check for the extraneous events too.

pkg/operator/controller/ca-configmap/ca_configmap.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2021.
```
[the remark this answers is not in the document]
Miciah: The other new files still have "Copyright 2021."  (It isn't all that important though.)

pkg/operator/controller/utils/test/events.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// CollectFatal collects the given number of events. Fails the test if the timeout is reached while not all events are received.
+func (c *EventCollector) CollectFatal(num int, timeout time.Duration) []Event {
+	c.T.Helper()
+
+	return c.collectWithAction(num, timeout, func() {
+		c.T.Fatalf("timed out waiting for all expected events")
+	})
+}
```
Miciah, line 154: This seems to be unused.

README.md
```diff
@@ -158,7 +162,22 @@ The following procedure describes how to deploy the `ExternalDNS` Operator for A
       sourceNamespace: openshift-marketplace
     EOF
     ```
-    **Note**: You can install the `ExternalDNS` Operator through the web console: Navigate to  `Operators` -> `OperatorHub`, search for the `ExternalDNS operator`,  and install it in the `external-dns-operator` namespace.
+
+10. Create the `OperatorGroup` object:
+    ```yaml
```
Miciah, line 167:
```suggestion
    ```bash
```

pkg/operator/controller/ca-configmap/controller.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Config holds all the things necessary for the controller to run.
+type Config struct {
+	SourceNamespace string
+	TargetNamespace string
+	CAConfigMapName string
+}
+
+type reconciler struct {
+	scheme *runtime.Scheme
```
Miciah, line 54: Come to think of it, we don't use `scheme` either, do we?

alebedev87: you are right, no, we don't. removed it.

pkg/operator/controller/utils/test/events.go
alebedev87: yeah, I thought that maybe useful for someone to fail when the timeout is expired. But it's unpractical to keep all these functions if they are no use. Refactored back to 1 single function.

Miciah:
Thanks!
/lgtm

PR #114, opened 2022-01-25
WIP: Propose escalate and bind for external-dns-operator GA

sherine-k:
Hi @dbaker-rh this is how I think the operator's rbac will evolve into if we want to use escalate and bind. 
Please provide us feedback about these permissions.

Hi @joelanford, @kevinrizza  I'll use this PR to test the OLM installation and check the operator has bind and escalate as expected.

CC @alebedev87 @lmzuccarelli

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -81,3 +81,85 @@ rules:
   - get
   - list
   - watch
+# Operator wants to be able to ensure (create/update/delete)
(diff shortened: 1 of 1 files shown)


dbaker-rh:
These permissions should be avoided unless *absolutely* necessary.

sherine-k:
This PR was used for a POC while working on a solution for externalDNS RBAC constraints with OLM team. This is not useful anymore.

PR #119, opened 2022-02-24
credentials request test: log encoding errors properly

alebedev87:
To remove the usage of uninitialized `*testing.T` and to not log to `fmt`.


alebedev87:
`waiting for Prometheus openshift-monitoring/k8s: expected 2 replicas, got 0 updated replicas`

/retest

PR #124, opened 2022-03-24
e2e: support infoblox provider

alebedev87:
Infoblox helper added.

Differences from the other (public) providers:
- Credentials cannot be requested from CCO so they have to be provided explicitly: via environment variables
- Init is more complex because the self signed certificate of the Infoblox (currently manually deployed) has to be trusted
- Some operations can not be done with the [Infoblox golang client](https://github.com/infobloxopen/infoblox-go-client), namely the addition of the nameserver to an authoritative zone or the restart of the DNS service. The client had to be slightly enhanced to do the missing operations by calling Infoblox REST API


test/e2e/infoblox.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	authZone.Ref = ref
+
+	// NS record is not added automatically with the zone creation
+	if err = h.client.addNameServer(authZone.Ref, defaultGridMasterHostname); err != nil {
+		return "", nil, fmt.Errorf("failed to add nameserver to authoritative zone: %w", err)
+	}
+
+	// creation of an authoritative zone needs a restart of the DNS service
+	if err = h.client.restartServices(); err != nil {
```
arjunrn, line 103: Should we wait here after restarting?

test/e2e/infoblox.go
```diff
@@ -0,0 +1,296 @@
...
+	findOperatorSubscription := func() (*olmv1alpha1.Subscription, error) {
+		list := &olmv1alpha1.SubscriptionList{}
+		if err := kubeClient.List(context.TODO(), list, client.InNamespace(operatorNs)); err != nil {
+			return nil, err
+		}
+		for i := range list.Items {
+			if strings.HasPrefix(list.Items[i].Name, "external-dns-operator") {
+				return &list.Items[i], nil
+			}
+		}
```
arjunrn, line 242:
```suggestion
		for _, subscription := range list.Items {
			if strings.HasPrefix(subscription.Name, "external-dns-operator") {
				return &subscription, nil
			}
		}
```

test/e2e/infoblox.go
```diff
@@ -0,0 +1,296 @@
...
+		// best effort: return the first one
+		if len(list.Items) > 0 {
```
arjunrn, line 244: It's better to error here because if there's more than one subscription then probably then we cannot safely assume later that the configmap has been correctly injected into the operator deployment.

arjunrn (commented):
Couple of minor changes but looks good overall.

test/e2e/infoblox.go
alebedev87: I agree on this one. However I noticed that the subscription created by the prow job has a generated name which is hard to guess (something like `o-abcxyz`). The listing of the subscriptions is already filtered by `external-dns-operator` namespace and after checking the must-gather I didn't see any other subscription in this namespace. Not ideal but I didn't come up with a better idea, do you have any?

test/e2e/infoblox.go
```diff
@@ -0,0 +1,296 @@
...
+	wapiUsername string
+	wapiPassword string
+}
+
+var _ providerTestHelper = &infobloxTestHelper{}
+
+func newInfobloxHelper(kubeClient client.Client) (*infobloxTestHelper, error) {
+	helper := &infobloxTestHelper{}
+
+	if err := helper.prepareConfigurations(kubeClient); err != nil {
```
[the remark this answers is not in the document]
alebedev87: :) that one is funny. I passed the client because I knew I'd need it for the TLS configmap injection. But then I forgot to pass it down to the function which does the injection. However there is `kubeClient` global variable too which got used at the end.
Fixed this.

test/e2e/infoblox.go
arjunrn: Kind of difficult to make it work both in CI and while running from local. I guess this is good enough for now.

alebedev87:
/test all

Infoblox was moved to the long running QE account

PR #123, opened 2022-03-18
docs: Manual test of Infoblox provider

alebedev87:
Instructions similar to the other DNS providers like GCP and Azure.
Tested manually against [Infoblox NIOS DDI AWS product](https://aws.amazon.com/marketplace/pp/prodview-opxe3p2cgudwe).

PR #126, opened 2022-03-25
infoblox: TXT prefix and trusted CA support for non RHEL distributions

alebedev87:
Missing features after [the manual test of Infoblox provider](https://github.com/openshift/external-dns-operator/pull/123).

### DIFF
--- a/pkg/operator/controller/externaldns/deployment_test.go
+++ b/pkg/operator/controller/externaldns/deployment_test.go
@@ -234,6 +234,12 @@ func TestDesiredExternalDNSDeployment(t *testing.T) {
 							"--fqdn-template={{.Name}}.test.com",
 							"--txt-prefix=external-dns-",
 						},
+						Env: []corev1.EnvVar{
+							{
+								Name:  "SSL_CERT_DIR",
+								Value: "/etc/pki/ca-trust/extracted/pem",
+							},
+						},
 						VolumeMounts: []corev1.VolumeMount{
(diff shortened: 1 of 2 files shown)


pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -234,6 +234,12 @@ func TestDesiredExternalDNSDeployment(t *testing.T) {
 							"--fqdn-template={{.Name}}.test.com",
 							"--txt-prefix=external-dns-",
 						},
+						Env: []corev1.EnvVar{
+							{
+								Name:  "SSL_CERT_DIR",
+								Value: "/etc/pki/ca-trust/extracted/pem",
```
arjunrn, line 240: Is this an existing directory on the image?

alebedev87: If it's a RHEL distribution: yes. For the others: it depends but mostly no.
Btw this directory will be created if not there by the volumeMount.

alebedev87: Yeah, the `volumeMount` creation was there before, it was implemented as part the support of the proxy's trustedCA: [code](https://github.com/openshift/external-dns-operator/blob/ccc4fbbc3a2965886dce8c961d8a9038c7f5176e/pkg/operator/controller/externaldns/pod.go#L223-L229).

PR #128, opened 2022-04-04
Use operator-sdk to generate OLM bundle

alebedev87:
The main file to verify is `bundle/manifests/external-dns-operator.clusterserviceversion.yaml`. It now has all the RBAC embedded and CR examples for all the providers.

Some notes:
- volume and volume mount used for the webhook certificates are now added only for non OLM deployment. OLM injects them automatically.
- webhook service (`config/webhook/service.yaml`) was scaffolded by operator-sdk but doesn't need to be in the bundle as again OLM will generate a webhook service. However it has to be in `bundle/manifests/` during the bundle generation, otherwise `webhookdefintions` of CSV won't be generated properly. That's why it has been added to the ignore files.


sherine-k (approved):
Nice work @alebedev87  :+1: 
This looks much more concise now

PR #130, opened 2022-04-05
[WIP] Credentials for AWS provider to be consumed as a volume mount

DhritiShikhar:
** This is work in progress**


pkg/operator/controller/externaldns/pod.go
```diff
@@ -60,6 +60,9 @@ const (
 	awsAccessKeySecretEnvVar = "AWS_SECRET_ACCESS_KEY"
 	awsAccessKeyIDKey        = "aws_access_key_id"
 	awsAccessKeySecretKey    = "aws_secret_access_key"
+	awsSharedCredentialsFile = "AWS_SHARED_CREDENTIALS_FILE"
+	awsCredentialsVolumeName = "aws-credentials-file"
+	awsCredentialsFileKey    = "aws-credentials.json"
```
arjunrn, line 65: The credentials file is a not a JSON file. Any reason you gave it a JSON file suffix?

pkg/operator/controller/externaldns/pod.go
```diff
@@ -325,30 +328,33 @@ func (b *externalDNSContainerBuilder) fillAWSFields(container *corev1.Container)
 
 	env := []corev1.EnvVar{
 		{
-			Name: awsAccessKeyIDEnvVar,
+			Name: awsSharedCredentialsFile,
 			ValueFrom: &corev1.EnvVarSource{
 				SecretKeyRef: &corev1.SecretKeySelector{
```
arjunrn, line 333: The `AWS_SHARED_CREDENTIALS_FILE` environment variable should contain the path to the credentials file. Over here the variable contains the contents of the file. Not it's path.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -325,30 +328,33 @@ func (b *externalDNSContainerBuilder) fillAWSFields(container *corev1.Container)
...
 				},
 			},
 		},
 	}
 
 	container.Env = append(container.Env, env...)
+
+	for _, v := range b.volumes {
+		if v.Name == awsCredentialsVolumeName {
+			container.Env = append(container.Env, corev1.EnvVar{
```
arjunrn, line 347: The environment variable is being appended here again. This one looks correct compared to the previous one.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -544,6 +552,33 @@ func (b *externalDNSVolumeBuilder) providerSpecificVolumes() []corev1.Volume {
...
+	if len(b.secretName) == 0 {
+		return nil
+	}
+	return []corev1.Volume{
+		{
+			Name: awsCredentialsVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: b.secretName,
+					Items: []corev1.KeyToPath{
```
arjunrn, line 566: The items should not be present. The point of using the shared credentials is that the secret could contain static credentials or configuration for STS.

arjunrn (changes requested):
The implementation of the credentials mount is only partially correct. The pod could receive a credentials file with a [web identity configuration](https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#assume-role-with-web-identity) or it could receive static credentials like this

```
[profile]
aws_access_key_id=foo
aws_secret_access_key=bar
```
Either way the operator should not mount specific keys from the secret but just mount that secret in the pod and set the environment variable so that the SDK can figure it out.

pkg/operator/controller/externaldns/pod.go
DhritiShikhar: Removed the `.json` suffix

pkg/operator/controller/externaldns/pod.go
DhritiShikhar: Removed this one.

pkg/operator/controller/externaldns/pod.go
DhritiShikhar: Updated the correct env var value

pkg/operator/controller/externaldns/pod.go
DhritiShikhar: Removed

pkg/operator/controller/externaldns/pod.go
```diff
@@ -56,10 +56,11 @@ const (
 	//
 	// AWS
 	//
-	awsAccessKeyIDEnvVar     = "AWS_ACCESS_KEY_ID"
-	awsAccessKeySecretEnvVar = "AWS_SECRET_ACCESS_KEY"
 	awsAccessKeyIDKey        = "aws_access_key_id"
 	awsAccessKeySecretKey    = "aws_secret_access_key"
```
thejasn, line 60: I don't think you'll need these anymore.

thejasn (changes requested):
Couple more things, 
- Seems like you are missing the docs on how to use CCO and generate a `secrets.yaml`. 
- e2e tests also will need to be updated

pkg/operator/controller/externaldns/pod.go
```diff
@@ -544,6 +540,23 @@ func (b *externalDNSVolumeBuilder) providerSpecificVolumes() []corev1.Volume {
...
+	if len(b.secretName) == 0 {
+		return nil
+	}
+	return []corev1.Volume{
+		{
+			Name: awsCredentialsVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: b.secretName,
+				},
```
alebedev87, line 554: Only `credentials` [key](https://github.com/openshift/cloud-credential-operator/blob/e31c1696e9e9e271bc53175e3bc0b52700dd8857/pkg/operator/constants/constants.go#L91-L93) needs to be consumed, similar to Azure or GCP volumes.

alebedev87 (changes requested):
The PR still needs the support of the backward compatibility with the previous secret format: when the secret is provided by the customer (not provisioned by CCO) and doesn't have `credentials` key in it.
A function like [this one from the ingress operator](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/util/aws/shared_credentials_file.go#L26-L50) can be used in [the credentials controller](https://github.com/openshift/external-dns-operator/blob/ea9859875676474644424ecf3ec7c0ce52836577/pkg/operator/controller/credentials-secret/credentials_secret.go#L127-L129).

pkg/operator/controller/externaldns/pod.go
arjunrn: There shouldn't be a credentials key. Basically the operator should copy over the credentials(in whatever form they may be) to the operand namespaced. And the operand should just consume that secret. I really don't see where the credentials key has to be used.

arjunrn: I stand corrected. There is in fact a `credentials` keys in the secret provisioned by the CCO. But when the user provisions the credentials the secret does not have that field. So I believe what you mean is that the controller should handle both scenarios.

I would recommend that the controller which copies over the credentials from the `external-dns-operator` namespace to the `external-dns` namespace prepare the secret in the target namespace so that it is in the aws credentials file format. This way the deployment have to be generated in 2 different ways.

alebedev87: Yes, the secret provided by the user may not have the field of for the credentials file. And yes, I [agree](https://github.com/openshift/external-dns-operator/pull/130#pullrequestreview-948257920) on the idea of taking care of `credentials` key in the credentials controller.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -56,10 +56,11 @@ const (
 	//
 	// AWS
 	//
-	awsAccessKeyIDEnvVar     = "AWS_ACCESS_KEY_ID"
-	awsAccessKeySecretEnvVar = "AWS_SECRET_ACCESS_KEY"
 	awsAccessKeyIDKey        = "aws_access_key_id"
 	awsAccessKeySecretKey    = "aws_secret_access_key"
+	awsCredentialsVolumeName = "aws-credentials"
+	awsCredentialEnvVarName  = "AWS_SHARED_CREDENTIALS_FILE"
+	awsCredentialsPath       = "/aws/credentials"
```
arjunrn, line 63: Another variable called `awsCredentialsFilePath = "/aws/credentials"

pkg/operator/controller/externaldns/pod.go
```diff
@@ -325,30 +326,23 @@ func (b *externalDNSContainerBuilder) fillAWSFields(container *corev1.Container)
...
-			ValueFrom: &corev1.EnvVarSource{
-				SecretKeyRef: &corev1.SecretKeySelector{
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: b.secretName,
-					},
-					Key: awsAccessKeySecretKey,
-				},
-			},
+			Name:  awsCredentialEnvVarName,
+			Value: awsCredentialsPath,
```
arjunrn, line 330: This should be `awsCredentialsFilePath`.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -325,30 +326,23 @@ func (b *externalDNSContainerBuilder) fillAWSFields(container *corev1.Container)
...
 		},
```
arjunrn, line 331: Another environment variable called `AWS_SDK_LOAD_CONFIG` with value "1".
[Reference](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/).

alebedev87: ExternalDNS' upstream uses [the enabled shared config option](https://github.com/kubernetes-sigs/external-dns/blob/f3e04dfe8c46853966b878ec76f3eeb6b2ed1441/provider/aws/aws.go#L194) which doesn't need this variable:

> Using the NewSessionWithOptions with SharedConfigState set to SharedConfigEnable will create the session as if the AWS_SDK_LOAD_CONFIG environment variable was set.

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -124,10 +124,18 @@ func desiredCredentialsSecret(sourceSecret *corev1.Secret, destName types.Namesp
...
-		secret.Data = sourceSecret.Data
+		return secret
+	}
+
+	if extDNS.Spec.Provider.Type == operatorv1alpha1.ProviderTypeAWS {
+		secret.StringData = make(map[string]string)
+		for key, value := range sourceSecret.Data {
+			secret.StringData[key] = string(value)
+		}
+		return secret
```
alebedev87, line 135: This code is no different from the simple assignment which happens after: `secret.Data = sourceSecret.Data`.
What should happen here is the following:
- if the provider is AWS
  - If the source secret doesn't have the key `credentials`
      - take the value of the key `aws_access_key_id` from the source secret
      - take the value of the key `aws_secret_access_key` from the source secret
      - use [the function](https://github.com/openshift/cluster-ingress-operator/blob/58be189af289878b0c9b141df78814c08b83fcfa/pkg/util/aws/shared_credentials_file.go#L52-L58) to transform the just retrieved KeyID and AccessSecret into a credentials file on the fly
      - add `credentials` field into the target secret
   - Else: `secret.Data = sourceSecret.Data`
- Else: `secret.Data = sourceSecret.Data`

This would make sure that we always have `credentials` key in the target secret even if the users still used the API the old way (with the keyID and accesssecret fields).

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -138,7 +146,14 @@ func (r *reconciler) updateCredentialsSecret(ctx context.Context, current, desir
...
 	updated := current.DeepCopy()
-	updated.Data = desired.Data
+	if desired.StringData != nil {
+		updated.StringData = make(map[string]string)
+		for key, value := range desired.StringData {
+			updated.StringData[key] = string(value)
+		}
+	} else {
+		updated.Data = desired.Data
+	}
```
alebedev87, line 156: This not needed, `desired` variable already contains what's needed to be passed to `updated`.

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -150,5 +165,5 @@ func (r *reconciler) updateCredentialsSecret(ctx context.Context, current, desir
 // the secrets should be considered equal for the purpose of determining
 // whether an update is necessary, false otherwise.
 func secretsEqual(a, b *corev1.Secret) bool {
-	return reflect.DeepEqual(a.Data, b.Data)
+	return reflect.DeepEqual(a.Data, b.Data) && reflect.DeepEqual(a.StringData, b.StringData)
```
alebedev87, line 168: The ` desiredCredentialsSecret` function should return all data in `.Data` field, no need to add something in `.StringData`.

DhritiShikhar: Removed

pkg/operator/controller/credentials-secret/credentials_secret.go
DhritiShikhar: Removed

pkg/operator/controller/credentials-secret/controller_test.go
```diff
@@ -102,10 +102,27 @@ func TestReconcile(t *testing.T) {
 		},
 		{
 			name:            "Target secret didn't change",
+			existingObjects: []runtime.Object{testAzureExtDNSInstance(), testSrcSecret(), testTargetSecret()},
+			inputConfig:     testConfig(),
+			inputRequest:    testRequest(),
+			expectedResult:  reconcile.Result{},
+		},
+		{
+			name:            "Target secret changed for AWS secret credentials",
```
DhritiShikhar, line 111: TODO:

[1] Add test for secret credential changed

DhritiShikhar: Added

docs/usage.md
```diff
@@ -23,9 +23,11 @@ kind: Secret
 metadata:
   name: aws-access-key
   namespace: #operator namespace
-data:
-  aws_access_key_id: # Base-64 encoded access key id
-  aws_secret_access_key: # Base-64 encoded access key secret
+StringData:
+  credentials: |- # Provide user profile, AWS access key id and AWS secret access key
+    [default]
+    aws_access_key_id = ""
```
darshan-nagaraj, line 29:
```suggestion
    aws_access_key_id = " <AWS_ACCESS_KEY_ID>"
```

docs/usage.md
```diff
@@ -23,9 +23,11 @@ kind: Secret
...
+    aws_secret_access_key = ""
```
darshan-nagaraj, line 30:
```suggestion
    aws_secret_access_key = "<AWS_SECRET_ACCESS_KEY>"
```

api/v1alpha1/externaldns_types.go
```diff
@@ -244,11 +244,18 @@ type ExternalDNSProvider struct {
...
 	// * aws_access_key_id
 	// * aws_secret_access_key
 	//
+	// Example:
+	// 		StringData:
+	//   		credentials: |-
+	// 			[developer]
+	// 			aws_access_key_id = "abc"
+	// 			aws_secret_access_key = "xyz"
+	//
```
alebedev87, line 258:
```suggestion
	// the shared credentials file (https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html)
	// under `credentials` key or following keys (with corresponding values):
	//
	// * aws_access_key_id
	// * aws_secret_access_key
	// 
```

docs/usage.md
```diff
@@ -23,9 +23,11 @@ kind: Secret
 metadata:
   name: aws-access-key
   namespace: #operator namespace
-data:
-  aws_access_key_id: # Base-64 encoded access key id
-  aws_secret_access_key: # Base-64 encoded access key secret
+StringData:
```
alebedev87, line 26:
```suggestion
stringData:
```

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -124,11 +128,27 @@ func desiredCredentialsSecret(sourceSecret *corev1.Secret, destName types.Namesp
 		case operatorv1alpha1.ProviderTypeAWS:
 			secret.Data = sourceSecret.Data
 		}
-	} else {
-		secret.Data = sourceSecret.Data
+		return secret, nil
 	}
 
-	return secret
+	if extDNS.Spec.Provider.Type == operatorv1alpha1.ProviderTypeAWS {
```
alebedev87, line 134: Can you please add a small comment which explains what this code is for?

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -124,11 +128,27 @@ func desiredCredentialsSecret(sourceSecret *corev1.Secret, destName types.Namesp
...
+		secret.Data = make(map[string][]byte)
```
alebedev87, line 135: Not needed, `Data` field is already initialized at the beginning of the function.

pkg/operator/controller/credentials-secret/credentials_secret.go
```diff
@@ -124,11 +128,27 @@ func desiredCredentialsSecret(sourceSecret *corev1.Secret, destName types.Namesp
...
+		switch {
+		case len(sourceSecret.Data["credentials"]) > 0:
+			secret.Data["credentials"] = sourceSecret.Data["credentials"]
+		case len(sourceSecret.Data["aws_access_key_id"]) > 0 && len(sourceSecret.Data["aws_secret_access_key"]) > 0:
+			secret.Data["credentials"] = newConfigForStaticCreds(
+				string(sourceSecret.Data["aws_access_key_id"]),
+				string(sourceSecret.Data["aws_secret_access_key"]),
+			)
+		default:
+			return nil, fmt.Errorf("invalid secret for aws credentials")
```
alebedev87, line 145: Since we check the contents of the secret now, it makes sense to do it for other providers too.
This can be done in a followup PR.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -56,10 +56,11 @@ const (
...
 	//
-	awsAccessKeyIDEnvVar     = "AWS_ACCESS_KEY_ID"
-	awsAccessKeySecretEnvVar = "AWS_SECRET_ACCESS_KEY"
-	awsAccessKeyIDKey        = "aws_access_key_id"
-	awsAccessKeySecretKey    = "aws_secret_access_key"
+	awsCredentialEnvVarName  = "AWS_SHARED_CREDENTIALS_FILE"
+	awsCredentialsVolumeName = "aws-credentials"
+	awsCredentialsDir        = "/aws"
+	awsCredentialsFilePath   = "/aws/credentials"
+	awsSecretKey             = "credentials"
```
alebedev87, line 63:
```suggestion
	awsCredentialsMountPath        = defaultConfigMountPath
	awsCredentialsFileKey             = "credentials"
	awsCredentialsFileName = "aws-credentials"
	awsCredentialsFilePath = filepath.Join(awsCredentialsMountPath, awsCredentialsFileName)
```
To match what's done for other providers.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -544,6 +532,29 @@ func (b *externalDNSVolumeBuilder) providerSpecificVolumes() []corev1.Volume {
...
+	return []corev1.Volume{
+		{
+			Name: awsCredentialsVolumeName,
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: b.secretName,
+					Items: []corev1.KeyToPath{
+						{
+							Key:  awsSecretKey,
+							Path: awsSecretKey,
```
alebedev87, line 549:
```suggestion
							Key:  awsCredentialsFileKey,
							Path: awsCredentialsFileName,
```

pkg/operator/controller/externaldns/pod.go
DhritiShikhar: @alebedev87 `throws error `filepath.Join(awsCredentialsMountPath, awsCredentialsFileName) (value of type string) is not constant`

Can I just put `	awsCredentialsFilePath   = awsCredentialsMountPath + "/" + awsCredentialsFileName`
`

pkg/operator/controller/credentials-secret/controller_test.go
```diff
@@ -102,10 +102,44 @@ func TestReconcile(t *testing.T) {
 		},
 		{
 			name:            "Target secret didn't change",
+			existingObjects: []runtime.Object{testAzureExtDNSInstance(), testSrcSecret(), testTargetSecret()},
+			inputConfig:     testConfig(),
+			inputRequest:    testRequest(),
+			expectedResult:  reconcile.Result{},
```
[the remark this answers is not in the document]
alebedev87: Because it's Azure ExtDns CR. You should use `AWS` test function.

pkg/operator/controller/credentials-secret/controller_test.go
```diff
@@ -534,8 +584,8 @@ func testTargetSecret() *corev1.Secret {
 			Namespace: testOperandNamespace,
 		},
 		Data: map[string][]byte{
-			"key1": []byte("val1"),
-			"key2": []byte("val2"),
+			"aws_access_key_id":     []byte("val1"),
+			"aws_secret_access_key": []byte("val2"),
```
alebedev87, line 588: The target secret should have `credentials` key.

DhritiShikhar: Updated

alebedev87 (approved):

alebedev87:
/label docs-approved

Based on https://github.com/openshift/external-dns-operator/pull/130#issuecomment-1113313154

PR #132, opened 2022-04-11
rbac: operand rbac out of the extras, add end user cluster roles

alebedev87:
- Reduce the number of the extra resources by moving the operand cluster role/binding in the bundle
- Add end user cluster roles to view and edit `ExternalDNS` CR


sherine-k (approved):
Nice step towards getting rid of extra-roles.yaml, hopefully soon!
kudos @alebedev87

PR #133, opened 2022-04-11
doc: release-branch mapping

alebedev87:
As per the feedback from https://issues.redhat.com/browse/CFE-322

### DIFF
--- a/docs/openshift.md
+++ b/docs/openshift.md
@@ -10,3 +10,8 @@ Use the following convenience script to secure communication between the API and
 ```bash
 $ ./hack/add-serving-cert.sh --namespace external-dns-operator --service webhook-service --webhook validating-webhook-configuration --secret webhook-server-cert
 ```
+
+## Release-branch mapping
+| OCP version | ExternalDNS Operator branch |
+| :---------: | :-------------------------: |
+| 4.10        | release-0.1                 |


docs/openshift.md
```diff
@@ -10,3 +10,8 @@ Use the following convenience script to secure communication between the API and
 ```bash
 $ ./hack/add-serving-cert.sh --namespace external-dns-operator --service webhook-service --webhook validating-webhook-configuration --secret webhook-server-cert
 ```
+
+## Compatibility matrix
+| ExternalDNS Operator version | OCP version |
+| ---------------------------- | ----------- |
+| 0.1.x                        | 4.10.x      |
```
Miciah, line 17:
```suggestion
| 0.1.z                        | 4.10.z      |
```
Where would I look to determine the ExternalDNS Operator version?  I don't see any releases or tags, but I do see a "release-0.1" branch, and I'm guessing OperatorHub shows the version, right?  Is this table is telling me which release branch to build from (or which version of OpenShift to use with a build from a given release branch) and what version I can select in OperatorHub?

alebedev87: Good questions. I agree it's not clear what this matrix is supposed to show. What I think I wanted to say is how to map between OCP releases and the operator branches.

Miciah:
Thanks!
/lgtm

PR #140, opened 2022-04-25
e2e: create ingress controller in openshift-ingress-operator namespace

alebedev87:
Cluster ingress operator doesn't reconcile from any namespace anymore: [PR](https://github.com/openshift/cluster-ingress-operator/pull/740).

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -462,9 +462,11 @@ func TestExternalDNSCustomIngress(t *testing.T) {
 	}
 
 	openshiftRouterName := "external-dns"
-	name := types.NamespacedName{Namespace: testIngressNamespace, Name: openshiftRouterName}
+	// ingress controllers are supposed to be created in the ingress operator namespace
+	name := types.NamespacedName{Namespace: "openshift-ingress-operator", Name: openshiftRouterName}
+	ingDomain := fmt.Sprintf("%s.%s", name.Name, hostedZoneDomain)
 	t.Log("Create custom ingress controller")
-	ing := newHostNetworkController(name, name.Name+"."+hostedZoneDomain)
+	ing := newHostNetworkController(name, ingDomain)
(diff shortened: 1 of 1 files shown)


test/e2e/operator_test.go
```diff
@@ -462,9 +462,11 @@ func TestExternalDNSCustomIngress(t *testing.T) {
 	}
 
 	openshiftRouterName := "external-dns"
-	name := types.NamespacedName{Namespace: testIngressNamespace, Name: openshiftRouterName}
+	// ingress controllers are supposed to be created in the ingress operator namespace
+	name := types.NamespacedName{Namespace: "openshift-ingress-operator", Name: openshiftRouterName}
+	ingDomain := name.Name + "." + hostedZoneDomain
```
arjunrn, line 467: Can you change this to a string format instead.

test/e2e/operator_test.go
```diff
@@ -566,3 +572,13 @@ func fetchRouterCanonicalHostname(t *testing.T, routeName types.NamespacedName)
 	}
 	return canonicalName, nil
 }
+
+func ingressConditionStatus(ing routev1.RouteIngress, t routev1.RouteIngressConditionType) corev1.ConditionStatus {
```
arjunrn, line 576: I would change this to `ingressConditionHasStatus(ing routev1.RouteIngress, condition routev1.RouteIngressConditionType, status corev1.ConditionStatus) bool` because right now we are return `corev1.ConditionUnknown` if we cannot find the condition which is strictly not true.

arjunrn (commented):
Just 2 small changes.

test/e2e/operator_test.go
alebedev87: Don't want to start a holy war but generally I prefer the standard concatenation over any formatting function for the small and simple patterns like here. Because it's reported as being faster, by quite some sources (at least it used to be by the time I took this habit).
However in this file I see the total usage of `Sprintf`, so changing it just to be coherent.

PR #141, opened 2022-04-26
Add v1beta1 version

alebedev87:
What's inside:
- New version added in `api/v1beta1/`
- No conversions added as `v1beta1` is identical to `v1alpha1`
- `v1beta1` is the storage version now
- Webhook is moved to `api/v1beta1/`, it handles only `v1beta1` (`v1alpha1` objects are also served by API, see the notes)
- Code of the controllers is migrated to `v1beta1`
- Manifests updated: 
   - CRD got a new new version
   - Webhook updated with the new version
   - `v1beta1` samples added
- e2e tests migrated to `v1beta1`, one test is added for the check of `v1alpha1` resources

Notes:
- The controller and the validating webhook are setup to work with `v1beta1` only. This works out even for the objects created before the upgrade to this version of the operator. Because the version which is being watched/used is the [preferred](https://book.kubebuilder.io/multiversion-tutorial/deployment.html#kubectl-and-preferred-versions)(has to do with [the version priority](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-priority) too), so API will take care of making the conversion. This is different from where the objects are stored though. The objects created before the new version was added to CRD are still stored as the previous version until they are touched (patch/update).
- Migration to `stable` channel is done in a followup PR: https://github.com/openshift/external-dns-operator/pull/146. The reason is to not block other PR as this change needs the CI update (bundle installation).


Makefile
```diff
@@ -44,7 +44,7 @@ SETUP_ENVTEST := go run sigs.k8s.io/controller-runtime/tools/setup-envtest
 K8S_ENVTEST_VERSION := 1.21.4
 
 PACKAGE=github.com/openshift/external-dns-operator
-MAIN_PACKAGE=$(PACKAGE)/cmd/external-dns-operator
+MAIN_PACKAGE=$(PACKAGE)
```
DhritiShikhar, line 47: `$(PACKAGE)` and `$(MAIN_PACKAGE)` has same value. Is there a reason to keep both of them?

api/v1beta1/externaldns_types.go
```diff
@@ -0,0 +1,603 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Kind string `json:"kind"`
+
+	// Version is the API version
+	// of the given resource kind for
+	// ExternalDNS to use.
+	//
+	// e.g. "externaldns.k8s.io/v1alpha1"
```
DhritiShikhar, line 553:
```suggestion
	// e.g. "externaldns.k8s.io/v1beta1"
```

config/samples/infoblox/operator_v1beta1_infoblox_openshift.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: externaldns.olm.openshift.io/v1beta1
```
DhritiShikhar, line 1: I see samples added for AWS, GCP, Azure and Infoblox but not Bluecat.. Is there a reason for that or just missed?

api/v1beta1/externaldns_types.go
alebedev87: It is the version of the [upstream CRD](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/contributing/crd-source/crd-manifest.yaml#L20), not our ExternalDNS API.

Makefile
alebedev87: Not anymore, removed `MAIN_PACKAGE`

config/samples/infoblox/operator_v1beta1_infoblox_openshift.yaml
alebedev87: Added BlueCat samples too

Makefile
Miciah: What was the reason for moving the main executable out of `cmd/`?  Is this change directly related somehow to the v1beta1 migration?

Miciah (commented):
Looks good overall.  However, I'm wondering about the move of the main executable out of `cmd/`; is this change directly related somehow to the v1beta1 migration?

bundle/manifests/external-dns-operator.clusterserviceversion.yaml
```diff
@@ -60,6 +60,36 @@ metadata:
...
+                }
+              },
+              "type": "BlueCat"
+            },
+            "source": {
+              "openshiftRouteOptions": {
+                "routerName": "default"
+              },
+              "type": "OpenShiftRoute"
+            }
```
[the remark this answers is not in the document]
alebedev87: Some examples have `zones` field, some `domains` I thought of giving multiple different examples.
BlueCat example uses `domains` field to filter the zones by the domain name.

PROJECT
```diff
@@ -15,4 +15,12 @@ resources:
...
     validation: true
     webhookVersion: v1
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: openshift.io
+  group: externaldns.olm
+  kind: ExternalDNS
+  path: github.com/openshift/external-dns-operator/api/v1beta1
+  version: v1beta1
```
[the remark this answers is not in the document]
alebedev87: Good question.
Looking at the [kubebuilder docs](https://book.kubebuilder.io/reference/project-config.html): `controller` and `webhooks` fields are added at the scaffolding time. Technically both the controller and the webhook were scaffolded for the version `v1alpha1`, so it reflects well the history behind this operator. Also, I didn't see any impact on the runtime.

alebedev87:
@Miciah : no, the move of `main.go` out of `cmd/` is not related to the migration to `v1beta1`. It's the new kubebuilder project layout: before it was scaffolding into `cmd/`, now it's putting `main.go` in the root of the repository. I updated `operator-sdk` version meanwhile and the newer version could not scaffold the new api if `main.go` was not in the root.

PR #146, opened 2022-04-29
v1beta1 api: use stable channel for OLM bundle

alebedev87:
`stable` channel is added as per the [OLM naming convention](https://olm.operatorframework.io/docs/best-practices/channel-naming/#naming). The convention may need some coding on the CPaaS side:
  - `X.Y` version can be appended on the CPaaS side where `X` and `Y` are known
  - `skipRange` can be added to CSV in CPaaS too

Follow-up of https://github.com/openshift/external-dns-operator/pull/141


DhritiShikhar:
@alebedev87 

Just a thought, should the value of `spec.maturity` here -> https://github.com/openshift/external-dns-operator/blob/main/bundle/manifests/external-dns-operator.clusterserviceversion.yaml#L355 

also be `stable` instead of `alpha`?

Refer -> https://docs.openshift.com/container-platform/4.6/operators/operator_sdk/osdk-generating-csvs.html

alebedev87:
/retest

`error creating EC2 NAT Gateway: InvalidElasticIpID.NotFound: The elasticIp ID 'eipalloc-0c269657ac123a6d6' does not exist`

PR #153, opened 2022-05-10
validating webhook: CRD source is not implemented

alebedev87:
CRD source is not implemented for the moment.

### DIFF
--- a/api/v1beta1/externaldns_webhook.go
+++ b/api/v1beta1/externaldns_webhook.go
@@ -48,13 +48,13 @@ var _ webhook.Validator = &ExternalDNS{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *ExternalDNS) ValidateCreate() error {
 	webhookLog.Info("validate create", "name", r.Name)
-	return r.validate()
+	return r.validate(nil)
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
-func (r *ExternalDNS) ValidateUpdate(_ runtime.Object) error {
+func (r *ExternalDNS) ValidateUpdate(old runtime.Object) error {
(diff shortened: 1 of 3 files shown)


Miciah:
What if someone creates an ExternalDNS CR with `spec.source.type: CRD` with an old version of the operator and then upgrades to a version with this validation?  Would the webhook then reject attempts to update the CR (unless the updates also changed `spec.source.type`)?  We could add ratcheting logic to `validateSources` to ignore an invalid value if it is already set to avoid problems in this scenario.

alebedev87:
Thanks for the review @Miciah!

`CRD` source wasn't never implemented, if a CR with `CRD` was ever created, its only result was errors in the operator without any instance of externaldns created. CRs will be kept anyway but don't we want to let the user know that he's doing something wrong especially if he keeps updating this CR?

Miciah:
> `CRD` source wasn't never implemented, if a CR with `CRD` was ever created, its only result was errors in the operator without any instance of externaldns created. 

The CRD would still be present though, with the invalid value, right?

> CRs will be kept anyway but don't we want to let the user know that he's doing something wrong especially if he keeps updating this CR?

We do.  The goal with ratcheting logic is that we reject attempts to set an invalid value, but even if an invalid value is already set, we allow updates to _other_ fields.  If we always reject updates where the field has an invalid value, then we block updates where the user or the operator tries to set some other field.  This is especially problematic if the operator is blocked from updating status because of an invalid value in spec.

alebedev87:
@Miciah : didn't implement a webhook test case when ExternalDNS CR with CRD source is updated as it was too much effort (if even possible). The test environment (`envtest`) is created with the webhook manifest which disallows the creation of CRs with CRD source. Had to test the update manually.

Miciah:
Thanks!
/lgtm

PR #154, opened 2022-05-10
Use only stable channel + flow diagram update

alebedev87:
Follow-up of https://github.com/openshift/external-dns-operator/pull/146.  This PR aims at finishing the migration to `v1beta1` API version but letting only the stable channel include it.
Plus a small bonus of the flow diagram which was out dated since the migration to the single namespace.


Miciah:
The channel changes look fine.  I have a some comments about the updated diagram in `docs/images/external-dns-flow-openshift.png`:

* The CRs in the green boxes are truncated at the bottom.
* The meaning of the diamond shapes is unclear.  You have two CRs named "one" and "two", and then two diamonds labeled "external-dns-one" and "external-dns-two"; do these represent deployments, or just abstract external-dns instances?  
* What's up with the AWS logo in the box for the hosted zone and the icons for Azure and GCP off to the side?  We want to indicate that the hosted zone could be configured using any one of these providers, right?  Could we put the provider icons together (in the hosted-zone box, on the edges between external-dns and the hosted zone, or just off to the side) to indicate that they represent alternatives and that no one of them is special?

alebedev87:
Thanks for the review @Miciah !

- CR truncation is fixed
- Yes, it was from draw.io Kubernetes specific charts and yes it meant a deployment. I put `Deployment` word to make this clear.
- Yeah, the idea was to show different options for the DNS provider. I put they side by side with Infoblox logo as well. Let me know if you have a better idea.

alebedev87:
Quota problem on GCP.
`compute.googleapis.com/n2_cpus is not available in us-central1 because the required number of resources (12) is more than remaining quota of 4`
/retest

Miciah:
Thanks!
/lgtm

alebedev87:
/hold
Stop the robot, GCP quota issue is ongoing

PR #156, opened 2022-05-13
Enable prometheus metrics scraping

alebedev87:
Result:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,10 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	cd config/manager && $(KUSTOMIZE) edit set image quay.io/openshift/origin-external-dns-operator=${IMG}
 	# webhook volume and service are added explicilty so that they don't land in the bundle where it's managed by OLM
 	cd config/default && $(KUSTOMIZE) edit add patch --path=manager_webhook_volume_patch.yaml
+	# disable tls config in service monitor
+	cd config/prometheus && $(KUSTOMIZE) edit add patch --path=insecure_tls_patch.yaml
+	# consume certificate from the service serving certificate
+	cd config/default && $(KUSTOMIZE) edit add patch --path=manager_insecure_tls_auth_proxy_patch.yaml
 	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ The `ExternalDNS` Operator allows you to deploy and manage [ExternalDNS](https:/
     - [Installing the ExternalDNS Operator using a custom index image on OperatorHub](#installing-the-externaldns-operator-using-a-custom-index-image-on-operatorhub)
 - [Running end-to-end tests manually](#running-end-to-end-tests-manually)
 - [Proxy support](#proxy-support)
+- [Metrics](#metrics)
 
 ## Deploying the `ExternalDNS` Operator
 The following procedure describes how to deploy the `ExternalDNS` Operator for AWS.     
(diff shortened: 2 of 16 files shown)


DhritiShikhar:
Should there be any tests for this @alebedev87 ?

Makefile
```diff
@@ -156,6 +156,10 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 	cd config/manager && $(KUSTOMIZE) edit set image quay.io/openshift/origin-external-dns-operator=${IMG}
 	# webhook volume and service are added explicilty so that they don't land in the bundle where it's managed by OLM
 	cd config/default && $(KUSTOMIZE) edit add patch --path=manager_webhook_volume_patch.yaml
+	# add service monitor with certificate from the service serving certificate
+	cd config/prometheus && $(KUSTOMIZE) edit add resource monitor.yaml
```
arjunrn, line 160: Why not add the monitor resource directly or put another way include it in the default target.

README.md
```diff
@@ -194,3 +195,20 @@ Prepare your environment for the installation commands.
 ## Proxy support
 
 [Configuring proxy support for ExternalDNS Operator](./docs/proxy.md)
+
+## Metrics
+
+The ExternalDNS Operator exposes [controller-runtime metrics](https://book-v1.book.kubebuilder.io/beyond_basics/controller_metrics.html) in the format expected by [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator).
```
arjunrn, line 201: Maybe link to the latest version of the [kubebuilder book](https://book.kubebuilder.io/reference/metrics.html)

README.md
```diff
@@ -194,3 +195,20 @@ Prepare your environment for the installation commands.
...
+`ServiceMonitor` object is created in the operator's namespace (by default `external-dns-operator`), make sure that your instance of the Prometheus Operator is properly configured to find it.     
+You can check `.spec.serviceMonitorNamespaceSelector` and `.spec.serviceMonitorSelector` fields of `prometheus` resource and edit the operator's namespace or service monitor accordingly:
```
arjunrn, line 203: If the namespace can be add by adding a label then shouldn't we be adding the label to the namespace resource? Or is it not possible when the namespace is created by OLM.

config/default/kustomization.yaml
```diff
@@ -6,16 +6,11 @@ resources:
 - ../rbac
 - ../manager
 - ../webhook
+- ../prometheus
```
arjunrn, line 9: So since the _prometheus_ directory is already included do you need to `add` it with kustomize as well?

arjunrn (commented):
Just a couple of questions. Otherwise looks good.

Makefile
alebedev87: Done, added `monitor.yaml` as default resource, `tlsConfig` patch is added for the bundle generation only.

config/default/kustomization.yaml
alebedev87: Not sure I understood this one. It's already in kustomize. You can deploy using `make deploy` or with a bundle. Both way will create `ServieMonitor` and all the necessary things.

alebedev87:
@DhritiShikhar : for the e2e test we'd need a viable way of verifying the target creation in prometheus. Will try to find something.

alebedev87:
As discussed with Dhriti, the e2e testing is possible but quite heavy. Prometheus targets are not exposed on the Kubernetes level (CRs or ConfigMaps/Secrets), they have to be retrieved from the prometheus API which most often is queried by `exec` inf into the prometheus POD.

README.md
arjunrn: AFAIK the cluster prometheus should not be modified to monitor non cluster operator. Instead another prometheus instance should be used. The instructions are [here](https://docs.openshift.com/container-platform/4.10/monitoring/enabling-monitoring-for-user-defined-projects.html).

config/default/kustomization.yaml
arjunrn: My point is was that if the `ServiceMonitor` was already part of the kustomize do you need to `add` it again for the `make deploy` target.

PR #157, opened 2022-05-17
Update gcp-openshift.md

pbertera:
updated `apiVersion` to `v1alpha1


alebedev87:
@pbertera : ExternalDNS API is now `v1beta1`, so all the examples should be using it.

PR #159, opened 2022-05-24
Add postponed requeuing and credentials exist status condition

alebedev87:
- `RequeueAfter` the configurable period if the credentials secret or the trusted CA configmap not found. This helps to greatly reduce the number of repeating error logs.
- `CredentialsSecretExists` status condition is added to signal a missing secret.
- Bonus: bug in the platform detection - error if the check for the `operator.openshift.io` API group fails

### DIFF
--- a/main.go
+++ b/main.go
@@ -47,6 +47,7 @@ func main() {
 	flag.BoolVar(&opCfg.EnableWebhook, "enable-webhook", operatorconfig.DefaultEnableWebhook, "Enable the validating webhook server. Defaults to true.")
 	flag.BoolVar(&opCfg.EnablePlatformDetection, "enable-platform-detection", operatorconfig.DefaultEnablePlatformDetection, "Enable the detection of the underlying platform. Defaults to true.")
 	flag.StringVar(&opCfg.HealthProbeBindAddress, "health-probe-bind-addr", operatorconfig.DefaultHealthProbeAddr, "The address the health endpoint binds to.")
+	flag.IntVar(&opCfg.RequeuePeriodSeconds, "requeue-period", operatorconfig.DefaultRequeuePeriodSeconds, "Requeue period for a failed reconciliation (in seconds).")
 	opts := zap.Options{
 		Development: true,
 	}
--- a/pkg/operator/config/config.go
+++ b/pkg/operator/config/config.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/discovery"
@@ -42,6 +43,7 @@ const (
 	DefaultEnablePlatformDetection = true
 	DefaultTrustedCAConfigMapName  = ""
 	DefaultHealthProbeAddr         = ":9440"
+	DefaultRequeuePeriodSeconds    = 5
 
 	openshiftKind              = "OpenShiftAPIServer"
 	openshiftResourceGroup     = "operator.openshift.io"
@@ -88,6 +90,9 @@ type Config struct {
 	// HealthProbeBindAddress is the TCP address that the operator should bind to for
 	// serving health probes (readiness and liveness).
 	HealthProbeBindAddress string
+
+	// RequeuePeriodSeconds is the number of seconds to wait after a failed reconciliation.
+	RequeuePeriodSeconds int
 }
 
 // DetectPlatform detects the underlying platform and fills corresponding config fields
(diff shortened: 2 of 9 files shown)


pkg/operator/controller/externaldns/controller.go
```diff
@@ -172,10 +176,30 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+						// show that the secret is not there yet
+						if err := r.updateExternalDNSStatus(ctx, externalDNS, currentDeployment, false); err != nil {
+							reqLogger.Error(err, "failed to update externalDNS custom resource")
+						}
+					case "configmaps":
+						res.RequeueAfter = r.config.RequeuePeriod
+					}
+				}
+			}
+		}
```
thejasn, line 198: Not against it, but there seems to be 3 sources to throw `NotFound` error (`getSecret()`, `getConfigMap()` and `getDeployment()`), we could have just relied on that if `ensureExternalDNSDeployment()` only ensured deployment.

Looks like `ensureExternalDNSDeployment()` tries to ensure its prerequisites as well, maybe these should be outside at the controller itself.

main.go
```diff
@@ -47,6 +47,7 @@ func main() {
 	flag.BoolVar(&opCfg.EnableWebhook, "enable-webhook", operatorconfig.DefaultEnableWebhook, "Enable the validating webhook server. Defaults to true.")
 	flag.BoolVar(&opCfg.EnablePlatformDetection, "enable-platform-detection", operatorconfig.DefaultEnablePlatformDetection, "Enable the detection of the underlying platform. Defaults to true.")
 	flag.StringVar(&opCfg.HealthProbeBindAddress, "health-probe-bind-addr", operatorconfig.DefaultHealthProbeAddr, "The address the health endpoint binds to.")
+	flag.IntVar(&opCfg.RequeuePeriodSeconds, "requeue-period", operatorconfig.DefaultRequeuePeriodSeconds, "Requeue period for a failed reconciliation (in seconds).")
```
thejasn, line 50: Potentially this could be different for different resources/controllers, adding it here may not be future-proof yet.

pkg/operator/controller/externaldns/controller.go
alebedev87: Yeah, that was my thought too, but trying to make a smallest change possible got me :)
Let me try to check the pre-requisites before ensuring the deployment.

pkg/operator/config/config.go
```diff
@@ -122,20 +130,25 @@ func (c *Config) InjectTrustedCA() bool {
...
 // isOCP returns true if the platform is OCP
-func isOCP(kubeClient discovery.DiscoveryInterface) bool {
+func isOCP(kubeClient discovery.DiscoveryInterface) (bool, error) {
 	// Since, CRD for OpenShift API Server was introduced in OCP v4.x we can verify if the current cluster is on OCP v4.x by
 	// ensuring that resource exists against Group(operator.openshift.io), Version(v1) and Kind(OpenShiftAPIServer)
 	// In case it doesn't exist we assume that external dns is running on non OCP 4.x environment
 	resources, err := kubeClient.ServerResourcesForGroupVersion(openshiftResourceGroup + "/" + openshiftResourceVersion)
 	if err != nil {
-		return false
+		return false, err
```
alebedev87, line 145: @thejasn : I saw the wrong OpenShift detection on my local cluster too (remember the case in the CI). This should help to propagate the error if there was one.

main.go
alebedev87: Yes, it's not 100% future proof I agree. However the main controller is the one which reconciles the ExternalDNS resource, the others may even be removed in the future as they are not that useful now, when the operands are deployed in the operator's namespace.

pkg/operator/config/config.go
thejasn: Looks like it didn't fail on the CI (maybe rerun, not sure).

alebedev87: Yeah, now if there is a problem during the platform detection the container will exit with non-zero status which will make it a subject to the restart. Since OLM makes sure the deployment of the operator has the minimum availability - the installation won't succeed until the platform is detected. The error I got on my local cluster was:
```
2022-05-30T14:44:17.293Z	ERROR	setup	failed to detect the platform	{"error": "Get \"https://10.217.4.1:443/apis/operator.openshift.io/v1?timeout=32s\": dial tcp 10.217.4.1:443: i/o timeout"}

$ oc get svc -o wide -A | grep 10.217.4.1
default                                            kubernetes                                 ClusterIP      10.217.4.1     <none> 
```
So, I suppose that this could be the problem for the CI clusters too, some temporary API unavailability which used to be interpreted as `Not OpenShift` case.

alebedev87:
Just a squash of the commits

alebedev87:
/retest

Infoblox temporary licences expired, had to recreate the test instance.

alebedev87:
/retest

Wrong grid hostname was setup on the test Infoblox.

alebedev87:
/retest

Self signed certificate didn't have infoblox DNS in SAN.

PR #162, opened 2022-06-02
Bump to k8s.io v1.24.0

alebedev87:
This updates the Kube dependencies to the newly released 1.24.0.


alebedev87:
/retest

Error:
`Failed to create external DNS "test-extdns": admission webhook "vexternaldns.kb.io" denied the request: credentials secret must be specified when provider type is AWS`
It's fixed in https://github.com/openshift/external-dns-operator/pull/159

PR #163, opened 2022-06-03
DO NOT MERGE: test downstream operand v0.12.0

alebedev87:
Plus carry patch of the fix: https://github.com/openshift/external-dns/pull/43


alebedev87:
/retest-required

AWS cluster deprovision failed.

PR #165, opened 2022-06-15
Bug 2097351: Enable the rolling update for the operator deployment

alebedev87:


bundle/manifests/external-dns-operator.clusterserviceversion.yaml
```diff
@@ -419,8 +419,7 @@ spec:
           selector:
             matchLabels:
               name: external-dns-operator
-          strategy:
-            type: Recreate
+          strategy: {}
```
thejasn, line 422: Since it is going to default to rolling update, I think it is best to provide some explicit values that we know for sure won't cause issues.

quarterpin:
/label qe-approved
Verified via pre-merge workflow. More details can be found in https://bugzilla.redhat.com/show_bug.cgi?id=2097351#c1

PR #171, opened 2022-08-30
OCPBUGS-819: Azure provider - add wildcard replacement for TXT records

alebedev87:
Since the addition of the new format of TXT records in external-dns `0.12.0` the TXT records are prepended with the type of the DNS record which doesn't comply with the standards of Azure DNS in case of the wildcards: the wildcard symbol (`*`) has to be the leftmost one.

Related issue: https://github.com/kubernetes-sigs/external-dns/issues/2922

### DIFF
--- a/pkg/operator/controller/externaldns/pod.go
+++ b/pkg/operator/controller/externaldns/pod.go
@@ -35,19 +35,20 @@ import (
 )
 
 const (
-	defaultMetricsAddress    = "127.0.0.1"
-	defaultOwnerPrefix       = "external-dns"
-	defaultMetricsStartPort  = 7979
-	defaultConfigMountPath   = "/etc/kubernetes"
-	defaultTXTRecordPrefix   = "external-dns-"
-	providerArg              = "--provider="
-	httpProxyEnvVar          = "HTTP_PROXY"
-	httpsProxyEnvVar         = "HTTPS_PROXY"
-	noProxyEnvVar            = "NO_PROXY"
-	trustedCAVolumeName      = "trusted-ca"
-	trustedCAFileName        = "tls-ca-bundle.pem"
-	trustedCAFileKey         = "ca-bundle.crt"
-	trustedCAExtractedPEMDir = "/etc/pki/ca-trust/extracted/pem"
+	defaultMetricsAddress         = "127.0.0.1"
+	defaultOwnerPrefix            = "external-dns"
+	defaultMetricsStartPort       = 7979
+	defaultConfigMountPath        = "/etc/kubernetes"
+	defaultTXTRecordPrefix        = "external-dns-"
+	defaultTXTWildcardReplacement = "any"
+	providerArg                   = "--provider="
+	httpProxyEnvVar               = "HTTP_PROXY"
+	httpsProxyEnvVar              = "HTTPS_PROXY"
+	noProxyEnvVar                 = "NO_PROXY"
+	trustedCAVolumeName           = "trusted-ca"
+	trustedCAFileName             = "tls-ca-bundle.pem"
+	trustedCAFileKey              = "ca-bundle.crt"
+	trustedCAExtractedPEMDir      = "/etc/pki/ca-trust/extracted/pem"
 	// RHEL path for the trusted certificate bundle may not work for some distributions (e.g. Alpine).
 	// This makes the usage of the trusted CAs impossible when the vanilla upstream image is given.
 	// SSL_CERT_DIR allows Golang's crypto library to override the default locations.
@@ -362,6 +363,9 @@ func (b *externalDNSContainerBuilder) fillAWSFields(container *corev1.Container)
 func (b *externalDNSContainerBuilder) fillAzureFields(zone string, container *corev1.Container) {
 	// https://github.com/kubernetes-sigs/external-dns/issues/2082
 	container.Args = addTXTPrefixFlag(container.Args)
+
+	// https://github.com/kubernetes-sigs/external-dns/issues/2922
+	container.Args = append(container.Args, fmt.Sprintf("--txt-wildcard-replacement=%s", defaultTXTWildcardReplacement))
 
 	// check the zone field for the keyword 'privatednszones', this ensures that the
 	// provider 'azure-private-dns' is passed to the container
--- a/pkg/operator/controller/externaldns/deployment_test.go
+++ b/pkg/operator/controller/externaldns/deployment_test.go
@@ -449,6 +449,7 @@ func TestDesiredExternalDNSDeployment(t *testing.T) {
 							"--ignore-hostname-annotation",
 							"--fqdn-template={{.Name}}.test.com",
 							"--txt-prefix=external-dns-",
+							"--txt-wildcard-replacement=any",
 							"--azure-config-file=/etc/kubernetes/azure.json",
 						},
 						VolumeMounts: []corev1.VolumeMount{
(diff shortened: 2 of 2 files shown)


alebedev87:
Infoblox temporary licences had to be renewed

/retest

candita:
@alebedev87 please add a JIRA so this can be tracked and tested.

Miciah:
What happens when a route or service with the host name "\*.foo" exists and another route or service with the host name "ws.foo" is created?  What happens in the reverse case: something with "ws.foo" exists and something with "\*.foo" is created?  

CoreDNS and SkyDNS treat "any" the same as "\*", so we could use that instead of "ws", but "any" isn't standard.  It doesn't seem like there is any value we can use in lieu of "ws" that wouldn't risk causing conflicts.

alebedev87:
> What happens when a route or service with the host name "*.foo" exists and another route or service with the host name "ws.foo" is created? What happens in the reverse case: something with "ws.foo" exists and something with "*.foo" is created?

A conflict of the TXT record name would happen, in both cases. So, external-dns would reject the creation of the conflicting TXT record which would prevent the creation of the "target" DNS record too.

> CoreDNS and SkyDNS treat "any" the same as "*", so we could use that instead of "ws", but "any" isn't standard. It doesn't seem like there is any value we can use in lieu of "ws" that wouldn't risk causing conflicts.

Even if it's not standard, `any` is used in at least 1 implementation and overall looks more educated than just the acronym (`wc` stands for wildcard).

pkg/operator/controller/externaldns/pod.go
```diff
@@ -363,6 +364,9 @@ func (b *externalDNSContainerBuilder) fillAzureFields(zone string, container *co
 	// https://github.com/kubernetes-sigs/external-dns/issues/2082
 	container.Args = addTXTPrefixFlag(container.Args)
 
+	// https://github.com/kubernetes-sigs/external-dns/issues/2922
+	container.Args = append(container.Args, fmt.Sprintf("--txt-wildcard-replacement=%s", defaultTXTWildcardReplacement))
```
gcs278, line 368: Do you think it would be better to limit this option only for `configv1.AzurePlatformType`? Smaller blast radius? Or do you think it's better to be consistent for all platforms?

gcs278:
Not sure if I completely understand this one. Say I want to make a wildcard TXT record for `grant-*.example.com` in azure, so that I can query `grant-a.example.com` and `grant-b.example.com`.

How can I still query `grant-a.example.com` if the `--txt-wildcard-replacement` changes it to `grant-any.example.com`? Am I misunderstanding what field `--txt-wildcard-replacement` is operating on? Does the wildcard query still work?

pkg/operator/controller/externaldns/pod.go
gcs278: Whoops, cancel that, I didn't realize the function was specific to the Azure provider already.

alebedev87:
@gcs278 : Thanks for having a look!

> Does the wildcard query still work?
Yes, the wildcard itself is not impacted by `--txt-wildcard-replacement`. It would still remain with the asterisk.

`--txt-wildcard-replacement` replaces the asterisk character for the corresponding TXT record. ExternalDNS Operator uses [the TXT registry](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/proposal/registry.md#txt-records) of the upstream ExternalDNS to track the ownership of the DNS records (A, CNAME for the moment). So each A or CNAME record is paired with a TXT record which contains [the owner ID](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/externaldns/pod.go#L160) (unique name of externaldns instance). The `--txt-wildcard-replacement`  flag impact only these TXT records.

pkg/operator/controller/externaldns/pod.go
alebedev87: Yes, this PR aims at fixing the problem which is seen only in Azure DNS as it doesn't allow wildcard records (TXT included) which don't start with the asterisk.

Miciah:
> Not sure if I completely understand this one. Say I want to make a wildcard TXT record for `grant-*.example.com` in azure, so that I can query `grant-a.example.com` and `grant-b.example.com`.
> 
> How can I still query `grant-a.example.com` if the `--txt-wildcard-replacement` changes it to `grant-any.example.com`? Am I misunderstanding what field `--txt-wildcard-replacement` is operating on? Does the wildcard query still work?

Per <https://www.rfc-editor.org/rfc/rfc4592#section-2.1.1>, the entire first label must be an asterisk for the record to be a wildcard record.  Accordingly, external-dns only substitutes ([source](https://github.com/kubernetes-sigs/external-dns/blob/cc82c6b727bd38af64454470e0bd97602e7588e6/registry/txt.go#L408-L411)) or matches ([source](https://github.com/kubernetes-sigs/external-dns/blob/cc82c6b727bd38af64454470e0bd97602e7588e6/registry/txt.go#L135-L140)) for "*" if it is equal to the entirety of the first label.

gcs278:
Oh okay, now I'm starting to understand. External DNS is creating TXT records to track what other DNS records it owns. And azure is rejecting TXT records with `*` not in the left most position. This fix just substitutes `*` with other characters.

But one thing I'm still confused about, is how/why a record such as `a-*.example.com` gets created, like in [the issue](https://github.com/kubernetes-sigs/external-dns/issues/2922). Isn't that already an invalid wildcard domain? Is someone purposely creating this type of record or is External DNS creating this `a-*.example.com` record in it's logic?

Miciah:
> But one thing I'm still confused about, is how/why a record such as `a-*.example.com` gets created, like in [the issue](https://github.com/kubernetes-sigs/external-dns/issues/2922). Isn't that already an invalid wildcard domain? Is someone purposely creating this type of record or is External DNS creating this `a-*.example.com` record in it's TXT registry logic?

Yeah, `a-*.example.com` is invalid, or at least imprudent and not preferred, per <https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1>.  DNS might allow it, but I'd expect its use as a host name to be rejected.  I'm surprised that `--txt-wildcard-replacement` resolved that issue.

alebedev87:
> Yeah, a-*.example.com is invalid, or at least imprudent and not preferred, per https://www.rfc-editor.org/rfc/rfc1035#section-2.3.1. DNS might allow it, but I'd expect its use as a host name to be rejected. I'm surprised that --txt-wildcard-replacement resolved that issue.

`a-*.example.com` is the domain name for the TXT record only, the corresponding DNS record would look like `*.example.com`. `a-` is the default prefix added to TXT records by external dns since the version 0.12.0.

gcs278:
> a-*.example.com is the domain name for the TXT record only, the corresponding DNS record would look like *.example.com. a- is the default prefix added to TXT records by external dns since the version 0.12.0.

Okay - that makes a more sense now. I was curious why this was even a problem (aka let's just stop making invalid records). But now I know External DNS is making them.

Okay yikes. Sounds like this is our best path forward at the moment, but @Miciah does bring up a good point about naming collisions. Correct me if I'm wrong, but you can't have `*.example.com` and `any.example.com` co-exist with External DNS on Azure with this PR.

I'm good with this, but I feel hesitant, so @Miciah could you `/lgtm` (since it's already approved, it'll merge after one lgtm).

alebedev87:
>  Correct me if I'm wrong, but you can't have *.example.com and any.example.com co-exist with External DNS on Azure with this PR.

You are right, @gcs278. These 2 records won't be able to be served by ExternalDNS on Azure. This is a limitation (not only in this downstream but upstream too) which I think is preferable over the regression in the support of the wildcard record which we are currently having (since version 0.12.0 of the upstream).

alebedev87:
/assign @xenolinux 

Note this comment: https://github.com/openshift/external-dns-operator/pull/171#issuecomment-1258844722

We need to mention the fact that wildcard TXT records get modified: asterisk is replaced with `any` on AzureDNS. And this may cause a conflict with records starting with `any`. So `any` prefixed records are to be avoided on Azure.

PR #174, opened 2022-09-07
e2e: reduce the polling timeout to speed up the failing case scenario

alebedev87:


alebedev87:
/label docs-approved
/label px-approved
/label qe-approved

Minor change in the e2e test config, no need for the formal acknowledgements.

PR #176, opened 2022-10-18
OCPBUGS-2793: deployment: ignore defaultmode for volumes, accept unsolicited containers

alebedev87:
Some lessons learnt from the other operators:
- `defaultMode` for volumes is not part of the expected specification however it's added by the defaulting mechanism which leads to the fight between the operator and API server
- accept unsolicited containers to enable use cases like side car container injection (service mesh or others)
- `Operator` type: unused fields removed

### DIFF
--- a/pkg/operator/controller/externaldns/deployment.go
+++ b/pkg/operator/controller/externaldns/deployment.go
@@ -363,19 +363,14 @@ func externalDNSAnnotationsChanged(current, expected, updated *appsv1.Deployment
 func externalDNSContainersChanged(current, expected, updated *appsv1.Deployment) bool {
 	changed := false
 
-	// number of container is different: let's reset them all
-	if len(current.Spec.Template.Spec.Containers) != len(expected.Spec.Template.Spec.Containers) {
-		updated.Spec.Template.Spec.Containers = expected.Spec.Template.Spec.Containers
-		return true
-	}
-
 	currentContMap := buildIndexedContainerMap(current.Spec.Template.Spec.Containers)
 	expectedContMap := buildIndexedContainerMap(expected.Spec.Template.Spec.Containers)
 
-	// let's check that all the current containers have the desired values set
-	for currName, currCont := range currentContMap {
-		// if the current container is expected: check its fields
-		if expCont, found := expectedContMap[currName]; found {
+	// ensure all expected containers are present,
+	// unsolicited ones are kept (e.g. service mesh proxy injection)
+	for expName, expCont := range expectedContMap {
+		// expected container is present
+		if currCont, found := currentContMap[expName]; found {
 			if currCont.Image != expCont.Image {
 				updated.Spec.Template.Spec.Containers[currCont.Index].Image = expCont.Image
 				changed = true
@@ -398,9 +393,9 @@ func externalDNSContainersChanged(current, expected, updated *appsv1.Deployment)
 			}
 
 		} else {
-			// if the current container is not expected: let's not dig deeper - reset all
-			updated.Spec.Template.Spec.Containers = expected.Spec.Template.Spec.Containers
-			return true
+			// expected container is not present - add it
+			updated.Spec.Template.Spec.Containers = append(updated.Spec.Template.Spec.Containers, expCont.Container)
+			changed = true
 		}
 	}
 
@@ -422,6 +417,10 @@ func externalDNSVolumesChanged(current, expected, updated *appsv1.Deployment) bo
 	currentVolumeMap := buildIndexedVolumeMap(current.Spec.Template.Spec.Volumes)
 	expectedVolumeMap := buildIndexedVolumeMap(expected.Spec.Template.Spec.Volumes)
 
+	// ignore defaultMode to prevent the fight with the defaulting
+	ignoreFieldsConfigMap := cmpopts.IgnoreFields(corev1.ConfigMapVolumeSource{}, "DefaultMode")
+	ignoreFieldsSecret := cmpopts.IgnoreFields(corev1.SecretVolumeSource{}, "DefaultMode")
+
 	// ensure all expected volumes are present,
 	// unsolicited ones are kept (e.g. kube api token)
 	for expName, expVol := range expectedVolumeMap {
@@ -429,9 +428,7 @@ func externalDNSVolumesChanged(current, expected, updated *appsv1.Deployment) bo
 			updated.Spec.Template.Spec.Volumes = append(updated.Spec.Template.Spec.Volumes, expVol.Volume)
 			changed = true
 		} else {
-			// deepequal is fine here as we don't have more than 1 item
-			// neither in the secret nor in the configmap
-			if !reflect.DeepEqual(currVol.Volume, expVol.Volume) {
+			if !cmp.Equal(currVol.Volume, expVol.Volume, cmpopts.EquateEmpty(), ignoreFieldsConfigMap, ignoreFieldsSecret) {
 				updated.Spec.Template.Spec.Volumes[currVol.Index] = expVol.Volume
 				changed = true
 			}
--- a/pkg/operator/controller/externaldns/deployment_test.go
+++ b/pkg/operator/controller/externaldns/deployment_test.go
@@ -2840,7 +2840,7 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
 			expectedDeployment: testDeploymentWithContainers([]corev1.Container{testContainerWithImage("foo.io/test:latest")}),
 		},
 		{
-			description: "if externalDNS container args",
+			description: "if externalDNS container args change",
 			mutate: func(depl *appsv1.Deployment) {
 				depl.Spec.Template.Spec.Containers[0].Args = []string{"Nada"}
 			},
@@ -2873,15 +2873,12 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
 			description: "if externalDNS has extra container",
 			originalDeployment: testDeploymentWithContainers([]corev1.Container{
 				testContainer(),
-				testContainerWithName("second"),
+				testContainerWithName("extra"),
 			}),
 			mutate: func(depl *appsv1.Deployment) {
 				depl.Spec.Template.Spec.Containers = []corev1.Container{testContainer()}
 			},
-			expect: true,
-			expectedDeployment: testDeploymentWithContainers([]corev1.Container{
-				testContainer(),
-			}),
+			expect: false,
 		},
 		{
 			description: "if externalDNS annotation changes",
@@ -3001,19 +2998,47 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
 			}),
 			mutate: func(dep *appsv1.Deployment) {
 			},
-			expectedDeployment: testDeploymentWithContainers([]corev1.Container{
-				testContainerWithSecurityContext(&corev1.SecurityContext{
-					Capabilities: &corev1.Capabilities{
-						Drop: []corev1.Capability{allCapabilities},
-					},
-					Privileged:               pointer.Bool(false),
-					RunAsNonRoot:             pointer.Bool(true),
-					AllowPrivilegeEscalation: pointer.Bool(false),
-					SeccompProfile: &corev1.SeccompProfile{
-						Type: corev1.SeccompProfileTypeRuntimeDefault,
-					},
-				}),
-			}),
+		},
+		{
+			description: "if externalDNS has extra volume",
+			expect:      false,
+			originalDeployment: testDeploymentWithVolumes(
+				testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets"),
+				testConfigMapVolume("extravolume", "extra", "files", "/etc"),
+			),
+			mutate: func(dep *appsv1.Deployment) {
+				dep.Spec.Template.Spec.Volumes = []corev1.Volume{testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets")}
+			},
+		},
+		{
+			description:        "if API server sets defaultMode on secret volume",
+			expect:             false,
+			originalDeployment: testDeploymentWithVolumes(testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets")),
+			mutate: func(dep *appsv1.Deployment) {
+				dep.Spec.Template.Spec.Volumes[0].VolumeSource.Secret.DefaultMode = nil
+			},
+		},
+		{
+			description:        "if API server sets defaultMode on configmap volume",
+			expect:             false,
+			originalDeployment: testDeploymentWithVolumes(testConfigMapVolume("testcerts", "testcerts", "key", "/etc/pki/trust")),
+			mutate: func(dep *appsv1.Deployment) {
+				dep.Spec.Template.Spec.Volumes[0].VolumeSource.ConfigMap.DefaultMode = nil
+			},
+		},
+		{
+			description: "if volumes change ordering",
+			expect:      false,
+			originalDeployment: testDeploymentWithVolumes(
+				testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets"),
+				testConfigMapVolume("testconfig", "config", "files", "/etc"),
+			),
+			mutate: func(dep *appsv1.Deployment) {
+				dep.Spec.Template.Spec.Volumes = []corev1.Volume{
+					testConfigMapVolume("testconfig", "config", "files", "/etc"),
+					testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets"),
+				}
+			},
 		},
 	}
 
@@ -3026,6 +3051,9 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
 
 			mutated := original.DeepCopy()
 			tc.mutate(mutated)
+			// original is the object that's in the API.
+			// mutated is the new desired object, which mutates the original API object based on some update:
+			// changes to the externaldns, or changes to the operator itself.
 			if changed, updated := externalDNSDeploymentChanged(original, mutated); changed != tc.expect {
 				t.Errorf("Expect externalDNSDeploymentChanged to be %t, got %t", tc.expect, changed)
 			} else if changed {
@@ -3646,7 +3674,7 @@ func TestEnsureExternalDNSDeployment(t *testing.T) {
 								},
 								Containers: []corev1.Container{
 									{
-										Name:  "external-dns-unexpected",
+										Name:  "external-dns-injected",
 										Image: test.OperandImage,
 									},
 								},
@@ -3717,6 +3745,10 @@ func TestEnsureExternalDNSDeployment(t *testing.T) {
 								},
 							},
 							Containers: []corev1.Container{
+								{
+									Name:  "external-dns-injected",
+									Image: test.OperandImage,
+								},
 								{
 									Name:  ExternalDNSContainerName,
 									Image: test.OperandImage,
@@ -4269,7 +4301,11 @@ func TestEnsureExternalDNSDeployment(t *testing.T) {
 								},
 								Containers: []corev1.Container{
 									{
-										Name:  "external-dns-unexpected",
+										Name:  "external-dns-injected-1",
+										Image: test.OperandImage,
+									},
+									{
+										Name:  "external-dns-injected-2",
 										Image: test.OperandImage,
 									},
 								},
@@ -4340,6 +4376,14 @@ func TestEnsureExternalDNSDeployment(t *testing.T) {
 								},
 							},
 							Containers: []corev1.Container{
+								{
+									Name:  "external-dns-injected-1",
+									Image: test.OperandImage,
+								},
+								{
+									Name:  "external-dns-injected-2",
+									Image: test.OperandImage,
+								},
 								{
 									Name:  ExternalDNSContainerName,
 									Image: test.OperandImage,
@@ -5035,6 +5079,12 @@ func testDeploymentWithAnnotations(annotations map[string]string) *appsv1.Deploy
 	return depl
 }
 
+func testDeploymentWithVolumes(volumes ...corev1.Volume) *appsv1.Deployment {
+	depl := testDeployment()
+	depl.Spec.Template.Spec.Volumes = volumes
+	return depl
+}
+
 func testContainer() corev1.Container {
 	return corev1.Container{
 		Name:                     "first",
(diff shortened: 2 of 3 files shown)


alebedev87:
`failed to connect to the bootstrap machine: dial tcp 35.223.24.74:22: connect: connection refused`

/retest

candita:
@alebedev87 can you add a Jira link for this?  @Miciah is going to contact you about this informal story on this one too.

alebedev87:
@candita : I was wondering whether we need a formal Jira bug for things we noticed ourselves which weren't reported by the customer and don't have real bugs (performance improvement at best). Maybe I missed this, is this a requirement now?

candita:
> @candita : I was wondering whether we need a formal Jira bug for things we noticed ourselves which weren't reported by the customer and don't have real bugs (performance improvement at best). Maybe I missed this, is this a requirement now?

Yes, we all create our own Jira bugs for things we've noticed ourselves.  Without it, there is no qe, docs, release note, etc.  On all the other repos there is a requirement for a valid bugzilla in the tide output.  Not sure why it wasn't set up on this repo.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -372,10 +372,11 @@ func externalDNSContainersChanged(current, expected, updated *appsv1.Deployment)
 	currentContMap := buildIndexedContainerMap(current.Spec.Template.Spec.Containers)
 	expectedContMap := buildIndexedContainerMap(expected.Spec.Template.Spec.Containers)
 
-	// let's check that all the current containers have the desired values set
-	for currName, currCont := range currentContMap {
-		// if the current container is expected: check its fields
-		if expCont, found := expectedContMap[currName]; found {
+	// ensure all expected containers are present,
```
candita, line 375: This looks good, but have you considered using a deep copy instead of assignment in line 368, where you reset them all?

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -3015,6 +3015,39 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
...
 		},
+		{
+			description: "if externalDNS has extra volume",
+			expect:      false,
+			originalDeployment: testDeploymentWithVolumes(
+				testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets"),
+				testConfigMapVolume("extravolume", "extra", "files", "/etc"),
+			),
+			mutate: func(dep *appsv1.Deployment) {
+				dep.Spec.Template.Spec.Volumes = []corev1.Volume{testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets")}
```
candita, line 3026: Did you mean to append a new volume here?

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -3015,6 +3015,39 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
...
+				testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets"),
+				testConfigMapVolume("extravolume", "extra", "files", "/etc"),
+			),
+		},
+		{
+			description:        "if ignored field was added to secret volume",
+			expect:             false,
+			originalDeployment: testDeploymentWithVolumes(testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets")),
+			mutate: func(dep *appsv1.Deployment) {
+				dep.Spec.Template.Spec.Volumes[0].VolumeSource.Secret.DefaultMode = nil
```
candita, line 3038:
```suggestion
				dep.Spec.Template.Spec.Volumes[0].VolumeSource.Secret.DefaultMode = somethingthat'snotnil
```

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -3015,6 +3015,39 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
...
+			},
+			expectedDeployment: testDeploymentWithVolumes(testSecretVolume("testcreds", "testsecret", "creds", "/run/secrets")),
+		},
+		{
+			description:        "if ignored field was added to configmap volume",
+			expect:             false,
+			originalDeployment: testDeploymentWithVolumes(testConfigMapVolume("testcerts", "testcerts", "key", "/etc/pki/trust")),
+			mutate: func(dep *appsv1.Deployment) {
+				dep.Spec.Template.Spec.Volumes[0].VolumeSource.ConfigMap.DefaultMode = nil
```
candita, line 3047:
```suggestion
				dep.Spec.Template.Spec.Volumes[0].VolumeSource.ConfigMap.DefaultMode = somethingnotnil
```

candita (commented):
Looks good overall, just a few comments.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: The testcase is weirdly designed. It make my head turn a little every time I approach it. Which is a sign to rewrite it.

`originalDeployment` is current state, `mutate` function changes the current state into the desired deployment. Then the current and desired deployments are passed to `externalDNSDeploymentChanged` which says if the current state needs to be updated and to which state (`expectedDeployment`).

So here, we have 2 volumes in the current (cluster) state and 1 volume as desired by the operator and we expect no changes to the current state as we now allow unsolicited volumes.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: I think it comes form [the misleading test design](https://github.com/openshift/external-dns-operator/pull/176/files#r1002094852). Here we are saying that the desired deployment doesn't have the defaultMode set.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Yes, we could do a deep copy, that would be a little more cpu cycles but a bit safer in terms of the expected containers modifications done via `updated` variable by mistake. But here, we were safe as 1) no other function touches containers and 2) we return immediately so the same function doesn't have a chance to modify `updated` variable.

However your comment made me think about the `if` condition which you pointed into. Actually it prevents the extra containers from being kept in case the lengths of the current `pod.spec.containers` is not the same as expected, so in a lot of cases. I removed this condition to enable the injection of the extra container(s) in any case.

alebedev87:
Failed GCP installation.

/retest

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -363,19 +363,14 @@ func externalDNSAnnotationsChanged(current, expected, updated *appsv1.Deployment
...
 	changed := false
 
-	// number of container is different: let's reset them all
-	if len(current.Spec.Template.Spec.Containers) != len(expected.Spec.Template.Spec.Containers) {
-		updated.Spec.Template.Spec.Containers = expected.Spec.Template.Spec.Containers
-		return true
-	}
-
 	currentContMap := buildIndexedContainerMap(current.Spec.Template.Spec.Containers)
 	expectedContMap := buildIndexedContainerMap(expected.Spec.Template.Spec.Containers)
```
candita, line 367: Is it ok to not have a container map for the `updated.Spec.Template.Spec.Containers`?  You are mixing assignment by index with append in L397, within the same loop - are you sure that's going to work for every scenario?

pkg/operator/controller/externaldns/deployment_test.go
candita: 👍🏼

pkg/operator/controller/externaldns/deployment_test.go
candita: Can you add a test where the defaultMode is set to something other than nil?

pkg/operator/controller/externaldns/deployment.go
alebedev87: > Is it ok to not have a container map for the `updated.Spec.Template.Spec.Containers`?

Yes, the current and the updated are the same in terms of the containers (what this function is supposed to do) at the beginning of the function. That's actually the whole point: `updated == current + updated things`.

> You are mixing assignment by index with append in L397, within the same loop - are you sure that's going to work for every scenario?

That's an interesting finding, didn't think of it indeed. Good enough it doesn't impact the algorithm: the indexes are still pointing to the right containers as we append to the end of the container list and the index container map is not refreshed.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: Done.

pkg/operator/controller/externaldns/deployment_test.go
candita: Actually, you now allow unsolicited containers.  Is it unsolicited volumes are allowed as well as unsoliciated containers?  And neither of them are marked as changes.

api/v1alpha1/groupversion_info.go
```diff
@@ -15,8 +15,8 @@ limitations under the License.
 */
 
 // Package v1alpha1 contains API Schema definitions for the operator v1alpha1 API group
-//+kubebuilder:object:generate=true
-//+groupName=externaldns.olm.openshift.io
+// +kubebuilder:object:generate=true
```
candita, line 18: Do you need to regenerate something after the api changes?

pkg/operator/controller/externaldns/deployment_test.go
candita: > we expect no changes to the current state as we now allow unsolicited volumes.
But shouldn't you expect the expectedDeployment to only have one volume?

pkg/operator/controller/externaldns/deployment.go
candita: Okay

api/v1alpha1/groupversion_info.go
alebedev87: That was just golang 1.19 new formatting of the comment markers, now they need a whitespace after `//`. No need to regenerate anything, if there would be anything we would catch it in the CI: it runs the markers to code and code to manifests generators to see if there is anything not committed.

pkg/operator/controller/externaldns/deployment_test.go
candita: You start with 2 volumes - `testcreds` and `extravolume`.  Then you mutate the volumes to become just one volume `testcreds`.  But your expected deployment still desires the original 2 volumes.  I don't understand the test case - you don't accept the requested change and that is ok?

alebedev87: - I start with 2 volumes `testcreds` and `extravolume` - that the current state.
- Then I mutate to have just `testcreds` - the mutation is to get the desired state (the one which is supposed to be enforced by the operator).
- The expected (output) deployment still desires 2 volumes because the current state has all the desired volumes(`testcreds`) and we accept having `extravolume`. BTW this test case should have been implemented at the time of the volumes change but I missed it so added it here even though it doesn't concern any code change I did in the PR.

I believe it's the mutation which misleads you here. The mutation here is to create the desired state from the current. As I already mentioned the test case is a total mislead, we just need to have 3 fields: `current` `desired` and `expected` and remove all the mutation logic which only confuses everything.

alebedev87: Added a `TODO` to refactor the testcase.

candita: The `tc.mutate` function is not confusing or misleading, it's a common pattern.  You take the current, mutate it, then compare it to the expected state after the mutation.  The part I question is why doesn't the mutation become permanent in the expected outcome?

You mentioned: "The expected (output) deployment still desires 2 volumes because the current state has all the desired volumes(testcreds) and we accept having extravolume.", but if you have applied the mutate function to the current set of volumes, it should have removed that volume.  The expected state should only have the single volume you set in the mutate function.

candita: @Miciah could you take a quick look at this test case and let me know if you agree?

alebedev87: > You take the current, mutate it, then compare it to the expected state after the mutation

No. The mutation creates the desired (by the operator) state.

> but if you have applied the mutate function to the current set of volumes, it should have removed that volume. The expected state should only have the single volume you set in the mutate function.

Desired deployment != expected deployment. The desired deployment is what the operator desires to see. The expected deployment is what we expect to be in the cluster after the operator made the reconciliation. The enforcement of the desired state is done in the soft manner - only the volumes mentioned in the desired deployment has to be present, the rest is left intact. This test case demonstrates exactly this: we have 2 volumes in the cluster and one of them is exactly as desired, so we `expect` no change (`false`). `expectedDepolyment` field here is set but it's not even used as we only check it if `expect` is `true`.

api/v1alpha1/groupversion_info.go
Miciah: This is already in <https://github.com/openshift/external-dns-operator/pull/179/commits/a935fda8a25498c8098f920074ad8fc4915f6f8f#diff-dab150f060822d42b5ee452169a22a98dca6669fef3b58f23a116caf26084955>.  Presumably this change and the one in `api/v1beta1/groupversion_info.go` would vanish with a rebase.

pkg/operator/controller/externaldns/deployment.go
Miciah: `externalDNSDeploymentChanged` already makes a deep copy of `current` to make `update`, so it seems to me that it would be superfluous to do another deep copy here.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -422,16 +417,18 @@ func externalDNSVolumesChanged(current, expected, updated *appsv1.Deployment) bo
...
 	// unsolicited ones are kept (e.g. kube api token)
 	for expName, expVol := range expectedVolumeMap {
 		if currVol, currExists := currentVolumeMap[expName]; !currExists {
 			updated.Spec.Template.Spec.Volumes = append(updated.Spec.Template.Spec.Volumes, expVol.Volume)
 			changed = true
 		} else {
-			// deepequal is fine here as we don't have more than 1 item
-			// neither in the secret nor in the configmap
-			if !reflect.DeepEqual(currVol.Volume, expVol.Volume) {
+			if !cmp.Equal(currVol.Volume, expVol.Volume, ignoreFieldsConfigMap, ignoreFieldsSecret) {
```
Miciah, line 431: Before we changed cluster-ingress-operator to hash the deployment, cluster-ingress-operator similarly used go-cmp to compare volumes—it might be worth a look to see whether there are any logic or unit tests worth adapting to external-dns-operator:

https://github.com/openshift/cluster-ingress-operator/blob/5a82b658c4952e8af5d4658d13e8802b93ded870/pkg/operator/controller/ingress/deployment.go#L359
https://github.com/openshift/cluster-ingress-operator/pull/182/commits/48c17ff3ab3dda4d259d5f26d196490c7ff89d12
https://github.com/openshift/cluster-ingress-operator/pull/182/commits/d32578a2d7e6628fdcb0487fdd941dbaef0c89fe
https://github.com/openshift/cluster-ingress-operator/pull/293/commits/5a82b658c4952e8af5d4658d13e8802b93ded870

For example, the ["if the volumes change ordering" test case](https://github.com/openshift/cluster-ingress-operator/pull/182/commits/48c17ff3ab3dda4d259d5f26d196490c7ff89d12#diff-c008ca2c033fa1a43f2d74d0af1febffdfbaf62e9fae623c90f38e4ecdf979edR333-R340) might be a good one to add to external-dns-operator.  

Cluster-ingress-operator's `cmpSecretVolumeSource` and `cmpConfigMapVolumeSource` also explicitly tested each field in the volume source types, which would reduce the likelihood of problems resulting from defaulting if upstream added new fields in these types.  Explicitly testing each field might not be worth the bother though.

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -2817,6 +2817,8 @@ func TestDesiredExternalDNSDeployment(t *testing.T) {
 }
 
 func TestExternalDNSDeploymentChanged(t *testing.T) {
+	// TODO: refactor this test case to remove the mutation logic as it's misleading.
+	// This test case should have current/desired/expected fields.
```
Miciah, line 2821: In my opinion, it makes the test cases easier to read when each test case only specifies what changes, which is what the `mutate` func shows.  I'm biased though.

Miciah: It is weird, though, that some test cases also specify the original deployment and the expected deployment.

pkg/operator/controller/externaldns/deployment_test.go
Miciah: Andrey's explanations in https://github.com/openshift/external-dns-operator/pull/176#discussion_r1002094852 and https://github.com/openshift/external-dns-operator/pull/176#discussion_r1022873335 make sense to me.  `original` represents whatever is in etcd, `tc.mutate(original.DeepCopy())` represents what the operator expects, and `tc.expect` is the result of reconciling those two things, and since the goal is for the operator to ignore volumes and sidecar containers that some external agent may inject, `tc.expect` should include any additional sidecars or volumes.  

> The unsolicited volumes were accepted even before this PR, that was done to cover the cases when, for instance, the kube api token is added by the api server.

Those only get injected into the pods, not into the deployment's pod template spec.  

> The testcase is weirdly designed. It makes my head turn a little every time I approach it.

This whole discussion reminds me of another: https://github.com/openshift/cluster-ingress-operator/pull/611#discussion_r630455754

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -2873,14 +2875,15 @@ func TestExternalDNSDeploymentChanged(t *testing.T) {
...
+				testContainerWithName("extra"),
 			}),
 			mutate: func(depl *appsv1.Deployment) {
 				depl.Spec.Template.Spec.Containers = []corev1.Container{testContainer()}
 			},
-			expect: true,
+			expect: false,
 			expectedDeployment: testDeploymentWithContainers([]corev1.Container{
 				testContainer(),
+				testContainerWithName("extra"),
```
Miciah, line 2886: If `expect` is false, why specify `expectedDeployment`?

pkg/operator/controller/externaldns/deployment_test.go
Miciah: `nil` seems correct to me.  Maybe the test case would be clearer if it were named something like "if the API sets a default mode on a configmap volume".

api/v1alpha1/groupversion_info.go
alebedev87: Yes, rebased from `main`.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: No reason, removed it in all the test cases to avoid confusion.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Ok, now I see where Steve was inspired from :)

> For example, the ["if the volumes change ordering" test case](https://github.com/openshift/cluster-ingress-operator/pull/182/commits/48c17ff3ab3dda4d259d5f26d196490c7ff89d12#diff-c008ca2c033fa1a43f2d74d0af1febffdfbaf62e9fae623c90f38e4ecdf979edR333-R340) might be a good one to add to external-dns-operator.

The order of the volumes doesn't matter in this comparison as we construct maps from the volume slices and compare by keys, added the test cases to confirm that though.

> Cluster-ingress-operator's cmpSecretVolumeSource and cmpConfigMapVolumeSource also explicitly tested each field in the volume source types, which would reduce the likelihood of problems resulting from defaulting if upstream added new fields in these types. Explicitly testing each field might not be worth the bother though.

To me, the real added value of `cmpSecretVolumeSource` and `cmpConfigMapVolumeSource` is the check of the `defaultMode` if it's set in the desired deployment. `EquateEmpty` options works well [even if specified](https://go.dev/play/p/59LQEP5eXW-) on the top level of the comparison chain, so I added it on the `Volume` level. I'll keep it ignored for the moment, let's revisit the comparison once we'll need to set desired value for it.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: Yes, I think `mutate()` indeed may keep test cases shorter and clearer (clear what changed, I use test case name for this but it's less accurate). However I expected the mutate function to modify the desired (by the operator) state to get the current state while it's the opposite now.
The expected deployment may be different from the desired (e.g. unsolicited containers, volumes). I think it makes sense to verify what `externalDNSDeploymentChanged` sends to the API server at the end, that's why we need `expectedDeployment`.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: Changed the name to the suggested.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: > > The unsolicited volumes were accepted even before this PR, that was done to cover the cases when, for instance, the kube api token is added by the api server.

> 
> Those only get injected into the pods, not into the deployment's pod template spec.

Yes, you are right, I saw the api token volume in the POD, not in the deployment.  Do you think we don't need to be that soft for the volumes?
I personally like the concept of the hard enforcement only of those things which can change the behavior of the operated PODs like `args` `command`, `image` etc. The volumes, in my opinion, can be approached softer - letting the cluster user to add some if it's needed.

> > The testcase is weirdly designed. It makes my head turn a little every time I approach it.

> 
> This whole discussion reminds me of another: https://github.com/openshift/cluster-ingress-operator/pull/611#discussion_r630455754

I agree more with Steve's ordering but that's more religious than factual. Let remove the TODO comment though and add your explanation to the test case, I think it helps understand the idea behind.

Miciah: > Yes, you are right, I saw the api token volume in the POD, not in the deployment. Do you think we don't need to be that soft for the volumes?

I don't think we need to be that permissive about volumes, but it's fine to leave it permissive in this PR (which is focused on allowing sidecars).  

> I personally like the concept of the hard enforcement only of those things which can change the behavior of the operated PODs like `args` `command`, `image` etc. The volumes, in my opinion, can be approached softer - letting the cluster user to add some if it's needed.

The risk of loose enforcement is that someone might start modifying the deployment in unexpected ways, and if some future change in external-dns-operator breaks that customization (for example, adding a conflicting volume with the same name), then that person will complain.  However, we can leave things permissive for the purposes of getting this PR merged.

PR #177, opened 2022-10-18
crd: add short name for ExternalDNS CR

alebedev87:


alebedev87:
Closing as per the discussion with JoelSpeed: since `externaldns` will not be often used, overall shortnames are discouraged.

PR #178, opened 2022-10-18
CFE-613: bump k8s to 1.25.3 including all prerequisites 

alebedev87:
K8S modules 0.25.3 needs golang >= 1.18 hence the update to the latest version available which in turn required the updates of `golangci-lint` and `kustomize`.
`controller-runtime` and `openshift-api` bumps are here to be aligned with the minor K8S version update.


alebedev87:
https://github.com/openshift/external-dns-operator/pull/179

PR #181, opened 2022-10-24
CFE-611: use dedicated quay.io organization for operand image

alebedev87:
The operand image built in [openshift/external-dns](https://github.com/openshift/external-dns) repository is now mirrored to the dedicated quay.io organization. This breaks the dependency on the openshift quay.io organization therefore unlocking the potential for a different life cycle.

### DIFF
--- a/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
@@ -442,7 +442,7 @@ spec:
                     fieldRef:
                       fieldPath: metadata.namespace
                 - name: RELATED_IMAGE_EXTERNAL_DNS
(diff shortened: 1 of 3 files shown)


alebedev87:
Failed Azure installation

/retest

frobware:
Assuming the contents are the same and the location has changed, then...

/lgtm

Adding ahold in case my assumption does not hold true. Feel free to remove if this change just represents a new location for the image.

/hold

alebedev87:
@frobware : you are right, the contents are the same. Just the location which changed: from `openshift` quay.io organization to `external-dns-operator`'s.

CFields651:
label /px-approved

alebedev87:
Azure installation failed.

/retest

PR #185, opened 2022-11-14
OCPBUGS-3630: deployment: enforce recreate strategy to avoid conflicts during upgrades

alebedev87:
As per the upstream update: https://github.com/kubernetes-sigs/external-dns/pull/2772. ExternalDNS is currently unable of handling multiple parallel instances (no leader election implemented). This may result into split brain situation during the rolling upgrades.

### DIFF
--- a/pkg/operator/controller/externaldns/deployment.go
+++ b/pkg/operator/controller/externaldns/deployment.go
@@ -231,6 +231,9 @@ func desiredExternalDNSDeployment(cfg *deploymentConfig) (*appsv1.Deployment, er
 			Selector: &metav1.LabelSelector{
 				MatchLabels: matchLbl,
 			},
+			Strategy: appsv1.DeploymentStrategy{
+				Type: appsv1.RecreateDeploymentStrategyType,
+			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels:      matchLbl,
@@ -332,11 +335,37 @@ func (r *reconciler) updateExternalDNSDeployment(ctx context.Context, current, d
 // externalDNSDeploymentChanged evaluates whether or not a deployment update is necessary.
 // Returns a boolean if an update is necessary, and the deployment resource to update to.
 func externalDNSDeploymentChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
+	changed := false
 	updated := current.DeepCopy()
-	annotationChanged := externalDNSAnnotationsChanged(current, expected, updated)
-	containersChanged := externalDNSContainersChanged(current, expected, updated)
-	volumesChanged := externalDNSVolumesChanged(current, expected, updated)
-	return annotationChanged || containersChanged || volumesChanged, updated
+
+	if expected.Spec.Replicas != nil {
+		expReplicas := *expected.Spec.Replicas
+		if current.Spec.Replicas == nil {
+			updated.Spec.Replicas = &expReplicas
+			changed = true
+		} else if *expected.Spec.Replicas != *current.Spec.Replicas {
+			updated.Spec.Replicas = &expReplicas
+			changed = true
+		}
+	}
+
+	if externalDNSDeploymentStrategyChanged(current, expected, updated) {
+		changed = true
+	}
+
+	if externalDNSAnnotationsChanged(current, expected, updated) {
+		changed = true
+	}
+
+	if externalDNSContainersChanged(current, expected, updated) {
+		changed = true
+	}
+
+	if externalDNSVolumesChanged(current, expected, updated) {
+		changed = true
+	}
+
+	return changed, updated
 }
 
 // externalDNSAnnotationsChanged returns true if any annotation from the podspec differs from the expected.```
@@ -442,6 +471,28 @@ func externalDNSVolumesChanged(current, expected, updated *appsv1.Deployment) bo
 				updated.Spec.Template.Spec.Volumes[currVol.Index] = expVol.Volume
 				changed = true
 			}
+		}
+	}
+
+	return changed
+}
+
+// externalDNSDeploymentStrategyChanged returns true if the current update strategy differs from expected.
+func externalDNSDeploymentStrategyChanged(current, expected, updated *appsv1.Deployment) bool {
+	changed := false
+
+	// reset the strategy type if it differs
+	if expected.Spec.Strategy.Type != "" &&
+		expected.Spec.Strategy.Type != current.Spec.Strategy.Type {
+		updated.Spec.Strategy.Type = expected.Spec.Strategy.Type
+		changed = true
+	}
+
+	// if the expected strategy is Recreate: make sure no rolling update params are present
+	if expected.Spec.Strategy.Type == appsv1.RecreateDeploymentStrategyType {
+		if current.Spec.Strategy.RollingUpdate != nil {
+			updated.Spec.Strategy.RollingUpdate = nil
+			changed = true
 		}
 	}
 
--- a/pkg/operator/controller/externaldns/deployment_test.go
+++ b/pkg/operator/controller/externaldns/deployment_test.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -75,6 +76,8 @@ const (
 )
 
 func TestDesiredExternalDNSDeployment(t *testing.T) {
+	one := int32(1)
+
 	testCases := []struct {
 		name                        string
 		inputSecretName             string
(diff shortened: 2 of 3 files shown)


alebedev87:
Azure cluster install failure.

/retest

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -3081,6 +3091,9 @@ func TestEnsureExternalDNSDeployment(t *testing.T) {
 							"app.kubernetes.io/name":     ExternalDNSBaseName,
 						},
 					},
+					Strategy: appsv1.DeploymentStrategy{
```
candita, line 3094: Why do you need to add the `DeploymentStrategy` to all of these tests?  Isn't is defaulted in `desiredExternalDNSDeployment`?

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -4983,6 +5020,260 @@ func TestSecurityContextChanged(t *testing.T) {
...
+
+	for _, tc := range []struct {
+		name    string
+		current *appsv1.Deployment
+		desired *appsv1.Deployment
+		updated *appsv1.Deployment
+		changed bool
+	}{
+		{
+			name:    "Recreate added for first time",
```
candita, line 5038: For this case, shouldn't you check the `updated` is the same as `desired`?

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -441,6 +471,69 @@ func externalDNSVolumesChanged(current, expected, updated *appsv1.Deployment) bo
...
+	// if the expected strategy is Recreate: make sure no rolling update params are present
+	if expected.Spec.Strategy.Type == appsv1.RecreateDeploymentStrategyType {
+		if current.Spec.Strategy.RollingUpdate != nil {
+			updated.Spec.Strategy.RollingUpdate = nil
+			changed = true
+		}
+	}
+
+	// if the expected strategy is Rolling: set the expected parameters if any
+	if expected.Spec.Strategy.Type == appsv1.RollingUpdateDeploymentStrategyType &&
```
candita, line 494: I thought this PR was to implement the enforcement of the recreate strategy.  Why is the rolling update strategy allowed?

candita:
On hold until comments addressed
/hold

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: Each testcase in this test function sets `expectedDeployment` which is how the deployment should look like after `ensureExternalDNSDeployment` returns.
This test file needs to be refactored to remove the boilerplate. There was already some intent in that direction like this [externaldns CR builder](https://github.com/openshift/external-dns-operator/blob/main/pkg/operator/controller/utils/test/builders.go#L25). Similar builder needs to be done for the deployment.

pkg/operator/controller/externaldns/deployment.go
alebedev87: Yeah, that was the dilemma I had.
At first I implemented `externalDNSDeploymentStrategyChanged` as a strict enforcer of  the`Recreate` strategy only. Then I looked at the name of the function and I saw that it doesn't really do what it says. I got blown away by this idea. Also because I knew that [the supporting of the multiple replicas](https://github.com/kubernetes-sigs/external-dns/issues/2430) is around upstream.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: `updated` field was removed after the support for the desired rolling strategy was droppped.

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -4983,6 +5020,260 @@ func TestSecurityContextChanged(t *testing.T) {
...
+				Spec: appsv1.DeploymentSpec{
+					Strategy: appsv1.DeploymentStrategy{
+						Type: "Recreate",
+					},
+				},
+			},
+			desired: &appsv1.Deployment{
+				Spec: appsv1.DeploymentSpec{
+					Strategy: appsv1.DeploymentStrategy{
+						Type: "RollingUpdate",
```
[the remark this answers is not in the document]
alebedev87: Dropped the support for the rolling strategy, the test case is removed.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -332,11 +335,37 @@ func (r *reconciler) updateExternalDNSDeployment(ctx context.Context, current, d
...
 // Returns a boolean if an update is necessary, and the deployment resource to update to.
 func externalDNSDeploymentChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
+	changed := false
 	updated := current.DeepCopy()
-	annotationChanged := externalDNSAnnotationsChanged(current, expected, updated)
-	containersChanged := externalDNSContainersChanged(current, expected, updated)
-	volumesChanged := externalDNSVolumesChanged(current, expected, updated)
-	return annotationChanged || containersChanged || volumesChanged, updated
+
+	if expected.Spec.Replicas != nil {
```
candita, line 341: Do you have to cover the case where `expected.Spec.Replicas == nil` and `current.Spec.Replicas !=nil`?

pkg/operator/controller/externaldns/deployment_test.go
```diff
@@ -5002,6 +5039,96 @@ func TestSecurityContextChanged(t *testing.T) {
...
+			current: &appsv1.Deployment{
+				Spec: appsv1.DeploymentSpec{
+					Strategy: appsv1.DeploymentStrategy{
+						Type: "Recreate",
+					},
+				},
+			},
+			desired: &appsv1.Deployment{},
+			changed: false,
+		},
```
candita, line 5114: You need a test case here where you try to update it from Recreate to RollingUpdate, and it should not change.

alebedev87: What real life case does this cover? A developer who changes the desired strategy from Rolling to Recreate?

pkg/operator/controller/externaldns/deployment.go
alebedev87: I try to not enforce more than needed to not spawn more fights between the operator and the other actors. If something is not set in the desired state then we don't really care about it and let it be defaulted or set by the user. BTW the replicas field is a good example, it defaults to 1 if not set, then if we'd try to enforce `nil` (given the desired is `nil`) the operator will start fighting for with the API server.

pkg/operator/controller/externaldns/deployment.go
```diff
@@ -448,6 +477,28 @@ func externalDNSVolumesChanged(current, expected, updated *appsv1.Deployment) bo
...
+// externalDNSDeploymentStrategyChanged returns true if the current update strategy differs from expected.
+func externalDNSDeploymentStrategyChanged(current, expected, updated *appsv1.Deployment) bool {
+	changed := false
+
+	// reset the strategy type if it differs
+	if expected.Spec.Strategy.Type != "" &&
+		expected.Spec.Strategy.Type != current.Spec.Strategy.Type {
+		updated.Spec.Strategy.Type = expected.Spec.Strategy.Type
+		changed = true
+	}
```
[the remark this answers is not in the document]
alebedev87: I'm not resetting the whole strategy for the same reason as I'm not setting the replicas field if it's `nil`: to not fight with the api server defaulting.  For the recreate strategy it's less important as it doesn't have any fields other than `Type` however resetting the whole strategy may give a wrong impression that all the types are handled. However for the rolling strategy where `MaxUnavailable` and `MaxSurge` have the defaults we risk to get into the fighting loop if any of them is left unset in the desired state.

pkg/operator/controller/externaldns/deployment_test.go
candita: No, a change from Recreate to Rolling.

pkg/operator/controller/externaldns/deployment.go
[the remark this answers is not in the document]
candita: You should make the changes to the API to enforce the recreate strategy if you think there's going to be flapping to the defaults.

pkg/operator/controller/externaldns/deployment_test.go
alebedev87: If the user changes from recreate to rolling - this case is covered in `Changed from non empty RollingUpdate to Recreate` test.
If the developer changes the desired from the recreate to rolling - there must be a lot more code to be done including the test cases. The code I dropped after your suggestion to not allow the rolling strategy was essentially doing this.

pkg/operator/controller/externaldns/deployment.go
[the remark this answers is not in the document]
alebedev87: About the flapping: I was referring to the potential future improvements (Rolling strategy as desired). For the recreate strategy the resetting of the type is enough as it's the only field in it. BTW which API you were thinking I could make changes to?

pkg/operator/controller/externaldns/deployment_test.go
candita: I am asking about a `Changed from Recreate to RollingUpdate should fail` test.  `from` is the previous state and `to` is the attempted updated state.

alebedev87: But this assumes that `Rolling` is a valid **desired** strategy which we already said we didn't want.

candita: In unit tests, we test invalid stuff to make sure it doesn't work.

alebedev87: I'm sorry but I don't quite see how this test case should look like and how the reconciliation is supposed to fail. What exactly should `externalDNSDeploymentStrategyChanged` function do and return if the expected strategy has been set as`Rolling`?

candita: If the expected strategy is set to `Rolling` and we want to enforce the `Recreate` strategy, then you set it to `Recreate` and validate that.  I can't answer the second question, but it doesn't mean it isn't possible to do so.

pkg/operator/controller/externaldns/deployment.go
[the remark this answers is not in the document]
candita: I suppose the same one you referred to when you said "...to not fight with the api server defaulting."  I still don't understand why you need to fight with the api server defaults, tbh.  I still think you should do it the suggested way to make sure there are no dangling parameters that you have to take care of in the next step.

pkg/operator/controller/externaldns/deployment_test.go
candita: Miciah mentioned a better way to test an attempted change from Recreate to RollingUpdate would be in a different test function, the TestDesiredExternalDNSDeployment.  That would be fine.

pkg/operator/controller/externaldns/deployment.go
candita: I see that expected.Spec.Replicas is always hard-coded to 1, not defaulted, which explains it.

pkg/operator/controller/externaldns/deployment.go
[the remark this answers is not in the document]
candita: I don't think this is the same.  I am just asking you to reset the updated.Spec.Strategy, which includes the Type, instead of just the Type.  The only other parameter is called RollingUpdate, in https://github.com/openshift/external-dns-operator/blob/0cd8bee31e64f3086f9a44e3f5e299a3a1111175/vendor/k8s.io/api/apps/v1/types.go#L407 .  I am asking you to remove the RollingUpdate parameter instead of potentially leaving it behind when you reset the Type from RollingUpdate to Recreate.  It doesn't make sense to leave the RollingUpdate parameter around if you reset the Type from RollingUpdate to Recreate.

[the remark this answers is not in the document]
alebedev87: Yes, I agree the rolling update parameters have to be cleaned and this is what I already do [here](https://github.com/openshift/external-dns-operator/blob/d8e0462d71ecb2d5fc30e45cc9d86e1c082f8d0d/pkg/operator/controller/externaldns/deployment.go#L491-L497). There is a [dedicated unit test case](https://github.com/openshift/external-dns-operator/blob/d8e0462d71ecb2d5fc30e45cc9d86e1c082f8d0d/pkg/operator/controller/externaldns/deployment_test.go#L6008-L6028) for it too.

alebedev87:
Cluster installation failed.

/test e2e-infoblox-operator

alebedev87:
```
Back-off pulling image "registry.redhat.io/redhat/redhat-marketplace-index:v4.12"
```
One of the OLM catalog images is missing, again...

/retest

alebedev87:
Rebased from `main`, new test added for SharedVPC support had to be adopted.

PR #193, opened 2023-04-18
NE-1362: External DNS should honor ingress type

datianshi:
This PR allows end user to add source type as k8s Ingress (Now only OpenShift Route and K8s Service), which upstream already supported long time ago.

The use case can be.

**Given** User uses [AWS Load Balancer Operator](https://docs.openshift.com/container-platform/4.12/networking/aws_load_balancer_operator/install-aws-load-balancer-operator.html)
**And** User uses External DNS Operator 
**When** User uses K8s Ingress to configure AWS **Application Load Balancer**
**Then** The external dns operator creates DNS record based on the Ingress Address (In this case it is ALB CNAME)

And example configuration can be

```
---
apiVersion: externaldns.olm.openshift.io/v1alpha1
kind: ExternalDNS
metadata:
  name: sample-aws 
spec:
  domains:
  - filterType: Include   
    matchType: Exact   
    name: ${domain}
  provider:
    type: AWS
  source:  
    type: Ingress
```

### DIFF
--- a/api/v1alpha1/externaldns_types.go
+++ b/api/v1alpha1/externaldns_types.go
@@ -468,13 +468,14 @@ type ExternalDNSSourceUnion struct {
 	OpenShiftRoute *ExternalDNSOpenShiftRouteOptions `json:"openshiftRouteOptions,omitempty"`
 }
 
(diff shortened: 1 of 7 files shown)


Conan-Kudo (approved):
This looks great to me. 👍🏾

Issue #190, opened 2023-02-20
Support for IRSA for AWS Provider


scottd018:
As per https://github.com/openshift/external-dns-operator/blob/main/api/v1beta1/externaldns_types.go#L260 IRSA is not currently supported.  This should allow a consumer of the external-dns to pass in a Role ARN and have the service account annotated as follows for an AWS provider:

```yaml
metadata:
  annotations:
    eks.amazonaws.com/role-arn: "arn:aws:iam::<ACCOUNT>:role/<ROLE>"
```

scottd018:
See https://github.com/openshift/external-dns-operator/pull/191 for a proposed change to support this feature

Seth-Karlo:
This is something that I also have a customer requesting. Can this be reviewed please?

Thanks!

alebedev87:
@Seth-Karlo , @scottd018 : sorry for the late reply. I'm interested in exploring specific customer cases that require this change. As far as I understand, IRSA only functions with EKS, so please correct me if I'm mistaken. Additionally, since kiam is currently in maintenance mode, it would be challenging to depend on it in the API provided by the operator. However, I'm uncertain about kube2iam.
Having a customer using OpenShift who requires this feature would help to move this request forward. However we will need to go through the RFE process regardless.

scottd018:
> @Seth-Karlo , @scottd018 : sorry for the late reply. I'm interested in exploring specific customer cases that require this change. As far as I understand, IRSA only functions with EKS, so please correct me if I'm mistaken. Additionally, since kiam is currently in maintenance mode, it would be challenging to depend on it in the API provided by the operator. However, I'm uncertain about kube2iam. Having a customer using OpenShift who requires this feature would help to move this request forward. However we will need to go through the RFE process regardless.

@alebedev87 Traditionally, you are correct in that IRSA is an EKS construct.  However, ROSA (Red Hat OpenShift Service on AWS) also employs the IRSA standard and uses the same webhook that EKS uses in order to provide pod-authentication via STS.  Most customers like the idea of using ROSA in STS mode so that they do not have to use hard-coded access/secret keys, so this would be useful for customers using ROSA in STS mode (today, most customers use STS rather than non-STS).

alebedev87:
[...]

PR #195, opened 2023-06-26
NE-1323: Add AWS RoleARN for Shared VPC support

gcs278:
Adds the ability to specify a RoleARN to create Route 53 DNS records in a different AWS Account. Supports Shared VPC scenario.

The new API is propagated to the existing external-dns argument `--aws-assume-role`. 

Epic Link: https://issues.redhat.com/browse/NE-1299
EP PR: https://github.com/openshift/enhancements/pull/1436


api/v1beta1/externaldns_types.go
```diff
@@ -101,6 +101,11 @@ type ExternalDNSSpec struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	Zones []string `json:"zones,omitempty"`
+
+	// AWSRoleARN is a string
+	// +kubebuilder:validation:Optional
+	// +optional
+	AWSRoleARN *string `json:"awsRoleARN,omitempty"`
```
alebedev87, line 108: `RoleARN` is better to be put into [AWSproviderOptions](https://github.com/openshift/external-dns-operator/blob/0cd8bee31e64f3086f9a44e3f5e299a3a1111175/api/v1beta1/externaldns_types.go#L246) as it's AWS specific.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -194,6 +194,10 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
 		args = append(args, "--ignore-hostname-annotation")
 	}
 
+	if b.externalDNS.Spec.AWSRoleARN != nil {
+		args = append(args, fmt.Sprintf("--aws-assume-role=%s", *b.externalDNS.Spec.AWSRoleARN))
```
alebedev87, line 198: I'm wondering whether we should expose [the external ID](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html) as the API knob too. ExternalDNS has [a dedicated flag](https://github.com/PG2000/external-dns/blob/91aaf6803df58463c0380bedbf647e230062b510/pkg/apis/externaldns/types.go#L428) for it. I see that we didn't bother with in [C-I-O](https://github.com/openshift/cluster-ingress-operator/pull/928).

alebedev87 (commented):
I believe you need to add `sts:AssumeRole` to [the AWS CredentilasRequest](https://github.com/openshift/external-dns-operator/blob/842551bf4264f1bc04fe7be9213708bcc5029e4f/pkg/operator/controller/externaldns/credentials_request.go#L232-L255) too.

alebedev87:
@gcs278 : what are your thoughts about testing this feature?

api/v1beta1/externaldns_types.go
gcs278: Yea this was intentionally sloppy to get a POC, but I didn't notice the `provider` field, that makes much more sense, will do soon.

pkg/operator/controller/externaldns/pod.go
gcs278: That's a good point, I hadn't heard of the ExternalID until now. But, if we didn't bother with it in CIO, I wonder if we should bother with it here. I can point it out to the Shared VPC folks so they are aware something like this exists.

api/v1beta1/externaldns_types.go
```diff
@@ -257,6 +257,13 @@ type ExternalDNSAWSProviderOptions struct {
 	// +kubebuilder:validation:Required
 	// +required
 	Credentials SecretReference `json:"credentials"`
+
+	// RoleARN contains the ARN of a IAM role that will be assumed when using the AWS API.
+	// It provides the ability to use a hosted zone in another AWS account.
+	//
+	// +kubebuilder:validation:Optional
```
gcs278, line 264: @alebedev87 Thoughts on adding validation? 

We did for CIO: https://github.com/openshift/api/blob/d976e23f87d346073608c0963d80f8a9e9d03d3a/config/v1/types_dns.go#L133

But, `--aws-assume-role` doesn't validate either. Is this something you think we should offer as a courtesy, or do you think it just makes it more fragile?

alebedev87: I'm fine with the mentioned validation pattern. It's always better to catch problems earlier.

gcs278: Added validation.

api/v1beta1/webhook_test.go
```diff
@@ -175,6 +175,31 @@ var _ = Describe("ExternalDNS admission webhook", func() {
 			Expect(err).ShouldNot(Succeed())
 			Expect(err.Error()).Should(ContainSubstring("credentials secret must be specified when provider type is AWS"))
 		})
+		It("valid RoleARN", func() {
```
gcs278, line 178: @alebedev87 this felt a bit awkward putting this in `webhook_test.go`, since it's not technically testing the webhook, but instead CRD validation.

What are your thoughts? Okay here? Or should we move elsewhere?

gcs278:
> @gcs278 : what are your thoughts about the testing of this feature?

I would like to E2E test it, but we need a E2E test job that stands up a shared VPC in Account A, and stands up a cluster with a role ARN in Account B. My assumption is that we need a new, seperate E2E job that does these things as prerequisites, and separately runs a subset of Shared VPC E2E tests.

I'm waiting on https://issues.redhat.com/browse/CORS-2650 to finish up, as this story will hopefully paving a way, creating items in the CI Step Registry that we could potentially leverage.

Do you have any thoughts?

pkg/operator/controller/externaldns/credentials_request.go
```diff
@@ -248,6 +248,7 @@ func createProviderConfig(externalDNS *operatorv1beta1.ExternalDNS, platformStat
 							"route53:ListHostedZones",
 							"route53:ListResourceRecordSets",
 							"tag:GetResources",
+							"sts:AssumeRole",
```
gcs278, line 251: Based on https://github.com/openshift/cluster-ingress-operator/pull/928/files#diff-63c1fb44f68de2ac21427ab91024d4a6afefb0ddb6a4912f3569ad66e468dc61R29

api/v1beta1/externaldns_types.go
gcs278: @alebedev87 switched validation from CRD to webhook and used `IsARN` https://github.com/openshift/external-dns-operator/blob/838b8fa38be41a4510ceafd255d780319b6a27c6/vendor/github.com/aws/aws-sdk-go/aws/arn/arn.go#L81.

Took idea from https://github.com/openshift/external-dns-operator/issues/190

api/v1beta1/webhook_test.go
gcs278: Well, as the current state, the validation IS in the webhook, so my question here doesn't make sense anymore.

gcs278:
@alebedev87 thanks for your time. Update: 
https://github.com/openshift/external-dns-operator/compare/840c17c4d4ac43b3a7d0bcc51b5cf04b4fe5056a..2c7e1d8c69ab95cb7049fb54b0b8db89b4629edf

1. Moved CRD validation to webhook to use IsARN() from AWS SDK
2. Made spec.provider.aws.credentials optional in CRD
3. Added another webhook validation test for RoleARN in openshift platform
4.  Fixed broken unit tests

pkg/operator/controller/externaldns/pod.go
```diff
@@ -194,6 +194,10 @@ func (b *externalDNSContainerBuilder) fillProviderAgnosticFields(seq int, zone s
 		args = append(args, "--ignore-hostname-annotation")
 	}
 
+	if b.externalDNS.Spec.Provider.AWS != nil && b.externalDNS.Spec.Provider.AWS.AssumeRole != nil && b.externalDNS.Spec.Provider.AWS.AssumeRole.ARN != nil {
+		args = append(args, fmt.Sprintf("--aws-assume-role=%s", *b.externalDNS.Spec.Provider.AWS.AssumeRole.ARN))
+	}
+
```
alebedev87, line 200: In this function we put the flags independent from any provider. `--aws-assume-role` is AWS specific, there is a dedicated method for it: `fillAWSFields`.

alebedev87 (commented):
Overall LGTM, API looks solid. Waiting for the e2e. Just a couple of nitpicks.

api/v1beta1/externaldns_webhook.go
```diff
@@ -155,3 +157,13 @@ func (r *ExternalDNS) validateProviderCredentials() error {
 	}
 	return nil
 }
+
+func (r *ExternalDNS) validateAWSRoleARN() error {
+	// Ensure we have a valid arn if it is specified.
+	provider := r.Spec.Provider
+	if provider.AWS != nil && provider.AWS.AssumeRole != nil && provider.AWS.AssumeRole.ARN != nil && !arn.IsARN(*provider.AWS.AssumeRole.ARN) {
+		return errors.New(fmt.Sprintf("arn %s is not a valid AWS ARN", *provider.AWS.AssumeRole.ARN))
```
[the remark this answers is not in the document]
gcs278: done.

pkg/operator/controller/externaldns/pod.go
gcs278: ah yes, I didn't see that. done.

gcs278:
@alebedev87 I removed the non-compatible API update of changing credentials to optional and added a snippet in the docs. As far as I have tested, this workaround supports OpenShift users using the assumeRole ARN (no need to update webhook or other operator logic).

docs/usage.md
```diff
@@ -55,11 +55,57 @@ spec:
...
+        arn: arn:aws:iam::123456789012:role/role-name # Replace with the desire Role ARN
+  zones: # Replace with the desired hosted zone IDs
+    - "Z3URY6TWQ91KXX"
+  source:
+    type: Service
+    fqdnTemplate:
+    - '{{.Name}}.mydomain.net'
+```
+
+**Note**: Due to a limitation of the `v1beta1` API requiring the `credentials` field, OpenShift users will be required
```
gcs278, line 85: BTW I am planning to add a section on using a Shared VPC to OpenShift's docs: https://docs.openshift.com/container-platform/4.13/networking/external_dns_operator/nw-creating-dns-records-on-aws.html

If you think we don't need both (this repo docs and OpenShift docs), I can drop this one. But I figure better safe than sorry.

I know the docs in this repo are not just for OpenShift users, but it does feel a little strange there isn't a mention of the Cloud Credential Operator providing the credentials for OpenShift anywhere. I wonder if we should introduce that concept somewhere around here, as a new user, that wasn't clear to me I didn't need credentials from reading these docs.

I feel like we should also have a `aws-openshift.md` doc just like the azure and gcp openshift docs in https://github.com/openshift/external-dns-operator/blob/main/docs. That's seems odd to me.

gcs278:
Whoops, missed a `verify` failure telling me to use `fmt.Errorf`.

api/v1beta1/externaldns_types.go
```diff
@@ -257,9 +257,14 @@ type ExternalDNSAWSProviderOptions struct {
 	// +kubebuilder:validation:Required
 	// +required
 	Credentials SecretReference `json:"credentials"`
-	// TODO: Additionally support access for:
-	// - kiam/kube2iam enabled clusters ("iam.amazonaws.com/role" POD's annotation to assume IAM role)
-	// - EKS clusters ("eks.amazonaws.com/role-arn" ServiceAccount's annotation to assume IAM role)
+
+	// AssumeRole is a reference to the IAM role that the
+	// controller will be assuming in order to perform
+	// any DNS updates.
```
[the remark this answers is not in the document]
gcs278: Ack done.

api/v1beta1/externaldns_types.go
```diff
@@ -485,6 +490,15 @@ const (
 	HostnameAnnotationPolicyAllow  HostnameAnnotationPolicy = "Allow"
 )
 
+type ExternalDNSAWSAssumeRoleOptions struct {
+	// ARN is an AWS role ARN that the external-dns
+	// operator will assume when making DNS updates.
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	ARN *string `json:"arn,omitempty"`
```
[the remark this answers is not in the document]
gcs278: Ah good catch thanks. Done.

api/v1beta1/externaldns_webhook.go
```diff
@@ -69,6 +70,7 @@ func (r *ExternalDNS) validate(old runtime.Object) error {
 		r.validateFilters(),
 		r.validateSources(old),
 		r.validateHostnameAnnotationPolicy(),
+		r.validateAWSRoleARN(),
 		r.validateProviderCredentials(),
```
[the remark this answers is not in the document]
gcs278: No reason in particular. Moved it to the bottom - I don't think it matters either way.

docs/usage.md
gcs278: Okay thanks. I'll leave what I have now, and possibly open another PR later with details on `CredentialsRequest` to keep this PR cleaner.

docs/usage.md
```diff
@@ -55,11 +55,57 @@ spec:
 Once this is created the _external-dns-operator_ will create a deployment of _external-dns_ which is configured to
 manage DNS records in AWS Route53.
 
-## AWS GovCloud
+## AWS Assume Role
```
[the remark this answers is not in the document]
gcs278: Removed `AWS` from both.

api/v1beta1/webhook_test.go
```diff
@@ -52,6 +52,33 @@ var _ = Describe("ExternalDNS admission webhook when platform is OCP", func() {
...
 			})
+			It("valid RoleARN", func() {
+				resource := makeExternalDNS("test-valid-rolearn-openshift", nil)
+				resource.Spec.Provider = ExternalDNSProvider{
+					Type: ProviderTypeAWS,
+					AWS: &ExternalDNSAWSProviderOptions{
+						AssumeRole: &ExternalDNSAWSAssumeRoleOptions{
+							ARN: pointer.String("arn:aws:iam::123456789012:role/foo"),
+						},
+					},
```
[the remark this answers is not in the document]
gcs278: This test is for OpenShift, and the CRDs aren't involved, just the webhook. The webhook [doesn't validate ](https://github.com/openshift/external-dns-operator/blob/cc366fce0e8f579525d204ab1085aa88c10361ec/api/v1alpha1/externaldns_webhook.go#L114)credentials for Openshift. The CRDs were the issue due to [`+required`](https://github.com/openshift/external-dns-operator/blob/16721ccd0ccc38a1b1f128d6e12e6b9299578930/api/v1beta1/externaldns_types.go#L258) on the `Credentials` field.

The webhook works without issue, but use of the CRD would require an empty `credentials` field.

gcs278 (commented):
Thanks for the review @alebedev87

I'm working on E2E tests at the moment. I may add them to this PR if it's still open, but I don't think we have to hold this PR up for the E2E tests.

alebedev87:
> I'm working on E2E tests at the moment. I may add them to this PR if it's still open, but I don't think we have to hold this PR up for the E2E tests.

Agree, we can do follow up PRs in this repository if the CI rehearsal will be failing.

api/v1beta1/webhook_test.go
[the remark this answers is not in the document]
alebedev87: Yeah, you are right, I forgot it was the webhook test.

So the particular case when `spec.provider.credentials.name` has to be empty is not tested in this PR. Will it be in a follow up e2e PR?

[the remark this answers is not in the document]
alebedev87: As discussed, the e2e test will be handled in another PR.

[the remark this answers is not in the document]
gcs278: > So the particular case when spec.provider.credentials.name has to be empty is not tested in this PR. Will it be in a follow up e2e PR?

Good point. I'll make a note to add that in the E2E PR I just made: https://github.com/openshift/external-dns-operator/pull/198

gcs278:
@alebedev87 forgot to run `make manifests`

alebedev87:
Cluster installation failed.

/test e2e-gcp-operator

gcs278:
cluster install failure
/test e2e-gcp-operator

gcs278:
Cluster install failed
/test e2e-gcp-operator

PR #198, opened 2023-09-08
NE-1324: E2E tests for Assume Role in Shared VPC Cluster

gcs278:
Add E2E tests for AssumeRole functionality added in https://github.com/openshift/external-dns-operator/pull/195/. It creates a ExternalDNS object with AssumeRole configured inside of a Shared VPC cluster. It then confirms that an example DNS record gets created in a targeted AWS account's hosted zone as well as it queries the DNS record inside the cluster to test end-to-end DNS record functionality.

This E2E test update is also dependent on https://github.com/openshift/release/pull/42894 which adds a new CI test job `e2e-aws-shared-vpc-phz-operator` to this repo. Without this, this new E2E test will be skipped.

Dependent on https://github.com/openshift/external-dns-operator/pull/195/ and will need rebase when it merges.


gcs278:
`An error occurred (HostedZoneAlreadyExists) when calling the CreateHostedZone operation: A hosted zone has already been created with the specified caller reference.`: Seems like the last failed job didn't clean up after itself. Will look into it.

/test  e2e-aws-shared-vpc-phz-operator

gcs278:
Round 1 - Success:
/test e2e-aws-shared-vpc-phz-operator

gcs278:
cluster install fail
/test e2e-azure-operator

gcs278:
https://github.com/openshift/release/pull/43517 merged. testing again.
/test e2e-aws-shared-vpc-phz-operator

test/e2e/operator_test.go
```diff
@@ -678,6 +614,124 @@ func TestExternalDNSSecretCredentialUpdate(t *testing.T) {
...
+func TestExternalDNSAssumeRole(t *testing.T) {
+	// Only run this test if the DNS config contains the privateZoneIAMRole which indicates it's a "Shared VPC" cluster.
+	// Note: Only AWS supports privateZoneIAMRole.
+	dnsConfig := configv1.DNS{}
+	err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &dnsConfig)
+	if err != nil {
+		t.Errorf("Failed to get dns 'cluster': %v\n", err)
+	}
+	if dnsConfig.Spec.Platform.AWS == nil || dnsConfig.Spec.Platform.AWS.PrivateZoneIAMRole == "" {
+		t.Skipf("Test skipped on non-shared-VPC cluster")
```
alebedev87, line 628: Before I start a detailed review I would like to discuss the generic test/code organization.

The SharedVPC/AssumeRole support is different from other test cases. All the other test cases try to fit into the main pattern:
- create DNZ zone
- do platform agnostic tests
- delete DNS zone

In the case of SharedVPC I see some patterns that stand alone:
- DNZ zone is created by the CI
- DNZ zone is private
- Platform dependent - PrivateZoneRole is implemented only for AWS

I see that your code tries to balance in between "be like other test case" and "be specific". From one hand `providerTestHelper` interface is updated meaning that something is meant for all the providers while from the other hand the new test case is skipped for all the providers by AWS.

To make things more coherent I see 2 approaches:
1) Make an effort and generalize AssumeRole test case:
  -  add `helper.buildExternalDNSWithRole` method to `providerTestHelper` interface: AWS would get the role from the cluster's DNS config, other platforms would call the existing`buildExternalDNS()`.
  - keep the DNS record listing as part of  `providerTestHelper` interface but make the listing against different DNS zones: for AWS - from the other account, for the rest - from the same account.
  - `dig` probing can be kept as is
  
2) Move the new test case into a different file and update CI to run only the new test case in the shared-vpc CI job.

The first approach assumes that SharedVPC/AssumeRole features will come to the other providers too but makes a lot of unnecessary moves . The second approach is not generic but targets precisely the new feature.

gcs278: Good discussion. A lot of ways to proceed. I struggled a bit with this concept when writing this E2E. Couple of thoughts and questions:

- Sounds like the core question is whether it could ever be extended to other platforms. I see evidence (https://github.com/kubernetes-sigs/external-dns/issues/509) that folks are using service accounts to fulfill the assuming of a role in GCP.
    - It's possible that it could be extended, right? Or maybe there is a fundamental design difference that assume-role embedded in External DNS only makes sense in AWS.
- In addition to testing our new assume role functionality, is it reasonable to test non-assume-role functionality on a Shared VPC cluster?
   - At first, I thought this would a good test case, but the more I think about it, the more redundant it seems. Testing External DNS on a Shared VPC AWS cluster without using the Assume Role is not different than testing it in a non-Shared VPC vanilla AWS cluster.

> add helper.buildExternalDNSWithRole method to providerTestHelper interface: AWS would get the role from the cluster's DNS config, other platforms would call the existingbuildExternalDNS().

:+1: Except, I think you mean `AWS Shared VPC` instead of just `AWS`. Are you implying we should run the existing E2E tests under the `AWS Shared VPC` test job? It's possible, but messy. We are using a private hosted zone, so we'd need to change the logic to validate the test differently.

> keep the DNS record listing as part of providerTestHelper interface but make the listing against different DNS zones: for AWS - from the other account, for the rest - from the same account.

Are you suggesting to treat AWS Shared VPC as it's own platform (or provider)? i.e. add a new `awsAssumeRoleHelper` with a new file `aws_assume_role.go`? This could work, but we'd have to turn off the other non-assume-role E2E tests (like I mentioned above) and only run `TestExternalDNSAssumeRole` (due to private zone limitation).

I think my questions are:
1. Are you saying we should run the existing E2E tests under the `AWS Shared VPC` test job? And basically remove `TestExternalDNSAssumeRole` as a specific test?
2. Should `AWS Shared VPC` be treated as a it's own platform? I.e. in https://github.com/openshift/external-dns-operator/blob/829cd863bd5cbccdbcbb8f004d2c43fbd2ca229d/test/e2e/operator_test.go#L108-L120

alebedev87: > Sounds like the core question is whether it could ever be extended to other platforms. I see evidence (Unable to use IAM Service Account on GKE kubernetes-sigs/external-dns#509) that folks are using service accounts to fulfill the assuming of a role in GCP.
>
>    It's possible that it could be extended, right? Or maybe there is a fundamental design difference that assume-role embedded in External DNS only makes sense in AWS.

Indeed, GCP seems to provide [SharedVPC](https://cloud.google.com/dns/docs/zones/zones-overview#cross-project_binding) feature too with a possibility to make [cross-projects DNS updates](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/gke.md#cross-project-access-scenario-using-google-service-account). However the way ExternalDNS is granted the permission seems to be different. The credentials are supposed to be from the other (not cluster's) project, they can be [packed as a secret](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/gke.md#static-credentials) and feed to ExternalDNS through [the credentials field](https://github.com/openshift/external-dns-operator/blob/main/api/v1beta1/externaldns_types.go#L288) because `CredentialsRequest` will provision credentials for the cluster's project. However I don't fully exclude the possibility that AssumeRole API will not be propagated (possibly in a different form) to other providers. I didn't see Azure and even GCP may be somehow adopted if not upstream  then in ExtDNS Operator.

> Except, I think you mean AWS Shared VPC instead of just AWS. Are you implying we should run the existing E2E tests under the AWS Shared VPC test job? It's possible, but messy. We are using a private hosted zone, so we'd need to change the logic to validate the test differently.

No, I meant AWS test helper.  Only the new test scenario can use the new `buildExternalDNSWithRole` method. The rest of the scenarios are still using the zone created the e2e test in `TestMain`. Even in the current implementation that is what's happening - all the old test cases are still using the public DNS zone from the cluster's account.

> Are you saying we should run the existing E2E tests under the AWS Shared VPC test job? And basically remove TestExternalDNSAssumeRole as a specific test?
> Should AWS Shared VPC be treated as a it's own platform?

Neither. The CI job you added is AWS platform (from the e2e point of view), it would run the same tests as any other platform. Only `TestExternalDNSAssumeRole` test case would be different - it will read the private DNZ zone ID and IAM role from the DNSConfig and return `ExternalDNS` with `AssumeRole` field filled in. The old AWS CI job would still run the old test plus the new `TestExternalDNSAssumeRole` but since there will be no private DNS zone ID or role in the dns config the `ExternalDNS` instance returned will not have `AssumeRole` field.
So this way the old AWS CI job and the one you added are almost the same except for 1 test case. Redundant but `TestExternalDNSAssumeRole` will fit into the test battery.


I think that I start to lean more towards the approach where your new CI job runs only 1 test - `TestExternalDNSAssumeRole` and this test is not even added to `operator_test.go` file but put into a dedicated "AWS specific" file. `TestExternalDNSAssumeRole` can be copied as is, it still can use all the utils and constants, it just doesn't need to be part of the framework build in `TestMain`.

test/common/common.go
```diff
@@ -0,0 +1,446 @@
...
+						AllowPrivilegeEscalation: pointer.Bool(false),
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
+					},
+				},
+			},
+			RestartPolicy: corev1.RestartPolicyNever,
+		},
+	}
```
alebedev87, line 247:
```suggestion
	return &corev1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			Name:      name,
			Namespace: namespace,
		},
		Spec: corev1.PodSpec{
			Containers: []corev1.Container{
				{
					Name:    "dig",
					Image:   "openshift/origin-node",
					Command: []string{"/bin/dig"},
					Args:    []string{
					    address,
					    "+noall",
		                            "+answer",
		                            "+comments",
					},
					SecurityContext: &corev1.SecurityContext{
						Capabilities: &corev1.Capabilities{
							Drop: []corev1.Capability{"ALL"},
						},
						Privileged:               pointer.Bool(false),
						RunAsNonRoot:             pointer.Bool(true),
						AllowPrivilegeEscalation: pointer.Bool(false),
						SeccompProfile: &corev1.SeccompProfile{
							Type: corev1.SeccompProfileTypeRuntimeDefault,
						},
					},
				},
			},
			RestartPolicy: corev1.RestartPolicyNever,
		},
	}
```

Query options are documented to be at the end:
```
SYNOPSIS
       dig [@server] [-b address] [-c class] [-f filename] [-k filename] [-m] [-p port#] [-q name] [-t type] [-v]
           [-x addr] [-y [hmac:]name:key] [[-4] | [-6]] [name] [type] [class] [queryopt...]
```

test/common/common.go
```diff
@@ -0,0 +1,446 @@
...
+	if len(ips) == 0 {
+		return nil, fmt.Errorf("not found")
+	}
+	return ips, nil
+}
+
+// DnsQueryClusterInternal queries for a DNS hostname inside the VPC of a cluster using a dig pod. It returns a
+// map structure of the resolved IPs for easy lookup.
+func DnsQueryClusterInternal(ctx context.Context, t *testing.T, cl client.Client, cs *kubernetes.Clientset, namespace, hostname string) (map[string]struct{}, error) {
```
alebedev87, line 123:
```suggestion
func LookupARecordInternal(ctx context.Context, t *testing.T, cl client.Client, cs *kubernetes.Clientset, namespace, host string) ([]string, error) {
```
To make lookup functions look alike, WDYT?

test/common/common.go
```diff
@@ -0,0 +1,446 @@
...
+	defer func() {
+		_ = cl.Delete(ctx, clientPod)
+	}()
+
+	// Loop until dig pod starts, then parse logs for query results.
+	var responseCode string
+	var gotIPs map[string]struct{}
+	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 10*time.Minute, true, func(ctx context.Context) (bool, error) {
+		if err := cl.Get(ctx, types.NamespacedName{Name: clientPod.Name, Namespace: clientPod.Namespace}, clientPod); err != nil {
+			t.Errorf("Failed to get pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
```
alebedev87, line 139: Same polling question as before. Is this error fatal or we would like to retry? We have different answers for different pollings.

test/common/common.go
```diff
@@ -0,0 +1,446 @@
...
+		gotIPs = make(map[string]struct{})
+		for scanner.Scan() {
+			line := scanner.Text()
+
+			// Skip blank lines.
+			if strings.TrimSpace(line) == "" {
+				continue
+			}
+			// Parse status out (helpful for future debugging)
+			if strings.HasPrefix(line, ";;") && strings.Contains(line, "status:") {
```
alebedev87, line 181: Why reason to not use `+short` query option which prints only the target address without anything else?

test/common/common.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+		t.Logf("Got IPs: %v", gotIPs)
+		return true, nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to observe the expected dig results: %v", err)
+	}
+
+	// Need to delete pod in case we start again.
+	waitForDeletion(ctx, t, cl, clientPod, 5*time.Minute)
```
alebedev87, line 205: Why do we need a differed delete of the POD set up before? Btw don't we want these logics to be the same?

test/common/common.go
```diff
@@ -0,0 +1,446 @@
...
+// buildDigPod returns a pod definition for a pod with the given name and image
+// and in the given namespace that digs the specified address.
+func buildDigPod(name, namespace, address string, extraArgs ...string) *corev1.Pod {
+	digArgs := []string{
+		"+noall",
+		"+answer",
+		"+comments",
+	}
+	digArgs = append(digArgs, extraArgs...)
+	digArgs = append(digArgs, address)
```
alebedev87, line 219:
```suggestion
	digArgs := []string{
	        address,
		"+noall",
		"+answer",
		"+comments",
	}
	digArgs = append(digArgs, extraArgs...)
```
Query options are often at the end:
```
SYNOPSIS
       dig [@server] [-b address] [-c class] [-f filename] [-k filename] [-m] [-p port#] [-q name] [-t type] [-v]
           [-x addr] [-y [hmac:]name:key] [[-4] | [-6]] [name] [type] [class] [queryopt...]
```

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	t.Logf("Querying for DNS record inside cluster VPC using a dig pod")
+
+	// Verify that the IPs of the record created by ExternalDNS match the IPs of load balancer obtained in the previous
+	// step. We will start a pod that runs a dig command for the expected hostname and parse the dig output.
+	if err := wait.PollUntilContextTimeout(context.TODO(), common.DnsPollingInterval, common.DnsPollingTimeout, true, func(ctx context.Context) (done bool, err error) {
+		gotIPs, err := common.DnsQueryClusterInternal(ctx, t, kubeClient, kubeClientSet, testNamespace, expectedHost)
+		if err != nil {
+			t.Errorf("Failed to query hostname inside cluster: %v", err)
```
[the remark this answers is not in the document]
gcs278: Are you referring to the `t.Errorf`? Yea I should switch that to `t.Logf`.

So I am a little confused.  I thought `t.Errorf` isn't supposed to fail the test, but just classify the log as an "error" and keep going.

However, in my testing, if there was a `t.Errorf` logged, the test would keep going, then it would fail at the end since the last log statement was an `t.Errorf`.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+			return false, nil
+		}
+
+		// If all IPs of the loadbalancer are not present, query again.
+		if len(gotIPs) == 0 {
+			return false, nil
```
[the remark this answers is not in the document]
gcs278: Good point. Done.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}()
+
+	// Create a service of type LoadBalancer with the annotation targeted by the ExternalDNS resource.
+	t.Log("Creating source service")
+	service := common.DefaultService(testServiceName, testNamespace)
+	if err := kubeClient.Create(context.TODO(), service); err != nil {
+		t.Fatalf("Failed to create test service %s/%s: %v", testNamespace, testServiceName, err)
+	}
+	defer func() {
+		_ = kubeClient.Delete(context.TODO(), service)
```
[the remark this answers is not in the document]
gcs278: Hmm, yes that is true. Though, for existing tests, we delete the hosted zone: https://github.com/openshift/external-dns-operator/blob/d04af909927c7c9f47d308ae14e9744b7842c238/test/e2e/operator_test.go#L137

So we can be sure they eventually get deleted. 

For this new Shared VPC e2e test, we just delete service, but the private hosted zone doesn't get deleted (because we didn't create it).

What do you think about adding a function in the `defer` for deleting the service to wait for ExternalDNS to delete the records (and error if they weren't cleaned up)? Right now, it may be a race condition situation (1. delete service, then immediately 2. delete ExternalDNS object).

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}); err != nil {
+		t.Logf("Failed to verify that DNS has been correctly set.")
+	} else {
+		return
+	}
+	t.Fatalf("All nameservers failed to verify that DNS has been correctly set.")
+}
+
+func buildExternalDNSAssumeRole(name, zoneID, zoneDomain, roleArn string) operatorv1beta1.ExternalDNS {
+	extDns := operatorv1beta1.ExternalDNS{
```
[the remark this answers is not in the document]
alebedev87: Well, yes it can be a pointer as you need a pointer after you call this function.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	awsSession := session.Must(session.NewSession(&aws.Config{
+		Credentials: credentials.NewStaticCredentials(keyID, secretKey, ""),
+	}))
+
+	r53AssumeRoleClient := route53.New(awsSession)
+	r53AssumeRoleClient.Config.WithCredentials(stscreds.NewCredentials(awsSession, roleARN))
+
+	return nil, r53AssumeRoleClient
```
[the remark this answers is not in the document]
alebedev87: Sorry, my bad, I didn't check the signature.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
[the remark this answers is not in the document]
alebedev87: Yes, `T.Errorf()` continues the test execution but marks the test as failed.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+	if err := operatorv1.Install(scheme); err != nil {
+		panic(err)
+	}
+	if err := routev1.Install(scheme); err != nil {
+		panic(err)
+	}
+	if err := olmv1alpha1.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
+}
```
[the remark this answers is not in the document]
alebedev87: I think `scheme` variable from this package is a copy of the client-go's default scheme which is a pointer. So we take the default scheme and add Route OLM and other non Kube APIs to it.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	recordList := make(map[string]struct{})
+	if records.ResourceRecordSets[0].AliasTarget != nil {
+		recordList[*records.ResourceRecordSets[0].AliasTarget.DNSName] = struct{}{}
+	} else {
+		for _, record := range records.ResourceRecordSets[0].ResourceRecords {
+			recordList[*record.Value] = struct{}{}
+		}
+	}
```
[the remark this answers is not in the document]
alebedev87: Sorry, my bad, I focused on `[0]` index and didn't see `ResourceRecords` slice.

test/common/common.go
```diff
@@ -0,0 +1,446 @@
...
+	e2eSkipDNSProvidersEnvVar  = "E2E_SKIP_DNS_PROVIDERS"
+	DnsPollingInterval         = 15 * time.Second
+	DnsPollingTimeout          = 3 * time.Minute
+)
+
+func DefaultService(name, namespace string) *corev1.Service {
+	return testService(name, namespace, corev1.ServiceTypeLoadBalancer)
+}
+
+func testService(name, namespace string, svcType corev1.ServiceType) *corev1.Service {
```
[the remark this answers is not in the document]
gcs278: So I just copied and pasted the functions from the original `util.go`. This indirection was introduced by this commit: https://github.com/openshift/external-dns-operator/commit/f52e31052b4ec31164239b058dbb880320aea2dd, but I don't think it was ever used.

It looks like `clusterIPService` got removed in https://github.com/openshift/external-dns-operator/commit/c5de001aa96cace10ce9ecc36c20a4eeb415d4f4

I can remove the indirection now since I've touched it.

test/common/common.go
gcs278: Yes, consistency is good. Done.

test/common/common.go
gcs278: Yea sorry I think I misunderstood `t.Errorf`. Fixed.

test/common/common.go
gcs278: That's fair, done. I also added `"A"` to be explicit for the `A` record type (even though dig defaults to `A`).

test/common/common.go
gcs278: Ah yea good catch. I think this was here before I broke this into a new function call. Now the `defer` executes when the function exits.

gcs278 (commented):
Sorry, I missed a whole file.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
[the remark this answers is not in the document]
gcs278: Oh I see. Done.

gcs278:
@alebedev87 I should have addressed everything in your review, but I still need to work confirming that DNS records are deleted.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,307 @@
...
+		if _, found := recordValues[serviceAddress]; !found {
+			if _, foundWithDot := recordValues[serviceAddress+"."]; !foundWithDot {
+				t.Logf("DNS record with name %q didn't contain expected service IP %q", expectedHost, serviceAddress)
+				return false, nil
+			}
+		}
+		t.Logf("DNS record with name %q found in shared Route 53 private zone %q and matched service IPs", expectedHost, hostedZoneID)
+		return true, nil
+	}); err != nil {
+		t.Fatalf("Failed to verify that DNS has been correctly set.")
```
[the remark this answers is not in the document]
alebedev87: I thought "in a different account" would fit this check better than the next one where you put it. In the next one we don't really know which account the DNS record is defined in so more generic "Failed to verify that DNS has been correctly set" seems to fit better.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,281 @@
...
+	r53ClientAssumeRole *route53.Route53
+	roleARN             string
+	scheme              *runtime.Scheme
+	hostedZoneID        string
+	hostedZoneDomain    string
+)
+
+func init() {
+	common.InitScheme(scheme)
+}
```
alebedev87, line 49: Why not calling `initScheme()` in `common`'s `init()` and use `common.Scheme` in this package?

test/e2e_awssharedvpc/aws_shared_vpc_test.go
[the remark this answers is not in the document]
gcs278: Whoops, fixed that, also realized that was a `t.Logf` and not a `t.Fatalf`.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
gcs278: Yep that's fair. Done.

test/common/common.go
```diff
@@ -0,0 +1,483 @@
...
+// map structure of the resolved IPs for easy lookup.
+func LookupARecordInternal(ctx context.Context, t *testing.T, cl client.Client, cs *kubernetes.Clientset, namespace, host string) (map[string]struct{}, error) {
+	t.Helper()
+
+	clientPod := buildDigPod("digpod", namespace, host)
+	if err := cl.Create(ctx, clientPod); err != nil {
+		return nil, fmt.Errorf("failed to create pod %s/%s: %v", clientPod.Namespace, clientPod.Name, err)
+	}
+	defer func() {
+		waitForDeletion(ctx, t, cl, clientPod, 5*time.Minute)
```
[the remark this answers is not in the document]
gcs278: I think the motivation was to ensure the pod is gone before starting another dig pod query attempt, to avoid a naming collision (since they use the same name). I believe the API delete call isn't blocking.

I've updated to use a random pod name suffix (to ensure no naming collisions), and to call the delete without blocking.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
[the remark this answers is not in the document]
gcs278: Update: Added logic in the new E2E test that deletes the service and then waits for the DNS record to get deleted.

gcs278:
@alebedev87 Addressed latest comments, also missed a couple of other things: https://github.com/openshift/external-dns-operator/compare/e2330ab8695bfd3d3e0fd803bc95a547772e230a..cd5ab109630a647b08c711a96e5ba12e0e39a699

- Delete logic for the service (removed `waitForDeletion`)
- Using `KubeClient` in common, not really necessary to pass it around.
- Added check for DNS record after service is deleted
- `GetPlatformType` had `error` return type first
- bunch of `t.Logf` or `t.Fatalf` when they could have been `t.Log` or `t.Fatal`

test/common/common.go
```diff
@@ -0,0 +1,472 @@
...
+			},
+		},
+	}
+	return KubeClient.Create(ctx, &rb)
+}
+
+func initKubeClient() error {
+	kubeConfig, err := config.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get kube config: %v", err)
```
alebedev87, line 402: The approach taken in `initScheme` with a `panic` seems to be logical for an init function. WDYT about aligning them?

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,291 @@
...
+	if err = wait.PollUntilContextTimeout(context.TODO(), common.DnsPollingInterval, common.DnsPollingTimeout, true, func(ctx context.Context) (done bool, err error) {
+		recordValues, err := getResourceRecordValues(hostedZoneID, expectedHost, "A")
+		if len(recordValues) != 0 {
+			t.Logf("Waiting for DNS record %q to be deleted.", expectedHost)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("Failed to verify that ExternalDNS deleted DNS records: %v", err)
+	}
```
alebedev87, line 201: I think this should be wrapped into `defer` statement right after the service creation. Otherwise we risk to not make sure the DNS disappers if the test `Fatal`s before reaching the end.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
```diff
@@ -0,0 +1,291 @@
...
+	// Only run this test if the DNS config contains the privateZoneIAMRole which indicates it's a "Shared VPC" cluster.
+	// Note: Only AWS supports privateZoneIAMRole.
+	dnsConfig := configv1.DNS{}
+	err = common.KubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, &dnsConfig)
+	if err != nil {
+		fmt.Printf("Failed to get dns 'cluster': %v\n", err)
+		os.Exit(1)
+	}
+	if dnsConfig.Spec.Platform.AWS == nil || dnsConfig.Spec.Platform.AWS.PrivateZoneIAMRole == "" {
+		fmt.Printf("Test skipped on non-shared-VPC cluster\n")
```
alebedev87, line 72:
```suggestion
		fmt.Println("Test skipped on non-shared-VPC cluster")
```

alebedev87 (commented):
Looks fine to me, just a couple of last minute nit picks and 1 comment about the checking of the DNS record removal.

test/e2e_awssharedvpc/aws_shared_vpc_test.go
gcs278: Ah yea you are right. Done.

gcs278:
@alebedev87 ready for another round. Fixed your review comments, but also added a `make test-e2e-sharedvpc` alias for easy run of sharedvpc e2e tests.

Also, I renamed the directory name from `e2e_awssharedvpc` to `e2e_sharedvpc`. The package name was already `e2e_sharedvpc` and I think it's probably better to be platform generic in hopes maybe it will be supported on other platforms.

Makefile
```diff
@@ -122,6 +122,17 @@ test-e2e:
...
+.PHONY: test-e2e-sharedvpc
+test-e2e-sharedvpc:
+	go test \
+	$(GOBUILD_VERSION_ARGS) \
+	-timeout $(E2E_TIMEOUT) \
+	-count 1 \
+	-v \
+	-tags e2e \
+	-run "$(TEST)" \
+	./test/e2e_sharedvpc
```
alebedev87, line 134:
```suggestion
.PHONY: test-e2e-sharedvpc
test-e2e-sharedvpc:
	TEST_PKG="./test/e2e_sharedvpc" make test-e2e
	
```

gcs278: Done.

gcs278:
`* 2023-10-12T22:10:23Z 3x kubelet: Failed to create pod sandbox: rpc error: code = DeadlineExceeded desc = context deadline exceeded`

/test unit
/test verify

gcs278:
`test "verify" failed: pod pending for more than 30m0s: containers have not started in 32m26.844693716s: ci-scheduling-dns-wait, place-entrypoint, sidecar, test: `

/test verify
/test unit

gcs278:
cluster install
/test e2e-azure-infoblox-operator

gcs278:
Applying the labels myself as:
1. https://github.com/openshift/external-dns-operator/pull/195 as already ack'ed by everyone.
2. This is an E2E test, that has no user-facing changes to document or test.

/label docs-approved
/label px-approved
/label qe-approved

PR #199, opened 2023-09-28
NE-1323: Add default to AWS credentials field

gcs278:
By adding a default value to the `spec.provider.aws.credentials` field, it enables the configuration of `spec.provider.aws.assumeRole` for an OpenShift user that doesn't need to explicitly specify the `credentials` field. Otherwise, an OpenShift user must manually insert `""` as the credentials name to use the `assumeRole` field.

This improves user experience for the current v1beta1 release of ExternalDNS Operator while not introducing any breaking changes to the API. We are planning to properly fix this issue in a future ExternalDNS Operator API version by converting `Credentials` to an optional pointer.

Epic: https://issues.redhat.com/browse/NE-1299

### DIFF
--- a/api/v1beta1/externaldns_types.go
+++ b/api/v1beta1/externaldns_types.go
@@ -255,6 +255,7 @@ type ExternalDNSAWSProviderOptions struct {
 	// for more information.
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:default:={"name":""}
 	// +required
 	Credentials SecretReference `json:"credentials"`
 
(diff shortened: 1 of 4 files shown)


gcs278:
Going to put a hold on this while we get approval from docs, px, and qe and finish up enhancement review.
/hold

gcs278:
/unhold
We have approval on https://github.com/openshift/enhancements/pull/1436 (minus some minor comments) so I am proceeding with the implementation PRs.
CC: @Miciah

PR #205, opened 2023-11-13
NE-1390: stable-v1.2 channel

gcs278:
Add a new OLM channel for the upcoming `1.2` release.


gcs278:
Looks like cluster install failure
/test e2e-azure-operator

PR #206, opened 2023-11-20
OCPBUGS-23562: Add file based OLM catalog and remove SQLite index commands

alebedev87:
The SQLite based catalog format is [deprecated since OCP 4.11](https://docs.openshift.com/container-platform/4.11/operators/admin/olm-managing-custom-catalogs.html#olm-managing-custom-catalogs-sqlite). This PR adds [the file based catalog](https://docs.openshift.com/container-platform/4.14/operators/admin/olm-managing-custom-catalogs.html#olm-managing-custom-catalogs-fb) which is the latest iteration of the catalog format. [Creating file based catalog image doc](https://docs.openshift.com/container-platform/4.14/operators/admin/olm-managing-custom-catalogs.html#olm-creating-fb-catalog-image_olm-managing-custom-catalogs) was loosely followed to create the file based catalog. The commands which use SQLite based catalog are removed.


Dockerfile.catalog
```diff
@@ -0,0 +1,14 @@
...
+# /bin/opm (with a serve subcommand) and /bin/grpc_health_probe
+FROM registry.redhat.io/openshift4/ose-operator-registry:v4.14
+
+# Configure the entrypoint and command
+ENTRYPOINT ["/bin/opm"]
+CMD ["serve", "/configs"]
+
+# Copy declarative config root into image at /configs
+ADD catalog /configs
+
```
rfredette, line 11: In [ALBO's catalog.Dockerfile](https://github.com/openshift/aws-load-balancer-operator/blob/721cf7c520603efcc9f81ee7e218bc6c66fda575/catalog.Dockerfile), it looks like opm uses a cache (`"--cache-dir=/tmp/cache"`). Is there a reason we'd want the cached version there but not here?

Dockerfile.catalog
```diff
@@ -0,0 +1,14 @@
+# The base image is expected to contain
+# /bin/opm (with a serve subcommand) and /bin/grpc_health_probe
+FROM registry.redhat.io/openshift4/ose-operator-registry:v4.14
```
rfredette, line 3: Since this has been waiting for review for a while (my apologies), it looks like there are newer versions of the `ose-operator-registry` image. Should we update this to `v4.16`?

rfredette (commented):
I have a couple questions, but otherwise this looks good to me.

Dockerfile.catalog
alebedev87: Good catch. I don't remember exactly why they are different. Probably I generated both but for ALBO I also looked at how other operators did this. Both approaches work but I think I would align this one with ALBO (and other operators in `openshift` org).

Dockerfile.catalog
alebedev87: ` registry.redhat.io/openshift4/ose-operator-registry` image seems to be discontinued, `v4.15` and `v4.16` tags are not there. [The 4.16 OCP documentation](https://docs.openshift.com/container-platform/4.16/operators/admin/olm-managing-custom-catalogs.html#olm-creating-fb-catalog-image_olm-managing-custom-catalogs) suggests to use `registry.redhat.io/openshift4/ose-operator-registry-rhel9:v4.16`. It's fine with me because we would like to do RHEL9 migration in this ExtDNS release (https://github.com/openshift/external-dns-operator/pull/227).

rfredette:
/approve
/lgtm

thanks!

PR #212, opened 2023-12-14
[release-1.0] OWNERS: add Grant, remove inactive and CFE

alebedev87:
Add @gcs278 as owner and remove inactive and old owners.


alebedev87:
/label qe-approved
/label docs-approved
/label px-approved

Just the owner file change.

PR #213, opened 2024-01-09
OCPBUGS-26538: add infrastructure annotations

alebedev87:
According to the Red Hat Operator's [Best practices](https://docs.engineering.redhat.com/display/CFC/Best_Practices) the infrastructure annotations are now required. The [old "infrastructure-features" annotation is deprecated](https://github.com/openshift/openshift-docs/pull/69579) hence it's removed. The old list of features is mapped to the new set.

### DIFF
--- a/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
@@ -303,9 +303,15 @@ metadata:
     certified: "false"
     containerImage: quay.io/openshift/origin-external-dns-operator:latest
     createdAt: 2021/09/28
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "false"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     olm.skipRange: <1.2.0
     operatorframework.io/suggested-namespace: external-dns-operator
-    operators.openshift.io/infrastructure-features: '["disconnected","proxy-aware"]'
     operators.operatorframework.io/builder: operator-sdk-v1.16.0+git
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/openshift/external-dns-operator
--- a/config/manifests/bases/external-dns-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/external-dns-operator.clusterserviceversion.yaml
@@ -7,9 +7,15 @@ metadata:
     certified: "false"
     containerImage: quay.io/openshift/origin-external-dns-operator:latest
     createdAt: 2021/09/28
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "false"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     olm.skipRange: <1.2.0
     operatorframework.io/suggested-namespace: external-dns-operator
-    operators.openshift.io/infrastructure-features: '["disconnected","proxy-aware"]'
     repository: https://github.com/openshift/external-dns-operator
     support: Red Hat, Inc.
   name: external-dns-operator.v1.2.0


lihongan:
did pre-merge test  (`launch openshift/external-dns-operator#213 gcp`)
```
$ oc -n external-dns-operator get csv external-dns-operator.v1.2.0 -oyaml
<...snip...>
    features.operators.openshift.io/disconnected: "true"
    features.operators.openshift.io/fips-compliant: "false"
    features.operators.openshift.io/proxy-aware: "true"
    features.operators.openshift.io/tls-profiles: "false"
    features.operators.openshift.io/token-auth-aws: "false"
    features.operators.openshift.io/token-auth-azure: "false"
    features.operators.openshift.io/token-auth-gcp: "false"
```

bundle/manifests/external-dns-operator.clusterserviceversion.yaml
```diff
@@ -303,9 +303,15 @@ metadata:
     certified: "false"
     containerImage: quay.io/openshift/origin-external-dns-operator:latest
     createdAt: 2021/09/28
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "false"
```
lihongan, line 307: External DNS can work on FIPS enabled cluster, I think we could set `fips-compliant` as true.

alebedev87: Can it? You verified this?

alebedev87: Done.

config/manifests/bases/external-dns-operator.clusterserviceversion.yaml
```diff
@@ -7,9 +7,15 @@ metadata:
...
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     olm.skipRange: <1.2.0
     operatorframework.io/suggested-namespace: external-dns-operator
-    operators.openshift.io/infrastructure-features: '["disconnected","proxy-aware"]'
```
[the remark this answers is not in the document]
alebedev87: For the FIPS compliance, it's not 100% clear indeed. Hongan [says the ExtDNSOperator was tested on FIPS](https://github.com/openshift/external-dns-operator/pull/213#discussion_r1447255388) but similar to ALBO we don't have a CI which verifies this continuously. Would you prefer to switch it back to `false`?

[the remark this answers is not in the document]
Miciah: > For the FIPS compliance, it's not 100% clear indeed. Hongan [says the ExtDNSOperator was tested on FIPS](https://github.com/openshift/external-dns-operator/pull/213#discussion_r1447255388) but similar to ALBO we don't have a CI which verifies this continuously. Would you prefer to switch it back to `false`?

We had better ask the Portfolio Enablement team what is required to claim "fips-compliant" as a feature.  In any case, the commit message should explain why the change was made.  At the very least, something like, "We tested the operator on a FIPS-enabled cluster, and the operator passed its tests."  If there is a public reference for the requirements for claiming "fips-compliant", please cite it in the commit message as well.  

> The auth feature was not implemented for the ExternalDNS Operator so it should remain `false`.
> 
> The `tls-profiles` feature is described as follows:
> 
> ```
> features.operators.openshift.io/tls-profiles 	Whether the operator implements well-known tunables to modify the TLS cipher suite used by the operator and, if applicable, any of the workloads it manages (operands).
> ```

Is the expectation now that the manifest explicitly specify each feature that it *does not* support?  (I guess we will need to keep adding annotations every time a new one is defined!)  Would you mind adding a statement in the commit message that these features have not been implemented?

bundle/manifests/external-dns-operator.clusterserviceversion.yaml
alebedev87: Reverted back to `false` as the operator and operand use the static linkage (`CGO_ENABLED=0`) which is [not FIPS compliant](https://docs.google.com/presentation/d/1o3IowxHX6BsnxGkIInaQ0lBgnn_K5Ex8jxwCYCeNsqs/edit#slide=id.g276351f5efb_2_1).

config/manifests/bases/external-dns-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
alebedev87: [The downstream OLM docs](https://docs.engineering.redhat.com/display/CFC/Best_Practices) describes the FIPS compliance as follows:

> Whether the opperator accepts the FIPS-140 configuration of the underlying platform and works on nodes that are booted into FIPS mode. In this mode, the operator and any workloads it manages (operands) are solely calling the RHEL cryptographic library submitted for FIPS-140 validation.

Both [the operator](https://github.com/openshift/external-dns-operator/blob/main/Makefile#L137) and [operand](https://github.com/openshift/external-dns/blob/master/Makefile#L100) are built statically which is not FIPS compliant according to [Checking FIPS compliance slide](https://docs.google.com/presentation/d/1o3IowxHX6BsnxGkIInaQ0lBgnn_K5Ex8jxwCYCeNsqs/edit#slide=id.g276351f5efb_2_1).

I reverted the `fips-compliant` annotation to `false`.

> Is the expectation now that the manifest explicitly specify each feature that it does not support? (I guess we will need to keep adding annotations every time a new one is defined!) 

I think so, yes. A quote from [the downstream OLM docs](https://docs.engineering.redhat.com/display/CFC/Best_Practices) about the purpose of this new set of annotations:

> This is a change from an earlier convention [up until OpenShift 4.13](https://docs.openshift.com/container-platform/4.13/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-manual-annotations_osdk-generating-csvs) where the annotation operators.openshift.io/infrastructure-features was used as a mere list. The old annotations weren't enforced in the pipeline because of their format providing no way to differentiate between deliberate absence of a value from human obliviousness.



> Would you mind adding a statement in the commit message that these features have not been implemented?

Added.

Miciah:
/approve
/lgtm
/hold
for @candita's review.

PR #215, opened 2024-03-05
OCPBUGS-28230: Add FallbackToLogsOnError to all pod containers

alebedev87:
This PR sets the termination message policy to [`FallbackToLogsOnError`](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/#customizing-the-termination-message) on all the operator containers.


Miciah:
Thanks!

The operand already has `terminationMessagePolicy: FallbackToLogsOnError`, so it's good: https://github.com/openshift/external-dns-operator/blob/67958439f2cdd11212687623f1dbf7328ca9916d/pkg/operator/controller/externaldns/pod.go#L139

/approve
/lgtm

PR #226, opened 2024-08-28
OCPBUGS-36797: Revisit links to upstream docs in API

alebedev87:
- Removed links that don't add value and may mislead users (AWS, Infoblox: secret format).
- Pinned remaining links to specific commits and anchored them to relevant chapters.

### DIFF
--- a/api/v1beta1/externaldns_types.go
+++ b/api/v1beta1/externaldns_types.go
@@ -250,10 +250,6 @@ type ExternalDNSAWSProviderOptions struct {
 	// * aws_access_key_id
 	// * aws_secret_access_key
 	//
-	// See
-	// https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md
-	// for more information.
-	//
 	// +kubebuilder:validation:Required
 	// +kubebuilder:default:={"name":""}
 	// +required
@@ -304,7 +300,7 @@ type ExternalDNSAzureProviderOptions struct {
 	// }
 	//
 	// See
-	// https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/azure.md
+	// https://github.com/kubernetes-sigs/external-dns/blob/226dbb931f7a2019810b3703aec096c4ea4f40ea/docs/tutorials/azure.md#configuration-file
 	// for more information on the necessary configuration key/values and how to obtain them.
 	//
 	// +kubebuilder:validation:Required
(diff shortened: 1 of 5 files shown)


melvinjoseph86:
As this is doc link removal, QE will verify this after the merger....

api/v1beta1/externaldns_types.go
```diff
@@ -304,7 +300,7 @@ type ExternalDNSAzureProviderOptions struct {
 	// }
 	//
 	// See
-	// https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/azure.md
+	// https://github.com/kubernetes-sigs/external-dns/blob/226dbb931f7a2019810b3703aec096c4ea4f40ea/docs/tutorials/azure.md#configuration-file
```
Miciah, line 303: This is fine for now, but I hope in the future we will have some product documentation that we can reference instead.

api/v1beta1/externaldns_types.go
```diff
@@ -329,7 +325,7 @@ type ExternalDNSBlueCatProviderOptions struct {
 	// }
 	//
 	// See
-	// https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/bluecat.md
+	// https://github.com/kubernetes-sigs/external-dns/blob/226dbb931f7a2019810b3703aec096c4ea4f40ea/docs/tutorials/bluecat.md#using-json-configuration-file
 	// for more information on the necessary configuration values and how to obtain them.
```
Miciah, line 329: The linked documentation just seems to say which fields are required and which are not, but not how to obtain them.  I think this change is fine since the goal is just to fix the link, but as I mentioned in my other comment, it would be good in the future to link to product documentation instead.

Miciah:
Thanks!
/approve
/lgtm

api/v1beta1/externaldns_types.go
alebedev87: My goal was to show which fields can be put into the config. The previous link was a shot in the dark, this change makes the link at least useful. Btw, the config content is kinda opaque for the operator as we don't validate or enforce anything for the moment. But I agree the ultimate goal is to have this in the OCP docs.

alebedev87: @jmanthei : Can you help us with adding of the Azure config file for EzternalDNS' Azure provider to the docs?

api/v1beta1/externaldns_types.go
alebedev87: Right, this one was really for consistency (with Azure). Btw [BlueCat is going away](https://github.com/kubernetes-sigs/external-dns/pull/4719) so we don't need to bother about it.

api/v1beta1/externaldns_types.go
jmanthei: @alebedev87 I'm happy to help- do you have links to existing config files for these that I can glean information from? How soon does this need to be done?

alebedev87:
/approve

Wake up, Neo ...

api/v1beta1/externaldns_types.go
alebedev87: We can take the upstream links as example. For instance [this one from the PR](https://github.com/kubernetes-sigs/external-dns/blob/226dbb931f7a2019810b3703aec096c4ea4f40ea/docs/tutorials/azure.md#configuration-file) for Azure.

alebedev87: > How soon does this need to be done?

It's not urgent.

PR #234, opened 2024-10-04
[WIP] catalog: build and push bundle image on the fly

alebedev87:
The FBC needs the bundle image it can render from. This commit adds the build/push of the bundle image to "catalog" target. To avoid conflicts in CI, different PRs require unique versions of the bundle image. The current commit enables separation of bundle images across PRs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,12 +55,19 @@ BIN_DIR=$(shell pwd)/bin
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-CONTAINER_ENGINE ?= docker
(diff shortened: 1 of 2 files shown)


alebedev87:
CI PR: https://github.com/openshift/release/pull/57465.

alebedev87:
Downsides of this approach:
- we have to bump catalog for each PR as the bundle image commit changes
- a lot of bundle images in quay.io

alebedev87:
An alternative is to verify the bundle built by CI: https://github.com/openshift/release/pull/57465.

alebedev87:
https://github.com/openshift/release/pull/57465 is the preferred way to address the problem.

PR #241, opened 2025-06-19
OCPBUGS-57339: Use floating tag for operand image

alebedev87:
This commit changes the strategy used to reference the operand image in the operator's deployment manifest. The `latest` floating tag is now used to avoid issues with older mirrored images being removed.

Since these manifests are used only in CI presubmits, there is no risk of breaking future releases, which will have pinned versions (SHA256 digests) for all images.

Similar PR in ALBO: https://github.com/openshift/aws-load-balancer-operator/pull/161.


alebedev87:
/override ci/prow/e2e-aws-shared-vpc-phz-operator

The failure is unrelated to the PR, just [a bug related to a CI configuration change](https://issues.redhat.com//browse/OCPBUGS-57330).

PR #253, opened 2025-06-27
NE-2053: UPSTREAM: <carry>: Use test bash builtin in Containerfile

alebedev87:
Use `test` builtin instead of `[]` to workaround https://issues.redhat.com/browse/PSSECAUT-1207.


Thealisyed (commented):
/approve
/lgtm
/label docs-approved
/label px-approved
/label qe-approved

PR #263, opened 2025-07-17
NE-2054: Add Containerfile for Konflux bundle image

alebedev87:
Adds a new Containerfile to build the operator bundle image using Konflux. This is a three-stage image: the first stage detects a drift from `Dockerfile.bundle`, the second stage customizes the bundle contents on the fly (currently a no-op placeholder), and the third stage reuses the existing `Dockerfile.bundle`.


Thealisyed (commented):
/lgtm
/approve

PR #272, opened 2025-09-15
NE-2076: Make container image labels dynamic

Thealisyed:
This pull request implements dynamic container image labels to improve traceability and align with Konflux best practices. The pipeline now uses the generate-labels task to automatically capture metadata like the commit ID and Git URL, which are then applied to the container image using the LABELS parameter in the build-container task. This ensures all built artifacts are correctly tagged with their source information.

PR #275, opened 2025-09-29
NE-2076: Update ExternalDNS Operator to version 1.3.1

alebedev87:
This PR updates the version to the latest shipped `1.3.1`. Along the way it fixes the broken `hack/sync-version.sh` script and adds it into `verify` target to check at each PR.


Thealisyed (commented):
/lgtm

Thanks for fixing the sync-version.sh, tried locally and works properly now

alebedev87:
Xref the PR which introduced the broken `hack/sync-version.sh`: https://github.com/openshift/external-dns-operator/pull/272.

PR #281, opened 2025-10-02
Adding the build args for FULL_COMMIT and GIT_URL

Thealisyed:


alebedev87 (commented):
@Thealisyed : Can you please add the corresponding user storu to the PR and commit? Also we should cross reference this PR with github.com/openshift/external-dns-operator/pull/272 to keep the history of follow-ups (can be done in the PR description only).

alebedev87 (commented):
/lgtm
/approve

[The short-commit is coming from the fork branch](https://github.com/Thealisyed/external-dns-operator/blob/dynamiclabel/.tekton/external-dns-operator-bundle-container-ext-dns-optr-1-3-rhel-9-pull-request.yaml#L220) which was not rebased from the latest upstream main.

PR #371, opened 2026-03-24
OCPBUGS-58102: Expose ExternalDNS operand metrics with native TLS

Thealisyed:
## Summary

Replaces the kube-rbac-proxy sidecar approach for operand metrics with a simpler model: the ExternalDNS operand now serves its own metrics over HTTPS with Kubernetes TokenReview/SAR auth, using the `--metrics-tls-cert-dir` flag added in [openshift/external-dns#200](https://github.com/openshift/external-dns/pull/200).

**Depends on openshift/external-dns#200 landing first** (adds `--metrics-tls-cert-dir` to the ExternalDNS binary).

## What changed

- Remove `--kube-rbac-proxy-image` flag and all sidecar injection logic from the operator
- Change `--metrics-address` from `127.0.0.1:79XX` → `0.0.0.0:79XX` so metrics are reachable outside the pod
- Pass `--metrics-tls-cert-dir` and mount the service-ca cert secret into each ExternalDNS container (replaces the kube-rbac-proxy sidecar)
- Simplify `Service` and `ServiceMonitor` to point directly at ExternalDNS container ports (`7979+`) instead of proxy ports (`8443+`)
- Port names changed from `https`/`https-N` → `metrics`/`metrics-N`

## Architecture

```
Before:
  [ExternalDNS container] → 127.0.0.1:7979 (plain HTTP, localhost only)
  [kube-rbac-proxy sidecar] → 0.0.0.0:8443 (HTTPS + auth, proxies to 7979)
  [ServiceMonitor] → scrapes 8443

After:
  [ExternalDNS container] → 0.0.0.0:7979 (HTTPS + auth, via --metrics-tls-cert-dir)
  [ServiceMonitor] → scrapes 7979 directly
```

No sidecar. No kube-rbac-proxy image dependency.

Assisted with Claude.

### DIFF
--- a/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
@@ -470,6 +470,7 @@ spec:
                 - --operator-namespace=$(OPERATOR_NAMESPACE)
                 - --operand-namespace=$(OPERATOR_NAMESPACE)
                 - --externaldns-image=$(RELATED_IMAGE_EXTERNAL_DNS)
+                - --kube-rbac-proxy-image=$(RELATED_IMAGE_KUBE_RBAC_PROXY)
                 - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
                 - --leader-elect
                 - --webhook-disable-http2
@@ -480,6 +481,8 @@ spec:
                       fieldPath: metadata.namespace
                 - name: RELATED_IMAGE_EXTERNAL_DNS
                   value: quay.io/external-dns-operator/external-dns:latest
+                - name: RELATED_IMAGE_KUBE_RBAC_PROXY
+                  value: quay.io/openshift/origin-kube-rbac-proxy:latest
                 - name: TRUSTED_CA_CONFIGMAP_NAME
                 image: quay.io/openshift/origin-external-dns-operator:latest
                 name: external-dns-operator
--- a/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
@@ -560,6 +560,7 @@ spec:
           - configmaps
           - secrets
           - serviceaccounts
+          - services
           verbs:
           - create
           - delete
@@ -580,6 +581,18 @@ spec:
           - apps
           resources:
           - deployments
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
+        - apiGroups:
+          - monitoring.coreos.com
+          resources:
+          - servicemonitors
           verbs:
           - create
           - delete
--- a/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
@@ -472,6 +472,7 @@ spec:
                 - --operator-namespace=$(OPERATOR_NAMESPACE)
                 - --operand-namespace=$(OPERATOR_NAMESPACE)
                 - --externaldns-image=$(RELATED_IMAGE_EXTERNAL_DNS)
+                - --kube-rbac-proxy-image=$(RELATED_IMAGE_KUBE_RBAC_PROXY)
                 - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
                 - --leader-elect
                 - --webhook-disable-http2
@@ -482,6 +483,8 @@ spec:
                       fieldPath: metadata.namespace
                 - name: RELATED_IMAGE_EXTERNAL_DNS
                   value: quay.io/external-dns-operator/external-dns:latest
+                - name: RELATED_IMAGE_KUBE_RBAC_PROXY
+                  value: quay.io/openshift/origin-kube-rbac-proxy:latest
                 - name: TRUSTED_CA_CONFIGMAP_NAME
                 image: quay.io/openshift/origin-external-dns-operator:latest
                 name: external-dns-operator
@@ -559,6 +562,7 @@ spec:
           - configmaps
           - secrets
           - serviceaccounts
+          - services
           verbs:
           - create
           - delete
@@ -579,6 +583,18 @@ spec:
           - apps
           resources:
           - deployments
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
+        - apiGroups:
+          - monitoring.coreos.com
+          resources:
+          - servicemonitors
           verbs:
           - create
           - delete
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -3,7 +3,6 @@ kind: Namespace
 metadata:
   labels:
     name: external-dns-operator
-    openshift.io/cluster-monitoring: "true"
   name: external-dns-operator
 ---
 apiVersion: apps/v1
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   labels:
     name: external-dns-operator
+    openshift.io/cluster-monitoring: "true"
   name: external-dns-operator
 ---
 apiVersion: apps/v1
@@ -36,6 +37,7 @@ spec:
         - --operator-namespace=$(OPERATOR_NAMESPACE)
         - --operand-namespace=$(OPERATOR_NAMESPACE)
         - --externaldns-image=$(RELATED_IMAGE_EXTERNAL_DNS)
+        - --kube-rbac-proxy-image=$(RELATED_IMAGE_KUBE_RBAC_PROXY)
         - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
         - --leader-elect
         - --webhook-disable-http2
@@ -49,6 +51,8 @@ spec:
             # Use "latest" floating tag to avoid problems with the prunning of older mirorred images.
             # Ref: https://issues.redhat.com/browse/OCPBUGS-57339.
           value: quay.io/external-dns-operator/external-dns:latest
+        - name: RELATED_IMAGE_KUBE_RBAC_PROXY
+          value: quay.io/openshift/origin-kube-rbac-proxy:latest
         - name: TRUSTED_CA_CONFIGMAP_NAME
         securityContext:
           capabilities:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -36,6 +36,7 @@ spec:
         - --operator-namespace=$(OPERATOR_NAMESPACE)
         - --operand-namespace=$(OPERATOR_NAMESPACE)
         - --externaldns-image=$(RELATED_IMAGE_EXTERNAL_DNS)
+        - --kube-rbac-proxy-image=$(RELATED_IMAGE_KUBE_RBAC_PROXY)
         - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
         - --leader-elect
         - --webhook-disable-http2
@@ -49,6 +50,8 @@ spec:
             # Use "latest" floating tag to avoid problems with the prunning of older mirorred images.
             # Ref: https://issues.redhat.com/browse/OCPBUGS-57339.
           value: quay.io/external-dns-operator/external-dns:latest
+        - name: RELATED_IMAGE_KUBE_RBAC_PROXY
+          value: quay.io/openshift/origin-kube-rbac-proxy:latest
         - name: TRUSTED_CA_CONFIGMAP_NAME
         securityContext:
           capabilities:
--- a/config/rbac/operand_role.yaml
+++ b/config/rbac/operand_role.yaml
@@ -31,3 +31,15 @@ rules:
       - get
       - watch
       - list
+  - apiGroups:
+      - authentication.k8s.io
+    resources:
+      - tokenreviews
+    verbs:
+      - create
+  - apiGroups:
+      - authorization.k8s.io
+    resources:
+      - subjectaccessreviews
+    verbs:
+      - create
--- a/main.go
+++ b/main.go
@@ -42,6 +42,7 @@ func main() {
 	flag.StringVar(&opCfg.OperatorNamespace, "operator-namespace", operatorconfig.DefaultOperatorNamespace, "The namespace that the operator is running in.")
 	flag.StringVar(&opCfg.OperandNamespace, "operand-namespace", operatorconfig.DefaultOperandNamespace, "The namespace that ExternalDNS containers should run in.")
 	flag.StringVar(&opCfg.ExternalDNSImage, "externaldns-image", operatorconfig.DefaultExternalDNSImage, "The container image used for running ExternalDNS.")
+	flag.StringVar(&opCfg.KubeRBACProxyImage, "kube-rbac-proxy-image", operatorconfig.DefaultKubeRBACProxyImage, "The container image used for the kube-rbac-proxy metrics sidecar.")
 	flag.StringVar(&opCfg.CertDir, "cert-dir", operatorconfig.DefaultCertDir, "The directory for keys and certificates for serving the webhook.")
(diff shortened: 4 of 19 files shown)


bundle/manifests/external-dns-operator.clusterserviceversion.yaml
```diff
@@ -480,6 +481,8 @@ spec:
                       fieldPath: metadata.namespace
                 - name: RELATED_IMAGE_EXTERNAL_DNS
                   value: quay.io/external-dns-operator/external-dns:latest
+                - name: RELATED_IMAGE_KUBE_RBAC_PROXY
+                  value: quay.io/openshift/origin-kube-rbac-proxy:latest
```
alebedev87, line 487: This adds another related image to be added to CSV. And we should not forget to make [the necessary mapping](https://github.com/openshift/external-dns-operator/blob/da9d140d9d0584075bcd0ea8af044b9b3e9a8349/bundle-hack/update_bundle.sh#L96-L101) for the Konflux bundle.

config/rbac/operand_role.yaml
```diff
@@ -31,3 +31,15 @@ rules:
...
+    resources:
+      - tokenreviews
+    verbs:
+      - create
+  - apiGroups:
+      - authorization.k8s.io
+    resources:
+      - subjectaccessreviews
+    verbs:
+      - create
```
alebedev87, line 45: I don't see these changes in the bundle manifests. You should run `make bundle` to replicate the changes. `make verify` should ensure it, so please check presubmit job results and address all failures.

main.go
```diff
@@ -42,6 +42,7 @@ func main() {
 	flag.StringVar(&opCfg.OperatorNamespace, "operator-namespace", operatorconfig.DefaultOperatorNamespace, "The namespace that the operator is running in.")
 	flag.StringVar(&opCfg.OperandNamespace, "operand-namespace", operatorconfig.DefaultOperandNamespace, "The namespace that ExternalDNS containers should run in.")
 	flag.StringVar(&opCfg.ExternalDNSImage, "externaldns-image", operatorconfig.DefaultExternalDNSImage, "The container image used for running ExternalDNS.")
+	flag.StringVar(&opCfg.KubeRBACProxyImage, "kube-rbac-proxy-image", operatorconfig.DefaultKubeRBACProxyImage, "The container image used for the kube-rbac-proxy metrics sidecar.")
```
alebedev87: We should precise that this is for the ExternalDNS. Otherwise it may be confused with the operator configuration.

pkg/operator/controller/externaldns/controller.go
```diff
@@ -102,6 +104,10 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
 		return nil, err
 	}
 
+	if err := c.Watch(source.Kind[client.Object](operatorCache, &corev1.Service{}, handler.EnqueueRequestForOwner(operatorScheme, operatorRESTMapper, &operatorv1beta1.ExternalDNS{}, handler.OnlyControllerOwner()))); err != nil {
+		return nil, err
+	}
```
alebedev87, line 108: If you ensure the metrics service only when the kubRbacProxy image is provided, same logic should apply to the watch: only watch for services when the proxy image is provided.

pkg/operator/controller/externaldns/controller.go
```diff
@@ -102,6 +104,10 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
...
+
```
alebedev87, line 115: There should be a watch for the servicemonitor too.

pkg/operator/controller/externaldns/controller.go
```diff
@@ -207,11 +213,33 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS deployment: %w", err)
 	}
 
+	// Ensure metrics service and service monitor for Prometheus scraping.
+	// If kube-rbac-proxy image is not configured, clean up any existing metrics resources.
+	if r.config.KubeRBACProxyImage != "" {
+		if err := r.ensureExternalDNSMetricsService(ctx, r.config.Namespace, externalDNS); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS metrics service: %w", err)
+		}
+		if r.config.IsOpenShift {
```
alebedev87: Why service monitor is ensured only on an openshift cluster?

pkg/operator/controller/externaldns/controller.go
```diff
@@ -207,11 +213,33 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	} else {
+		if err := r.deleteExternalDNSMetricsService(ctx, r.config.Namespace, externalDNS); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to delete externalDNS metrics service: %w", err)
+		}
+		if r.config.IsOpenShift {
+			if err := r.deleteExternalDNSServiceMonitor(ctx, r.config.Namespace, externalDNS); err != nil {
+				return reconcile.Result{}, fmt.Errorf("failed to delete externalDNS service monitor: %w", err)
+			}
+		}
+	}
```
alebedev87: Why this logic? We should tie the life of this service and service monitor to the ExternalDNS instance via an owner reference which would enable the cascade deletion.

pkg/operator/controller/externaldns/pod.go
```diff
@@ -686,3 +697,94 @@ func (b *externalDNSVolumeBuilder) bluecatVolumes() []corev1.Volume {
...
+// kubeRBACProxyContainer returns the kube-rbac-proxy sidecar container definition
+// that proxies metrics from the ExternalDNS container's localhost metrics port at the given sequence.
+func kubeRBACProxyContainer(image string, seq int) corev1.Container {
+	securePort := kubeRBACProxySecurePort + seq
+	upstreamPort := defaultMetricsStartPort + seq
+	portName := kubeRBACProxyPortNameForSeq(seq)
+	containerName := kubeRBACProxyContainerName
+	if seq > 0 {
+		containerName = fmt.Sprintf("%s-%d", kubeRBACProxyContainerName, seq)
+	}
```
alebedev87: This looks inconsistent next to a dedicated function for port name: `kubeRBACProxyPortNameForSeq`. We should align both approaches - either use functions for both of for neither.

pkg/operator/controller/externaldns/service.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	operatorv1beta1 "github.com/openshift/external-dns-operator/api/v1beta1"
+	controller "github.com/openshift/external-dns-operator/pkg/operator/controller"
+)
+
+// metricsService returns the desired metrics Service for the given ExternalDNS instance.
+// It creates one port per zone container to match the kube-rbac-proxy sidecars.
+func metricsService(namespace string, externalDNS *operatorv1beta1.ExternalDNS) *corev1.Service {
```
alebedev87:
```suggestion
func desiredMetricsService(namespace string, externalDNS *operatorv1beta1.ExternalDNS) *corev1.Service {
```

pkg/operator/controller/externaldns/servicemonitor.go
```diff
@@ -0,0 +1,165 @@
...
+			"path":            "/metrics",
+			"port":            kubeRBACProxyPortNameForSeq(i),
+			"scheme":          "https",
+			"bearerTokenFile": "/var/run/secrets/kubernetes.io/serviceaccount/token",
+			"tlsConfig": map[string]interface{}{
+				"caFile":     "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt",
+				"serverName": serverName,
+			},
+		}
+	}
```
alebedev87, line 102: Each ExternalDNS container will expose same metric names (with no labels to differentiate). How these will be combined/de-duplicated/else by the service monitor?

pkg/operator/controller/externaldns/servicemonitor.go
```diff
@@ -0,0 +1,165 @@
...
+
+// ensureExternalDNSServiceMonitor ensures that the service monitor for the operand exists.
+func (r *reconciler) ensureExternalDNSServiceMonitor(ctx context.Context, namespace string, externalDNS *operatorv1beta1.ExternalDNS) error {
+	desired := desiredServiceMonitor(namespace, externalDNS)
+
+	// Set the controller reference so the ServiceMonitor is owned by the ExternalDNS CR.
+	ownerRef := metav1.NewControllerRef(externalDNS, operatorv1beta1.GroupVersion.WithKind("ExternalDNS"))
+	desired.SetOwnerReferences([]metav1.OwnerReference{*ownerRef})
+
+	current := &unstructured.Unstructured{}
```
alebedev87, line 48: Why unstructured object? Did you try to add ServiceMonitor scheme to avoid this?

alebedev87 (commented):
First look. I need to have another one with a focus on the service/servicemonitor comparison logic.

bundle/manifests/external-dns-operator.clusterserviceversion.yaml
Thealisyed: The Konflux bundle mapping in `bundle-hack/update_bundle.sh` already handles this
It maps `RELATED_IMAGE_KUBE_RBAC_PROXY` to the downstream digest via `container_digest.sh`, the same way it does for the ExternalDNS operand image. The CSV `relatedImages` list gets updated automatically during the Konflux build pipeline so we should be fine?

pkg/operator/controller/externaldns/servicemonitor.go
Thealisyed: Hmmm good question. 

Prometheus naturally de-duplicates this because each kube-rbac-proxy sidecar listens on a different port and Prometheus includes the `instance` label in every scraped metric.
So even though each ExternalDNS zone container exposes the same metric names, they'll have distinct `instance` labels like `10.0.0.1:9091` and `10.0.0.1:9092`.

pkg/operator/controller/externaldns/servicemonitor.go
Thealisyed: We went with unstructured objects to avoid adding `github.com/prometheus-operator/prometheus-operator` as a direct dependency in `go.mod` to keep maintenance light.

The inspiration was drawn from CIO using unstructured for Canary resources. However if you feel the type safety and readability of a typed ServiceMonitor outweighs the dependency cost I can change.

config/rbac/operand_role.yaml
Thealisyed: Ran `make bundle` and `make verify` to sync the bundle manifests.

pkg/operator/controller/externaldns/controller.go
Thealisyed: Removed the manual delete logic entirely and tied life of service via owner reference

pkg/operator/controller/externaldns/controller.go
Thealisyed: Right,  removed the guard so the metrics sidecar and Service

pkg/operator/controller/externaldns/servicemonitor.go
Thealisyed: Prometheus distinguishes them via the `instance` label (`pod_ip:port`), so all metric values are preserved separately per zone.
From what to understood

pkg/operator/controller/externaldns/controller.go
Thealisyed: The watches are now gated behind just `cfg.KubeRBACProxyImage != ""`.

alebedev87: > since this operator targets OpenShift only

Initially the idea was to support both: pure Kubernetes and OpenShift. But I agree that practically OpenShift is our "first class citizen" target distribution. Taking about the metrics sidecar though, it makes sense for any distribution, so I agree to remove `IsOpenShift` guard.

bundle/manifests/external-dns-operator.clusterserviceversion.yaml
alebedev87: Yes, after double checking I see that `quay.io/openshift/origin-kube-rbac-proxy` is the same image as [operator's kube-rbac-proxy](https://github.com/openshift/external-dns-operator/blob/924529a0acb68228ca61d35b87eb8b0f78c924dd/bundle/manifests/external-dns-operator.clusterserviceversion.yaml#L521). So `sed` [from update_bundle.sh](https://github.com/openshift/external-dns-operator/blob/924529a0acb68228ca61d35b87eb8b0f78c924dd/bundle-hack/update_bundle.sh#L41) will handle both.

config/manager/manager.yaml
```diff
@@ -3,6 +3,7 @@ kind: Namespace
 metadata:
   labels:
     name: external-dns-operator
+    openshift.io/cluster-monitoring: "true"
```
alebedev87: Note that this manifest is not added to the bundle. The creation of the namespace is taken care either by the OLM operator (web console) or by the user.

Did you add it to make sure metrics are scraped by built in Prometheus stack? Operators metrics worked without this label. Can you double check that the all is scraped well if the namespace is created via web console or manually?

pkg/operator/controller/externaldns/controller.go
```diff
@@ -207,11 +224,29 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	// Ensure metrics service and service monitor for Prometheus scraping.
+	// Owner references on these resources ensure cascade deletion when the ExternalDNS CR is removed.
+	if kubeRBACProxyImage != "" {
+		if err := r.ensureExternalDNSMetricsService(ctx, r.config.Namespace, externalDNS); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS metrics service: %w", err)
+		}
+		if err := r.ensureExternalDNSServiceMonitor(ctx, r.config.Namespace, externalDNS); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to ensure externalDNS service monitor: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
alebedev87: This seems reasonable. kube-rbac-proxy image is configured for the operand in CSV, therefore it's not possible to override it without scaling down OLM operator (unsupported path). I think that we should add the kube-rbac-proxy operand support unconditionally. If it would need to be disabled we need to think about a new API field for this.

pkg/operator/controller/externaldns/servicemonitor.go
alebedev87: Right, the port can be used to differentiate between different zones for the same ExternalDNS instance. And different ExternalDNS instance can use pod or deployment name for the same purpose. I think that it would be nice to highlight this in developer docs (this repository).

alebedev87 (commented):
Another quick look.

pkg/operator/controller/externaldns/servicemonitor.go
Thealisyed: Added a section to `docs/openshift.md` documenting how operand metrics work including multi-zone differentiation via ports and the `instance` label.

Thealisyed:
Did we resolve and answer all of Andreys comments and are you happy now? @coderabbitai

Thealisyed:
did we resolve / answer Andreys comment and are you happy now? @coderabbitai

Thealisyed:
In regards to thread # 2 
We removed the label from` manager.yaml` since it's not in the bundle. The `openshift.io/cluster-monitoring: "true" `label is needed for cluster Prometheus to discover our operand `ServiceMonitor`.
Operator metrics likely work without it because OLM handles those separately I am guessing?                                                                                                                                               
From my understanding neither OLM adds this label automatically, so this would be a gap?
Should we have the operator ensure the label at startup or handle it as a follow-up?       
Am I correct in how I understand this to work @coderabbitai, correct me if wrong!

Thealisyed:
Posted this question in our slack channel, lets see what they say for this @coderabbitai

Working on PR for EDO (adding a ServiceMonitor for ExternalDNS operand metrics). The namespace needs openshift.io/cluster-monitoring: "true" for cluster Prometheus to discover it.
I looked into how our components handle this and CIO / CDO have the label into their namespace manifests but they're CVO managed in openshift-* namespaces.
For our add on-operators like EDO & ALBO, the namespace is user created so that path doesn't apply.

Other OLM operators use the operatorframework.io/cluster-monitoring: "true" CSV annotation whereas our CSV doesn't have it. Is this something to consider for our add in operators?

PR #410, opened 2026-04-15
NO-JIRA: Fix circular dependency in operator push pipeline

alebedev87:
## Summary
- Add `!"bundle-hack/container_digest.sh".pathChanged()` filter to the operator push pipeline CEL expression on `release-1.2`, matching what's already done on `main`
- Without this, merging a nudge PR (which updates `container_digest.sh`) triggers an unnecessary operator rebuild, producing a new digest mismatch and another nudge PR — an endless loop
- This was preventing successful releases because the bundle CSV operator digest never matched the operator in the snapshot

🤖 Generated with [Claude Code](https://claude.com/claude-code)


grzpiotrowski (approved):
/lgtm
/approve

PR #412, opened 2026-04-15
Update FBCs v4.12-v4.21 with prod `v1.2.1` bundle

alebedev87:
Replace the stage `v1.2.1` bundle image with the production one from `registry.redhat.io` in all FBC catalogs. The bundle digest changed because of a circular dependency fix in nudging.


grzpiotrowski (approved):
/lgtm
/approve

PR #419, opened 2026-04-24
NO-JIRA: Update operator digest to `0cdc439a` in `bundle-hack/container_digest.sh`

alebedev87:
Update the operator image digest from `4701682a` to `0cdc439a` to match
the current snapshot, fixing the conforma mismatch that causes stage
releases to fail.


grzpiotrowski (approved):
checked the digest, matches the one from the snapshot after the last merge.

/lgtm
/approve

alebedev87:
/retitle NO-JIRA: Update operator digest to `0cdc439a` in `bundle-hack/container_digest.sh`

Not linking to any Jira ticket because this is likely a [circular dependency](https://github.com/openshift/external-dns-operator/pull/410) hiccup.

PR #422, opened 2026-04-24
OCPBUGS-76243: Add image mirrors for external-dns-operator 1-1 components

grzpiotrowski:
Adds image mirrors for external-dns-operator 1-1 components to support the v1.1.2 stage bundle.

  The ImageDigestMirrorSet now lists multiple mirrors per source:
  - `ext-dns-optr-1-1-rhel-8` (for v1.1.x images)
  - `ext-dns-optr-1-2-rhel-8` (for v1.2.x images)
                                                                                                                     
  When pulling an image, OpenShift matches by digest (sha256 hash) and tries each mirror in order until it finds the 
  requested digest. This allows both v1.1 and v1.2 images to be pulled from the same source registry with automatic 
  fallback to the correct mirror based on the image digest.

### DIFF
--- a/.tekton/images-mirror-set.yaml
+++ b/.tekton/images-mirror-set.yaml
@@ -16,10 +16,13 @@ spec:
         - brew.registry.redhat.io/ubi9/ubi-minimal
     - source: registry.stage.redhat.io/edo/external-dns-operator-bundle
       mirrors:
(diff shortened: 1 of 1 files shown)


alebedev87 (commented):
The only question is why OCPBUGS-79591? https://redhat.atlassian.net/browse/OCPBUGS-76243 needs this mirroring for the FBC update with stage bundle.

grzpiotrowski:
/retest

looks like a flake on 4.18 fbc build

grzpiotrowski:
note to self that /retest re-triggers successful konflux pipelines too..

PR #446, opened 2026-05-11
NO-JIRA: Add cpe label to match product security metadata

alebedev87:
The `check-labels` release task enforces the `cpe` label to match the product security metadata. Without this override, the UBI base image `cpe` label is used, which does not match.

Example of failed check-labels job: [link](https://konflux-ui.apps.kflux-prd-rh03.nnv1.p1.openshiftapps.com/ns/rhtap-releng-tenant/applications/ext-dns-optr-1-3-rhel-9/taskruns/managed-k4gpd-check-labels/logs).


grzpiotrowski (approved):
/lgtm
/approve

PR #453, opened 2026-05-13
OCPBUGS-79591: Update FBCs v4.12-v4.22 with v1.3.4 prod bundle

alebedev87:
## Summary

- Add the `v1.3.4` release to all FBC catalog templates (v4.12-v4.22) with the bundle entry in `stable-v1` and `stable-v1.3` channels
- Regenerate the catalogs with `OCP_VERSION=4.x make generate-catalog`


grzpiotrowski (approved):
/approve

PR #460, opened 2026-05-21
NO-JIRA: Add Renovate rules for UBI8 images on `release-1.x` branches

alebedev87:
MintMaker/Renovate reads config only from the default branch (`main`), which only had rules for `ubi9` images. The `release-1.1` and `release-1.2` branches use `ubi8/ubi-minimal` and `ubi8/go-toolset` in their Containerfiles, so Renovate silently skipped them.

Add `matchBaseBranches`-scoped rules for both UBI8 images to enable base image digest updates and Go toolchain version bumps on release branches.

PR #463, opened 2026-05-21
NE-2730: Bump `go-toolset` to 1.25.8 and update `ubi-minimal` digest

alebedev87:
Update `go-toolset` from 1.25.7 to 1.25.8 to match the operand, and refresh the `ubi-minimal` base image digest.
This is supposed to catch up on missed renovate updates (see https://github.com/openshift/external-dns-operator/pull/460).


grzpiotrowski (approved):
/approve
/lgtm

PR #465, opened 2026-05-25
NE-2727: Change bundle relatedImages to registry.stage.redhat.io/edo/

grzpiotrowski:
As a prerequisite to the v1.3.5 stage registry release change the bundle relatedImages to `registry.stage.redhat.io/edo/`.


Thealisyed:
/approve

As it did not accept Melvin's tag

PR #475, opened 2026-05-26
NE-2727: Prevent unnecessary operator and bundle builds

grzpiotrowski:
Disable the operator on-push (merge) pipeline from running if the changes only affect the following directories:
- `bundle-hack/container_digest.sh`
- `catalog/`
- `bundle/`

Disable the bundle on-push pipeline from running if the changes only affect the `catalog/` directory.

The pipelines will run if there are changes to any other file outside of that list also present in the merged changes.

Following the pipelines as code docs [example](https://pipelinesascode.com/docs/guide/matchingevents/#filtering-pipelineruns-to-exclude-non-code-changes) and a similar change already made in https://github.com/openshift/aws-load-balancer-operator/pull/261.


alebedev87 (commented):
/lgtm
/approve

PR #391, opened 2026-04-03
NO-JIRA: Add Konflux release process documentation

alebedev87:
Skeleton for the release process on Konflux. Goal is to be simple and list the main steps.

PR #479, opened 2026-05-28
NE-2730: Switch container digests to stage registry for 1.2.2 release

Thealisyed:
## Summary
- Switch operator and operand pullspecs in `container_digest.sh` from `registry.redhat.io` to `registry.stage.redhat.io`
- Digests remain the same as the nudging PRs (#449, #456)
- Required for the stage release pipeline to pass the Conforma test ([ref: release process doc](https://github.com/openshift/external-dns-operator/pull/391))

## Context
Per the [Konflux release process](https://github.com/openshift/external-dns-operator/pull/391):
> In order for the stage release pipeline to pass the Conforma test, the registry/repository in `bundle-hack/container_digest.sh` must be set to `registry.stage.redhat.io/edo/*`. Keep the digests the same as the ones created by the nudging PR.

## Note
After stage release is verified, a follow-up PR will swap back to `registry.redhat.io` for the production release.

### DIFF
--- a/bundle-hack/container_digest.sh
+++ b/bundle-hack/container_digest.sh
@@ -1,8 +1,8 @@
 # Do not remove comment lines, they are there to reduce conflicts
 # Operator
-export OPERATOR_IMAGE_PULLSPEC='registry.redhat.io/edo/external-dns-rhel8-operator@sha256:cb6bdce66f1c6808c841dbef4a37afa7d62768713e42dfbe6a08992626736fcc'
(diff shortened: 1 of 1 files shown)


Thealisyed:
/close
In favor of #477, Andrey's PR adds `registry.redhat.io/edo/*` mirrors to the deploy pipeline, allowing the bundle to reference the prod registry from the start. 
This eliminates the need to swap to `registry.stage.redhat.io` and back thereby saving two steps in the release process.

PR #483, opened 2026-06-01
NE-2695: Add interval and Infoblox maxResults to ExternalDNS CR

sanjaytripathi97:
## Related issue
- [RFE-6781](https://redhat.atlassian.net/browse/RFE-6781)
- [RFE-8890](https://redhat.atlassian.net/browse/RFE-8890)
- [RFE-8888](https://redhat.atlassian.net/browse/RFE-8888)

Customers integrating OpenShift External DNS Operator with Infoblox need two external-dns flags that were not exposed through the `ExternalDNS` CR:
- `--interval` — controls how often external-dns synchronizes DNS records
- `--infoblox-max-results` — sets the maximum number of DNS records Infoblox returns per request

Without these, operator-managed deployments only received a fixed set of hardcoded args. Manual Deployment patches were reverted on reconcile because the operator owns container args.

This PR exposes both settings declaratively via the `ExternalDNS` CR and wires them through the operator deployment builder.

### New CR fields
| CR field | external-dns flag | Scope |
|----------------|-------------------|-------|
| `spec.intervalSeconds` | `--interval` | All providers |
| `spec.provider.infoblox.maxResults` | `--infoblox-max-results` | Infoblox only |

Both fields are optional (v1beta1 only). When unset/`0`, the operator omits the corresponding flag and external-dns keeps its own defaults. CRD validation: `intervalSeconds` min 60 / max 3600; `maxResults` min 0 / max 10000 (no CRD defaults).

### DIFF
--- a/api/v1beta1/externaldns_types.go
+++ b/api/v1beta1/externaldns_types.go
@@ -101,6 +101,16 @@ type ExternalDNSSpec struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	Zones []string `json:"zones,omitempty"`
+
+	// intervalSeconds specifies the interval in seconds between two consecutive
+	// synchronizations performed by ExternalDNS. When unset, the default is determined by
+	// ExternalDNS, which is currently 60 seconds, but is subject to change over time.
+	// The minimum of 60 seconds is a safe lower bound to avoid hitting provider rate limits.
+	//
+	// +kubebuilder:validation:Minimum=60
+	// +kubebuilder:validation:Maximum=3600
+	// +optional
+	IntervalSeconds int32 `json:"intervalSeconds,omitempty"`
 }
 
 // ExternalDNSDomain describes how sets of included
@@ -361,6 +371,15 @@ type ExternalDNSInfobloxProviderOptions struct {
 	// +kubebuilder:validation:Required
 	// +required
 	WAPIVersion string `json:"wapiVersion"`
+
+	// maxResults sets the maximum number of DNS records that Infoblox returns per request.
+	// The Infoblox default is currently 1000 and requests exceeding it will fail. Increase this when
+	// managing zones with more than 1000 DNS records.
+	//
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=10000
+	// +optional
+	MaxResults int `json:"maxResults,omitempty"`
 }
 
 // SecretReference contains the information to let you locate the desired secret.
--- a/config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml
+++ b/config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml
@@ -692,6 +692,16 @@ spec:
                   - matchType
                   type: object
                 type: array
+              intervalSeconds:
+                description: |-
+                  intervalSeconds specifies the interval in seconds between two consecutive
+                  synchronizations performed by ExternalDNS. When unset, the default is determined by
+                  ExternalDNS, which is currently 60 seconds, but is subject to change over time.
+                  The minimum of 60 seconds is a safe lower bound to avoid hitting provider rate limits.
+                format: int32
+                maximum: 3600
+                minimum: 60
+                type: integer
               provider:
                 description: |-
                   Provider refers to the DNS provider that ExternalDNS
@@ -855,6 +865,14 @@ spec:
                       gridHost:
                         description: GridHost is the IP of the Infoblox Grid host.
                         type: string
+                      maxResults:
+                        description: |-
+                          maxResults sets the maximum number of DNS records that Infoblox returns per request.
+                          The Infoblox default is currently 1000 and requests exceeding it will fail. Increase this when
+                          managing zones with more than 1000 DNS records.
+                        maximum: 10000
+                        minimum: 0
+                        type: integer
                       wapiPort:
                         description: WAPIPort is the port for the Infoblox WAPI.
                         type: integer
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -4,9 +4,12 @@
 package e2e
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"io"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"testing"
@@ -25,6 +28,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -537,7 +541,179 @@ func TestExternalDNSSecretCredentialUpdate(t *testing.T) {
 	}
 }
 
+func TestExternalDNSSyncInterval(t *testing.T) {
+	const (
+		extDNSName               = "test-extdns-interval"
+		intervalSeconds    int32 = 120
+		syncLogMessage           = "All records are already up to date"
+		minObservedSyncGap       = 115 * time.Second
+		maxObservedSyncGap       = 125 * time.Second
+		deploymentTimeout        = 3 * time.Minute
+		logCheckTimeout          = 6 * time.Minute
+	)
+	logTimestampPattern := regexp.MustCompile(`time="([^"]+)"`)
+
+	t.Log("Creating credentials secret")
+	credSecret := helper.makeCredentialsSecret(common.OperatorNamespace)
+	if err := common.KubeClient.Create(context.TODO(), credSecret); err != nil {
+		t.Fatalf("Failed to create credentials secret %s/%s: %v", credSecret.Namespace, credSecret.Name, err)
+	}
+	defer func() {
+		if err := common.KubeClient.Delete(context.TODO(), credSecret); err != nil && !errors.IsNotFound(err) {
+			t.Errorf("Failed to delete credentials secret %s/%s: %v", credSecret.Namespace, credSecret.Name, err)
+		}
+	}()
+
+	t.Log("Creating external dns instance with intervalSeconds")
+	extDNS := helper.buildOpenShiftExternalDNS(extDNSName, hostedZoneID, hostedZoneDomain, "", credSecret)
+	extDNS.Spec.IntervalSeconds = intervalSeconds
+	if err := common.KubeClient.Create(context.TODO(), &extDNS); err != nil {
+		t.Fatalf("Failed to create external DNS %q: %v", extDNSName, err)
+	}
+	defer func() {
+		if err := common.KubeClient.Delete(context.TODO(), &extDNS); err != nil && !errors.IsNotFound(err) {
+			t.Errorf("Failed to delete external DNS %q: %v", extDNSName, err)
+		}
+	}()
+
+	deploymentKey := types.NamespacedName{
+		Namespace: common.OperatorNamespace,
+		Name:      fmt.Sprintf("external-dns-%s", extDNSName),
+	}
+	intervalArg := fmt.Sprintf("--interval=%ds", intervalSeconds)
+
+	t.Logf("Waiting for operand deployment %s with %s", deploymentKey, intervalArg)
+	var containerName string
+	if err := wait.PollUntilContextTimeout(context.TODO(), common.DnsPollingInterval, deploymentTimeout, true, func(ctx context.Context) (bool, error) {
+		deployment := &appsv1.Deployment{}
+		if err := common.KubeClient.Get(ctx, deploymentKey, deployment); err != nil {
+			t.Logf("Operand deployment not created yet: %v", err)
+			return false, nil
+		}
+		if deployment.Status.AvailableReplicas < 1 {
+			t.Log("Operand deployment is not available yet")
+			return false, nil
+		}
+
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if !strings.HasPrefix(container.Name, "external-dns") {
+				continue
+			}
+			if containsContainerArg(container.Args, intervalArg) {
+				containerName = container.Name
+				return true, nil
+			}
+		}
+		return false, fmt.Errorf("%q was not found", intervalArg)
+	}); err != nil {
+		t.Fatalf("Operand deployment did not get expected interval arg: %v", err)
+	}
+
+	t.Logf("Verifying sync interval from operand logs in %s", deploymentKey.Namespace)
+	if err := wait.PollUntilContextTimeout(context.TODO(), common.DnsPollingInterval, logCheckTimeout, true, func(ctx context.Context) (bool, error) {
+		list := &corev1.PodList{}
+		if err := common.KubeClient.List(ctx, list,
+			client.InNamespace(deploymentKey.Namespace),
+			client.MatchingLabels{
+				"app.kubernetes.io/name":     "external-dns",
+				"app.kubernetes.io/instance": extDNSName,
+			},
+		); err != nil {
+			t.Logf("Failed to list pods: %v", err)
+			return false, nil
+		}
+		var podName string
+		for i := range list.Items {
+			if list.Items[i].Status.Phase == corev1.PodRunning {
+				podName = list.Items[i].Name
+				break
+			}
+		}
+		if podName == "" {
+			t.Log("Operand pod is not running yet")
+			return false, nil
+		}
+
+		timestamps, err := collectSyncTimestamps(ctx, deploymentKey.Namespace, podName, containerName, syncLogMessage, logTimestampPattern)
+		if err != nil {
+			t.Logf("Failed to read pod logs: %v", err)
+			return false, nil
+		}
+		if len(timestamps) < 2 {
+			t.Logf("Waiting for at least 2 sync log entries, got %d", len(timestamps))
+			return false, nil
+		}
+
+		prev := timestamps[len(timestamps)-2]
+		curr := timestamps[len(timestamps)-1]
+		gap := curr.Sub(prev)
+		t.Logf("Observed sync gap between %s and %s: %s", prev.Format(time.RFC3339), curr.Format(time.RFC3339), gap)
+		if gap < minObservedSyncGap || gap > maxObservedSyncGap {
+			t.Logf("Sync gap %s outside expected range [%s, %s]; waiting for another cycle", gap, minObservedSyncGap, maxObservedSyncGap)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		t.Fatalf("Failed to verify sync interval from operand logs: %v", err)
+	}
+}
+
 // HELPER FUNCTIONS
+
+func containsContainerArg(args []string, expected string) bool {
+	for _, arg := range args {
+		if arg == expected {
+			return true
+		}
+	}
+	return false
+}
+
+func collectSyncTimestamps(ctx context.Context, namespace, podName, containerName, syncLogMessage string, logTimestampPattern *regexp.Regexp) ([]time.Time, error) {
+	opts := &corev1.PodLogOptions{
+		Container: containerName,
+		Follow:    false,
+	}
+	readCloser, err := common.KubeClientSet.CoreV1().Pods(namespace).GetLogs(podName, opts).Stream(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		_ = readCloser.Close()
+	}()
+
+	return parseSyncTimestamps(readCloser, syncLogMessage, logTimestampPattern)
+}
+
+func parseSyncTimestamps(r io.Reader, syncLogMessage string, logTimestampPattern *regexp.Regexp) ([]time.Time, error) {
+	var timestamps []time.Time
+	scanner := bufio.NewScanner(r)
+	buf := make([]byte, 0, 64*1024)
+	scanner.Buffer(buf, 1024*1024)
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		if !strings.Contains(line, syncLogMessage) {
+			continue
+		}
+		matches := logTimestampPattern.FindStringSubmatch(line)
+		if len(matches) != 2 {
+			continue
+		}
+		ts, err := time.Parse(time.RFC3339Nano, matches[1])
+		if err != nil {
+			ts, err = time.Parse(time.RFC3339, matches[1])
+			if err != nil {
+				continue
+			}
+		}
+		timestamps = append(timestamps, ts)
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+	return timestamps, nil
+}
 
 func verifyCNAMERecordForOpenshiftRoute(ctx context.Context, t *testing.T, canonicalName, host string) {
 	// try all nameservers and fail only if all failed
--- a/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/external-dns-operator.clusterserviceversion.yaml
@@ -307,12 +307,14 @@ metadata:
                 "name": "myzonedomain.com"
               }
             ],
+            "intervalSeconds": 300,
             "provider": {
               "infoblox": {
                 "credentials": {
                   "name": "infoblox-credentials"
                 },
                 "gridHost": "100.100.100.100",
+                "maxResults": 2000,
                 "wapiPort": 443,
                 "wapiVersion": "2.12.2"
               },
--- a/bundle/manifests/externaldns.olm.openshift.io_externaldnses.yaml
+++ b/bundle/manifests/externaldns.olm.openshift.io_externaldnses.yaml
@@ -692,6 +692,16 @@ spec:
                   - matchType
                   type: object
                 type: array
+              intervalSeconds:
+                description: |-
+                  intervalSeconds specifies the interval in seconds between two consecutive
+                  synchronizations performed by ExternalDNS. When unset, the default is determined by
+                  ExternalDNS, which is currently 60 seconds, but is subject to change over time.
+                  The minimum of 60 seconds is a safe lower bound to avoid hitting provider rate limits.
+                format: int32
+                maximum: 3600
+                minimum: 60
+                type: integer
               provider:
                 description: |-
                   Provider refers to the DNS provider that ExternalDNS
@@ -855,6 +865,14 @@ spec:
                       gridHost:
                         description: GridHost is the IP of the Infoblox Grid host.
                         type: string
+                      maxResults:
+                        description: |-
+                          maxResults sets the maximum number of DNS records that Infoblox returns per request.
+                          The Infoblox default is currently 1000 and requests exceeding it will fail. Increase this when
+                          managing zones with more than 1000 DNS records.
+                        maximum: 10000
+                        minimum: 0
+                        type: integer
                       wapiPort:
                         description: WAPIPort is the port for the Infoblox WAPI.
                         type: integer
--- a/config/samples/infoblox/operator_v1beta1_infoblox_openshift.yaml
+++ b/config/samples/infoblox/operator_v1beta1_infoblox_openshift.yaml
@@ -15,6 +15,8 @@ spec:
       gridHost: "100.100.100.100"
       wapiPort: 443
       wapiVersion: "2.12.2"
+      maxResults: 2000
+  intervalSeconds: 300
   source:
     # Source Type is route resource of OpenShift
     type: OpenShiftRoute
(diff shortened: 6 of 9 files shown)


theiratenate:
@alebedev87 @grzpiotrowski  Please review when you can. We have some developers looking to use CNAME records in our root domain and need the max results added to resolve their plight.

alebedev87:
@theiratenate, @sanjaytripathi97 : https://redhat.atlassian.net/browse/RFE-6781 is not accepted yet. But I think it would be easier to do so if the PR will be in a ready state. Let me have a look.

api/v1alpha1/externaldns_types.go
alebedev87, line 1: We don't need updates in alpha version, only beta update is enough.

api/v1beta1/externaldns_types.go
```diff
@@ -101,6 +101,14 @@ type ExternalDNSSpec struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	Zones []string `json:"zones,omitempty"`
+
+	// Interval specifies the interval between two consecutive synchronizations
+	// performed by ExternalDNS. When omitted, ExternalDNS uses its default
+	// interval of 1 minute.
+	//
+	// +kubebuilder:validation:Optional
+	// +optional
```
alebedev87: Default value as long as minimum and maximum values should be set the CRD. This will enable early validation without saving the state in the server. The default value should be the same at the current default to not change the existing behavior.

api/v1beta1/externaldns_types.go
```diff
@@ -361,6 +369,15 @@ type ExternalDNSInfobloxProviderOptions struct {
...
 	// +required
 	WAPIVersion string `json:"wapiVersion"`
+
+	// MaxResults sets the _max_results query parameter on Infoblox WAPI GET
+	// requests. This should be set when integrating with Infoblox grids that
+	// contain a large number of DNS records.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +optional
```
alebedev87: Similar comment, we need to set the default which is the same as the current used one and we need to think of what the maximum value can be reasonable for the max results.

Overall, I had small reservations about exposing max_results because the problems related to the need to customize this parameter are supposed to be fixed with pagination which is built-in in [the webhook implementation](https://github.com/AbsaOSS/external-dns-infoblox-webhook). However I can see that the webhook implementation does suppose max_results parameter too, so this field will make sense in the future too (when we'll migrate to the webhook provider for Infoblox).

api/v1beta1/externaldns_webhook.go
```diff
@@ -73,6 +73,8 @@ func (r *ExternalDNS) validate(old runtime.Object) error {
 		r.validateHostnameAnnotationPolicy(),
 		r.validateProviderCredentials(),
 		r.validateAWSRoleARN(),
+		r.validateInterval(),
+		r.validateInfobloxMaxResults(),
```
alebedev87: We prefer CRD validation over webhook because it's earlier in the life cycle of the CR.

pkg/operator/controller/externaldns/pod_interval_test.go
alebedev87, line 1: Unit test is a good start but we will need some e2e test scenarios to back up this change. You can find the e2e tests in `test/e2e/` (from the root of repository).

api/v1alpha1/externaldns_types.go
sanjaytripathi97: Agreed, I've reverted all `v1alpha1` changes. The new fields (`spec.interval` and `spec.provider.infoblox.maxResults`) are only added to `v1beta1`, and the CRD schema is regenerated for the storage version accordingly.

api/v1beta1/externaldns_types.go
sanjaytripathi97: Done. `spec.interval` now has CRD-level validation:

- default:`1m` (matches external-dns default)
- minimum: `10s`
- maximum: `24h` 

The operator passes `--interval` only when the value is greater than zero. With the CRD default of `1m`, new resources get the same behavior as external-dns today without requiring users to set the field explicitly.

api/v1beta1/externaldns_types.go
sanjaytripathi97: Done. `spec.provider.infoblox.maxResults` now has CRD-level validation:

- default: `0` (current behavior `_max_results` is not sent to WAPI)
- minimum: `0`
- maximum: `10000`

The operator only adds `--infoblox-max-results` when the value is greater than zero, so existing Infoblox deployments are unchanged unless the user opts in.

I agree this field will remain useful even after a potential move to the Infoblox webhook provider, since that implementation also supports `max_results`. This gives customers a supported way to configure it today via the `ExternalDNS` CR.

api/v1beta1/externaldns_webhook.go
sanjaytripathi97: Agreed I've removed `validateInterval()` and `validateInfobloxMaxResults()` from the webhook. Validation for both fields is now handled entirely via CRD schema defaults and bounds, so invalid values are rejected earlier without relying on the admission webhook.

pkg/operator/controller/externaldns/pod_interval_test.go
sanjaytripathi97: Added `test/e2e/infoblox_sync_options_test.go` with `TestExternalDNSInfobloxSyncOptions`. The test:

1. Creates an Infoblox `ExternalDNS` CR with `interval` and `maxResults` set
2. Waits for the operand deployment to become available
3. Verifies the external-dns container args include `--interval=...` and `--infoblox-max-results=...`

The test is skipped unless the e2e suite runs against the Infoblox provider. Unit tests in `pod_interval_test.go` continue to cover the deployment arg builder logic, let me know if this is fine.

sanjaytripathi97: @alebedev87  Can you pls review again ?

api/v1beta1/externaldns_types.go
alebedev87: After reviewing the [OpenShift API conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md), I'd like to change my earlier suggestion about the default value. The convention says to [default in the controller, not in the CRD](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#defaulting) so the platform can change defaults over time without an API change. So no `+kubebuilder:default` here - when the value is `0` (unset), the controller should just skip `--interval` and let external-dns use its own default (as you implemented it already).

Also, this field should be renamed to `IntervalSeconds` and use `int32` instead of `metav1.Duration`. The duration type has round-trip instability (the default `"1m"` becomes `"1m0s"` after Go marshaling) and requires CEL validation which gives cryptic errors on malformed input. A plain integer with `+kubebuilder:validation:Minimum=60` and `+kubebuilder:validation:Maximum=3600` is stable and follows the [convention of not using pointers](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#pointers) for optional CRD fields. The minimum of `60s` matches external-dns's own default - since the field is scoped to all providers lowering below could easily hit rate limits especially on AWS. The maximum of 1 hour is generous enough for load reduction without risking long DNS drift (let me know about strong reasons of having higher upper bound). Both bounds can be relaxed in the future if needed as that's a backwards-compatible change.

api/v1beta1/externaldns_types.go
alebedev87: Same as above, per [OpenShift API conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#defaulting) the default should be left for the controller. Drop `+kubebuilder:default:=0`. `Minimum=0` is correct here because external-dns itself skips setting the `_max_results` query param when the value is `0`. `Maximum=10000` seems reasonable - it's 10 times the Infoblox default of `1000` and well above typical usage. We can raise it later if needed as that's a backwards-compatible change.

pkg/operator/controller/externaldns/pod_interval_test.go
alebedev87: The current e2e test only checks that the flags are propagated to the operand deployment args. That's already covered by the unit test. For e2e, we should verify the actual behavior - that the sync interval is really in effect. One approach (however open for other ideas): create an ExternalDNS CR with 120 seconds interval, wait for the operand pod to start, then check the pod logs for consecutive sync entries and verify that the timestamps are spaced roughly 120 seconds apart, not the default 60 seconds. This way we will testing the real outcome, not just flag plumbing.

alebedev87 (commented):
Thanks a lot for the follow-up! I think we are closer to the good implementation. I re-reviewed the OpenShift API conventions meanwhile and had to reconsider the defaulting strategy PTAL.

api/v1beta1/externaldns_types.go
sanjaytripathi97: Updated per the OpenShift API conventions:

Renamed `interval` → `intervalSeconds` (`int32`)
Removed CRD default (controller leaves unset/`0` alone so external-dns keeps its own 60s default)
CRD validation: `Minimum=60`, `Maximum=3600`
Operator emits `--interval=<N>s` only when `intervalSeconds > 0`

api/v1beta1/externaldns_types.go
sanjaytripathi97: Updated:
Removed `+kubebuilder:default:=0`
Kept `Minimum=0` and `Maximum=10000`
When unset/`0`, operator does not pass `--infoblox-max-results` (same behavior as today)

pkg/operator/controller/externaldns/pod_interval_test.go
sanjaytripathi97: Updated `TestExternalDNSInfobloxSyncOptions` to verify real sync behavior:

Creates an Infoblox ExternalDNS with `intervalSeconds: 120` and `maxResults: 2000`
Confirms operand args include the expected flags
Reads operand pod logs for consecutive `"All records are already up to date"` entries
Asserts the gap is in `[90s, 150s]` (clearly above the 60s default)

sanjaytripathi97: @alebedev87 I have made the changes again, Thanks for the inputs !!

alebedev87:
I just checked out and found that https://redhat.atlassian.net/browse/RFE-8890 and https://redhat.atlassian.net/browse/RFE-8888 got approved in April. This PR addresses both of these RFEs. Let's reference them in the description.

api/v1beta1/externaldns_types.go
```diff
@@ -101,6 +101,16 @@ type ExternalDNSSpec struct {
 	// +kubebuilder:validation:Optional
 	// +optional
 	Zones []string `json:"zones,omitempty"`
+
+	// IntervalSeconds specifies the interval in seconds between two consecutive
+	// synchronizations performed by ExternalDNS. When unset (0), ExternalDNS
+	// uses its own default of 60 seconds. The minimum of 60 seconds matches
+	// that default; values below it risk hitting provider rate limits.
```
alebedev87: The comment needs to reference the json tag to be respected by `oc explain`. Also we need to mention that the default is a subject to change:
```suggestion
	// intervalSeconds specifies the interval in seconds between two consecutive
	// synchronizations performed by ExternalDNS. When unset, the default is determined by 
	// ExternalDNS, which is currently 60 seconds, but is subject to change over time.
	// The minimum of 60 seconds is a safe lower bound to avoid hitting provider rate limits.
```

api/v1beta1/externaldns_types.go
```diff
@@ -361,6 +371,16 @@ type ExternalDNSInfobloxProviderOptions struct {
 	// +kubebuilder:validation:Required
 	// +required
 	WAPIVersion string `json:"wapiVersion"`
+
+	// MaxResults sets the _max_results query parameter on Infoblox WAPI GET
+	// requests. When unset (0), the parameter is not set and the Infoblox
+	// server default is used. Set this to a positive value when integrating
+	// with Infoblox grids that contain a large number of DNS records.
```
alebedev87: To omit some implementation specific details: 
```suggestion
	// maxResults sets the maximum number of DNS records that Infoblox returns per request.
	// The Infoblox default is currently 1000 and requests exceeding it will fail. Increase this when
	// managing zones with more than 1000 DNS records.
```

config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml
alebedev87, line 1: Make sure you run `make bundle` to regenerate the bundle contents with the new CRD and sample contents.

pkg/operator/controller/externaldns/pod_interval_test.go
alebedev87, line 1: I think that we should add both interval and maxresults tests in the existing `TestDesiredExternalDNSDeployment` test function in `deployment_test.go`. A new Infoblox test case which specifies both of the new fields. The omitted values are supposed to be covered by the existing Infloblox test case(s).

test/e2e/infoblox_sync_options_test.go
```diff
@@ -0,0 +1,228 @@
...
+	minObservedSyncGap = 90 * time.Second
+	maxObservedSyncGap = 150 * time.Second
+)
+
+var logTimestampPattern = regexp.MustCompile(`time="([^"]+)"`)
+
+func TestExternalDNSInfobloxSyncOptions(t *testing.T) {
+	if helper.platform() != infobloxDNSProvider {
+		t.Skip("test only runs against Infoblox provider")
+	}
```
alebedev87: We try to keep all test provider agnostic and add them to `operator_test.go`. The `intervalSeconds` is exactly this case. Let's add it there and not skip for any provider.

test/e2e/infoblox_sync_options_test.go
```diff
@@ -0,0 +1,228 @@
...
+			t.Logf("operand deployment not ready yet: %v", err)
+			return false, nil
+		}
+		if deployment.Status.AvailableReplicas < 1 {
+			t.Log("operand deployment is not available yet")
+			return false, nil
+		}
+
+		intervalArg := fmt.Sprintf("--interval=%ds", testSyncIntervalSeconds)
+		maxResultsArg := fmt.Sprintf("--infoblox-max-results=%d", testInfobloxMaxResults)
```
alebedev87: A end to end test for max result is not that easy to implement. We would need to go beyond the current default which 1000 records. This implies quite a heavy scenario which would create more than 1000 routes or services. I think that we can rely solely on the unit testing for the max result field (at least for the moment).

alebedev87 (commented):
Another review with a deeper look into unit and e2e tests.

api/v1beta1/externaldns_types.go
sanjaytripathi97: Done. Updated the field comment to use the json tag name and note that the ExternalDNS default is currently 60 seconds but subject to change. Tried regenerate the CRD/bundle descriptions as well.

api/v1beta1/externaldns_types.go
sanjaytripathi97: Done. Updated the maxResults comment as suggested and regenerated the CRD/bundle.

pkg/operator/controller/externaldns/pod_interval_test.go
sanjaytripathi97: Done. Added an Infoblox case in TestDesiredExternalDNSDeployment that sets both intervalSeconds and maxResults and asserts --interval=120s and --infoblox-max-results=2000. Omitted values remain covered by the existing Nominal Infoblox case. Removed the standalone pod_interval_test.go.

test/e2e/infoblox_sync_options_test.go
sanjaytripathi97: Done. Moved intervalSeconds coverage into TestExternalDNSSyncInterval in operator_test.go. It runs for all providers (no Infoblox-only skip) and verifies both the operand arg and sync spacing from logs. Removed infoblox_sync_options_test.go.

test/e2e/infoblox_sync_options_test.go
sanjaytripathi97: Agreed. Dropped the Infoblox maxResults e2e. Coverage is in the unit test case in TestDesiredExternalDNSDeployment for now.

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
 	}
 }
 
+func TestExternalDNSSyncInterval(t *testing.T) {
```
alebedev87: We prefer to let the file grow by appending to the end. Can you please add this new test after the last `TestXXX` function?

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+	expected := []appsv1.DeploymentCondition{
+		{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionTrue},
+	}
+	if err := waitForOperatorDeploymentStatusCondition(context.TODO(), t, common.KubeClient, expected...); err != nil {
+		t.Fatalf("operator was not available: %v", err)
+	}
```
alebedev87: This is redundant `TestOperatorAvailable` should validate this before `TestExternalDNSSyncInterval` kicks in.

test/e2e/operator_test.go
```diff
@@ -33,13 +37,20 @@ const (
 	testServiceName = "test-service"
 	testRouteName   = "test-route"
 	testExtDNSName  = "test-extdns"
+
+	testSyncIntervalExtDNSName       = "test-extdns-interval"
+	testSyncIntervalSeconds    int32 = 120
+	syncLogMessage                   = "All records are already up to date"
+	minObservedSyncGap               = 90 * time.Second
+	maxObservedSyncGap               = 150 * time.Second
```
alebedev87: Those are variables which make sense only for `TestExternalDNSSyncInterval`, so should be moved in there.

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+		Name:      fmt.Sprintf("external-dns-%s", testSyncIntervalExtDNSName),
+	}
+	intervalArg := fmt.Sprintf("--interval=%ds", testSyncIntervalSeconds)
+
+	t.Logf("Waiting for operand deployment %s with %s", deploymentKey, intervalArg)
+	var containerName string
+	if err := wait.PollUntilContextTimeout(context.TODO(), 2*time.Second, 3*time.Minute, true, func(ctx context.Context) (bool, error) {
+		deployment := &appsv1.Deployment{}
+		if err := common.KubeClient.Get(ctx, deploymentKey, deployment); err != nil {
+			t.Logf("operand deployment not ready yet: %v", err)
```
alebedev87:
```suggestion
			t.Logf("Operand deployment not created yet: %v", err)
```

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+			return false, nil
+		}
+		if deployment.Status.AvailableReplicas < 1 {
+			t.Log("operand deployment is not available yet")
```
alebedev87:
```suggestion
			t.Log("Operand deployment is not available yet")
```

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+			if !containsContainerArg(container.Args, intervalArg) {
+				t.Logf("container %s is missing %q", container.Name, intervalArg)
+				return false, nil
+			}
+		}
+		if !foundExternalDNSContainer {
+			t.Log("external-dns container not found in deployment")
+			return false, nil
+		}
+		return true, nil
```
alebedev87: If container of args are not there it's a hard failure - no need to retry, they won't appear later:
```suggestion
		for _, container := range deployment.Spec.Template.Spec.Containers {
			if !strings.HasPrefix(container.Name, "external-dns") {
				continue
			}
			if containsContainerArg(container.Args, intervalArg) {
				return true, nil
			}
		}
        return false, fmt.Errorf("%q was not found", intervalArg)
```

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+				continue
+			}
+			podName = pod.Name
+			return true, nil
+		}
+		t.Log("operand pod not running yet")
+		return false, nil
+	}); err != nil {
+		t.Fatalf("operand pod did not become ready: %v", err)
+	}
```
alebedev87: I don't think we need this check, the deployment is there which means that the operator job is done. Pod creation is Kubernetes job.

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+
+	t.Logf("Verifying sync interval from logs of pod %s/%s", deploymentKey.Namespace, podName)
+	if err := wait.PollUntilContextTimeout(context.TODO(), 15*time.Second, 6*time.Minute, true, func(ctx context.Context) (bool, error) {
+		timestamps, err := collectSyncTimestamps(ctx, deploymentKey.Namespace, podName, containerName)
+		if err != nil {
+			t.Logf("failed to read pod logs: %v", err)
```
alebedev87:
```suggestion
			t.Logf("Failed to read pod logs: %v", err)
```

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+			return false, nil
+		}
+		if len(timestamps) < 2 {
+			t.Logf("waiting for at least 2 sync log entries, got %d", len(timestamps))
```
alebedev87:
```suggestion
			t.Logf("Waiting for at least 2 sync log entries, got %d", len(timestamps))
```

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+			return false, nil
+		}
+
+		prev := timestamps[len(timestamps)-2]
+		curr := timestamps[len(timestamps)-1]
+		gap := curr.Sub(prev)
+		t.Logf("observed sync gap between %s and %s: %s", prev.Format(time.RFC3339), curr.Format(time.RFC3339), gap)
```
alebedev87:
```suggestion
		t.Logf("Observed sync gap between %s and %s: %s", prev.Format(time.RFC3339), curr.Format(time.RFC3339), gap)
```

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+		if gap < minObservedSyncGap || gap > maxObservedSyncGap {
```
alebedev87: Why the range is so vast? The sync interval is `120s` and we are expecting it to be between `90s` and `150s`. I could understand `+/- 5s` but `+/- 30s` seems like a little too much.

test/e2e/operator_test.go
```diff
@@ -117,6 +128,192 @@ func TestOperatorAvailable(t *testing.T) {
...
+				continue
+			}
+		}
+		timestamps = append(timestamps, ts)
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+	return timestamps, nil
+}
```
alebedev87: Can you please move all the helper functions to the bottom of the file? There is a dedicated comment marker there.

test/common/common.go
```diff
@@ -317,6 +317,16 @@ func GetServiceIPs(ctx context.Context, t *testing.T, timeout time.Duration, svc
...
 
+// OperandNamespace returns the namespace where ExternalDNS operand Deployments are created.
+// By default the operator deploys operands into its own namespace; when
+// E2E_SEPARATE_OPERAND_NAMESPACE=true it uses the dedicated "external-dns" namespace.
+func OperandNamespace() string {
+	if os.Getenv(e2eSeparateOperandNsEnvVar) == "true" {
+		return operandNamespace
+	}
+	return OperatorNamespace
+}
```
alebedev87: Deploying operands in `external-dns-operator` is defacto standard now. No need to count for the other case, you can drop this function and use `common.OperatorNamespace` wherever the operand namespace is needed.

alebedev87 (commented):
The API LGTM, some remarks about the e2e test though.

test/e2e/operator_test.go
sanjaytripathi97: Done. Moved TestExternalDNSSyncInterval to the end of operator_test.go, after the other TestXXX functions.

test/e2e/operator_test.go
sanjaytripathi97: Done. Removed the extra operator availability wait; TestOperatorAvailable covers that earlier.

test/e2e/operator_test.go
sanjaytripathi97: Done. Moved the sync-interval-only constants/vars into TestExternalDNSSyncInterval.

test/e2e/operator_test.go
sanjaytripathi97: Done. Once the deployment exists, missing --interval is a hard failure instead of a retry.

test/e2e/operator_test.go
sanjaytripathi97: Done. Removed the dedicated pod readiness wait. Pod lookup now happens only while reading logs for the sync-interval check.

test/e2e/operator_test.go
sanjaytripathi97: Done. Tightened the expected gap to [115s, 125s] (±5s around the configured 120s interval), I hope this is fine now.

test/e2e/operator_test.go
sanjaytripathi97: Done. Moved the helper functions under the existing // HELPER FUNCTIONS marker.

test/common/common.go
sanjaytripathi97: Done. Removed OperandNamespace() and now use common.OperatorNamespace for operand lookups.

test/e2e/operator_test.go
```diff
@@ -537,8 +541,184 @@ func TestExternalDNSSecretCredentialUpdate(t *testing.T) {
...
+		minObservedSyncGap       = 115 * time.Second
+		maxObservedSyncGap       = 125 * time.Second
+	)
+	logTimestampPattern := regexp.MustCompile(`time="([^"]+)"`)
+
+	t.Log("Ensuring test namespace")
+	err := common.KubeClient.Create(context.TODO(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNamespace}})
+	if err != nil && !errors.IsAlreadyExists(err) {
+		t.Fatalf("Failed to ensure namespace %s: %v", testNamespace, err)
+	}
```
alebedev87: I just realized that we create a namespace here but it has no use in the test case. Usually we use it to create a source resource (Route or Service) but here we don't create any.

test/e2e/operator_test.go
```diff
@@ -537,8 +541,184 @@ func TestExternalDNSSecretCredentialUpdate(t *testing.T) {
...
+
+	deploymentKey := types.NamespacedName{
+		Namespace: common.OperatorNamespace,
+		Name:      fmt.Sprintf("external-dns-%s", extDNSName),
+	}
+	intervalArg := fmt.Sprintf("--interval=%ds", intervalSeconds)
+
+	t.Logf("Waiting for operand deployment %s with %s", deploymentKey, intervalArg)
+	var containerName string
+	if err := wait.PollUntilContextTimeout(context.TODO(), 2*time.Second, 3*time.Minute, true, func(ctx context.Context) (bool, error) {
```
alebedev87: Let's use a test level `const` for both timeouts: deployment and log check. The polling can be the same as  for other tests: `common.DnsPollingInterval` or you can create dedicated function level `const`s for them too.

alebedev87 (commented):
2 last finishing touches, the rest LGTM. Let's fix them and get the CI green. And one more thing: commits - let's fix them up into a single one with a good description, thanks!

test/e2e/operator_test.go
sanjaytripathi97: Done. Removed the unused namespace creation from TestExternalDNSSyncInterval since this test does not create a Route or Service source.

test/e2e/operator_test.go
sanjaytripathi97: Done. Added test-level deploymentTimeout and logCheckTimeout consts, and switched both waits to use common.DnsPollingInterval for polling.

sanjaytripathi97: I also squashed the PR into a single commit and updated the description. Please let me know if the description still needs any corrections.

alebedev87: Both (commit and description) look good, thanks a lot @sanjaytripathi97!

alebedev87:
All CI is green.

/approve
/lgtm

Holding to let NI&D team have a look.
/hold

candita:
/retitle [NE-2695](https://redhat.atlassian.net/browse/NE-2695): Add interval and Infoblox maxResults to ExternalDNS CR

alebedev87:
After a discussion with @bentito and @grzpiotrowski, things look good, unholding the PR.

/unhold

alebedev87:
/ok-to-test

Again?

alebedev87 (approved):

grzpiotrowski (approved):
/approve

PR #484, opened 2026-06-02
NE-2730: Update FBCs v4.12-v4.22 with v1.2.2 prod bundle

Thealisyed:
Replace the stage `v1.2.2` bundle image with the production one from `registry.redhat.io` in all FBC catalogs (v4.12-v4.22). The digest remains the same.


alebedev87 (commented):
/lgtm
/approve

PR #485, opened 2026-06-03
NO-JIRA: Add candidate tag to v4.22 FBC push pipeline

alebedev87:
## Summary
- The `ADDITIONAL_TAGS: ["candidate"]` parameter was missing from the `apply-tags` task in the v4.22 FBC push pipeline, while v4.19–v4.21 already had it.


grzpiotrowski (approved):
/approve

PR #488, opened 2026-06-05
OCPBUGS-82447: Bump Go to 1.25.9 to fix CVE-2026-32282

davidesalerno:
## Summary

- Bumps the `go` directive in `go.mod` from `1.25.0` to `1.25.9` to address [CVE-2026-32282](https://nvd.nist.gov/vuln/detail/CVE-2026-32282)
- Go 1.25.0–1.25.8 are affected: `Root.Chmod` can follow symlinks out of the root directory on Linux due to `fchmodat` silently ignoring `AT_SYMLINK_NOFOLLOW`
- No application code changes — toolchain version bump only
- Dockerfiles are fine because:
     
  - Containerfile.external-dns-operator (Konflux/production build) — already pinned to go-toolset:1.25.9-1778675823, which is the patched version. No change needed.
  - Dockerfile (upstream/CI build) — uses the floating tag go-toolset:1.25, which resolves to the latest 1.25.x image in Red Hat's registry (currently 1.25.9). Even if someone happened to pull a stale cached image, the go 1.25.9
  directive we just set in go.mod would force the Go toolchain to download and use Go 1.25.9 at build time. So it's covered from both sides.
  - drift-cache/Dockerfile — must be a SHA1-identical copy of Dockerfile (enforced by the Containerfile's drift detection stage). Since Dockerfile is unchanged, this stays in sync.

## Verification

- `make build-operator` passes
- All unit tests pass (`make test` — `covdata` error is pre-existing local toolchain issue, not related to this change)
- `go fmt` and `go vet` clean

## Test plan

- [x] `make build-operator` succeeds
- [x] All unit test packages pass
- [x] `go vet ./...` clean
- [x] No vendor changes needed

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve OCPBUGS-82447`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift/external-dns-operator
 
-go 1.25.0
+go 1.25.9
 
 require (
 	github.com/Azure/azure-sdk-for-go v60.1.0+incompatible


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/external-dns-operator
 
-go 1.25.0
+go 1.25.9
```
grzpiotrowski, line 3: I'm wondering if this is necessary to address the fix here, since according to [go docs](https://go.dev/doc/modules/gomod-ref#go-syntax) the `go 1.25.0` in `go.mod` specifies:
> The minimum version of Go required to compile packages in this module.

And more about this [here](https://go.dev/doc/toolchain#intro):
> In the standard configuration, the go command uses its own bundled toolchain when that toolchain is at least as new as the go or toolchain lines in the main module or workspace.

So when the `go-toolset:1.25.9` builder image runs with go `1.25.0` in go.mod, it uses the Go `1.25.9`.

PR #444 was an automated go-toolset builder image bump to `1.25.9` which includes the CVE fix. And this gave us the operator image built with the go version including the CVE fix.

alebedev87: Right, changing the minimal go version is not enough. https://github.com/openshift/external-dns-operator/pull/444 already bumped the Konflux go toolchain to `1.25.9` though.

davidesalerno: @grzpiotrowski this change is strictly not needed to fix the CVE if we are using the operator version built from the CI but if we want to explicitly document "this module requires 1.25.9+" as a policy signal (e.g., preventing anyone from accidentally building with a vulnerable toolchain locally), updating the go.mod will do this job. 

But that's a style/policy choice, not a security necessity.

Would we like to proceed in this direction?

alebedev87: As discussed with Davide in slack, changing the minimal version is an additional change which we never really did for toolchain CVE fixes. Sticking to a single PR (go toolset bump) keeps the implementation to the bare minimum helping to speed up the process. Also, it's covered by Konflux MintMaker (renovate) out of the box.

bentito: Just as part of my review and to summarize and make sure I'm understanding this discussion:
- We don't need this PR to fix the bug, #444 handled it
- We want to rely on MintMaker automation for bumps to go version and go toolchain version
Did I get that right?

bentito:
@alebedev87 I saw you retested this PR, but why? Based on your comment and Grzeg's I'd think just close this PR as #444 was done automatically and provides go upgrade and hence fix for the CVE, no?

alebedev87:
> I saw you retested this PR, but why?

@bentito : I was just wondering whether there was something failing, as an additional argument to give up on this change. It turned out to be just ephemeral hiccups, the change is valid.

> Based on your comment and Grzeg's I'd think just close this PR as https://github.com/openshift/external-dns-operator/pull/444 was done automatically and provides go upgrade and hence fix for the CVE, no?

Right, we can close this one and go to the releasing part directly.

PR #490, opened 2026-06-07
RFE-6781: Supportability for infoblox-flags --infoblox-max-results and --interval in external-dns-operator

swshende-cmd:
## Summary

* Adds a new optional `maxResults` field to `ExternalDNSInfobloxProviderOptions` (both v1alpha1 and v1beta1) that maps to the `--infoblox-max-results` CLI flag, enabling Infoblox customers with large DNS zones (1000+ records) to avoid "Result set too large" WAPI errors.
* Adds a new optional `interval` field to `ExternalDNSSpec` (both v1alpha1 and v1beta1) that maps to the `--interval` CLI flag, allowing customers to tune the polling interval for DNS record synchronization.
* Includes 4 new unit test cases, updated Infoblox sample YAML, and updated documentation.

### Problem

Infoblox customers with DNS zones containing more than 1000 records encounter a **"Result set too large"** error from the Infoblox WAPI. The ExternalDNS binary supports a `--infoblox-max-results` flag to work around this, but the external-dns-operator had no way to pass this flag through to the operand container.

Additionally, the default polling interval (1 minute) is hardcoded with no way to tune it. Customers who need faster DNS propagation (e.g., during deployments or failovers) or want to reduce API calls to Infoblox (e.g., to stay within rate limits) had no supported mechanism to adjust it.

Ref: [RFE-6781](https://redhat.atlassian.net/browse/RFE-6781)

### Solution

**Hand-written changes (5 files):**

| File | Change |
|------|--------|
| `api/v1beta1/externaldns_types.go` | Added `MaxResults int` to `ExternalDNSInfobloxProviderOptions` and `Interval *string` to `ExternalDNSSpec` with kubebuilder validation markers |
| `api/v1alpha1/externaldns_types.go` | Mirrored the same two fields |
| `pkg/operator/controller/externaldns/pod.go` | Added `--infoblox-max-results=N` in `fillInfobloxFields()` (only when MaxResults > 0) and `--interval=X` in `fillProviderAgnosticFields()` (only when Interval is non-nil) |
| `pkg/operator/controller/externaldns/deployment_test.go` | Added 4 new test cases: maxResults only, interval only, both together, maxResults=0 (suppressed) |
| `config/samples/infoblox/operator_v1beta1_infoblox_openshift.yaml` | Added commented examples for new fields |

**Auto-generated changes (3 files):**

| File | Generator |
|------|-----------|
| `api/v1alpha1/zz_generated.deepcopy.go` | `make generate` |
| `api/v1beta1/zz_generated.deepcopy.go` | `make generate` |
| `config/crd/bases/externaldns.olm.openshift.io_externaldnses.yaml` | `make manifests` |

**Documentation (1 file):**

| File | Change |
|------|--------|
| `docs/infoblox-openshift.md` | Documented `maxResults` and `interval` fields in the ExternalDNS CR creation step |

### Backward Compatibility

This change is **fully backward compatible**:

- Existing ExternalDNS CRs without `maxResults` or `interval` behave exactly as before -- no new flags are added to the container args
- All other providers (AWS, GCP, Azure, BlueCat) are completely unaffected -- `maxResults` is under the `infoblox` struct, `interval` only adds a flag when explicitly set
- No existing API fields are modified, renamed, or removed
- No default values change -- `MaxResults: 0` means "don't add the flag" (same as today), `Interval: nil` means "don't add the flag" (ExternalDNS uses its own default of 1m)
- Adding optional fields to a CRD is always backward compatible in Kubernetes; existing CRs remain valid with no migration needed
- **511 lines added, 0 lines deleted or modified**

### Customer Benefit

- **Unblocks large-scale Infoblox deployments** -- Enterprise customers with thousands of DNS records can now use the operator without encountering "Result set too large" errors
- **Reduces support cases** -- Provides a self-service fix for a common support ticket
- **Gives operational flexibility** -- Teams can tune polling frequency to match their SLA requirements without modifying operator-managed resources
- **Zero risk to existing users** -- Both fields are optional with safe defaults

### Unit Tests (47/47 PASS)

All 47 deployment test cases pass, including 4 new Infoblox-specific tests and all existing tests for every provider (AWS, Azure, GCP, BlueCat, Infoblox):

- `Infoblox with maxResults` -- set MaxResults=500, assert `--infoblox-max-results=500` in rendered args
- `Infoblox with interval` -- set Interval="5m", assert `--interval=5m` in rendered args
- `Infoblox with maxResults and interval` -- set both, assert both flags present
- `Infoblox with maxResults zero` -- set MaxResults=0, assert flag is NOT added

### Cluster Tests -- Operator Reconciliation (6/6 PASS)

Tested on a live OpenShift cluster (`api.sml.vmware.gsslab.pnq2.redhat.com`) with the operator image built and deployed:

1. **Both fields set** (maxResults=2000, interval="5m") -- both flags present in operand args
2. **Only maxResults** (maxResults=2000, interval removed) -- only `--infoblox-max-results` present
3. **Only interval** (interval="10m", maxResults=0) -- only `--interval` present
4. **Neither set** (both removed) -- neither flag present, behavior identical to unmodified operator
5. **maxResults=0** -- flag correctly suppressed
6. **Dynamic update** (1500/30s -> 3000/1h) -- operator reconciled and updated deployment args

### Cluster Tests -- CRD Schema Validation (3/3 PASS)

7. **Negative maxResults** (maxResults=-1) -- rejected: "should be greater than or equal to 0"
8. **Invalid interval** (interval="abc") -- rejected: "should match pattern"
9. **Complex interval** (interval="1h30m") -- accepted and passed as `--interval=1h30m`

### Cluster Tests -- Resilience (1/1 PASS)

10. **Pod restart** (operand pod deleted) -- new pod came up with both flags intact

**Result: 10/10 cluster tests PASS, 47/47 unit tests PASS**

Made with [Cursor](https://cursor.com)


swshende-cmd:
I have tested this and /verified by @swshende-cmd 

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Evidence from live cluster (OCP 4.20.0, External DNS Operator v1.3.5)
EVIDENCE 1: Container args generated by the operator
The operator created deployment external-dns-test-infoblox. The container args are:
--domain-filter=example.com
--metrics-address=127.0.0.1:7979
--txt-owner-id=external-dns-test-infoblox
--provider=infoblox
--source=openshift-route
--policy=sync
--registry=txt
--log-level=debug
--ignore-hostname-annotation
--fqdn-template={{""}}
--openshift-router-name=default
--infoblox-wapi-port=443
--infoblox-grid-host=10.0.0.1
--infoblox-wapi-version=2.3.1
--txt-prefix=external-dns-
--infoblox-max-results is MISSING. --interval is MISSING.
EVIDENCE 2: ExternalDNS config dump from pod logs
The pod logs show the full ExternalDNS config at startup. Two critical lines:
* InfobloxMaxResults:0 -- defaults to 0, meaning the Infoblox server default (1000) is used. There is no way to set it via the CR.
* Interval:1m0s -- defaults to 1 minute. There is no way to override it via the CR.
EVIDENCE 3: CRD schema confirms fields don't exist
Infoblox fields: ['credentials', 'gridHost', 'wapiPort', 'wapiVersion']
maxResults: NOT IN SCHEMA
interval (spec-level): NOT IN SCHEMA
All spec-level fields: ['domains', 'provider', 'source', 'zones']
The CRD has no maxResults under infoblox and no interval at the spec level. Even if a user tried to set them in YAML, they would be silently rejected by schema validation.
EVIDENCE 4: CR status is "healthy" but integration will fail
DeploymentAvailable: True
CredentialsSecretExists: True
The operator thinks everything is fine -- the Deployment is available. But if this were a real Infoblox server with >1000 records, the pod would crash with Result set too large (> 1000).

This is the proof that the operator has no mechanism to pass through --infoblox-max-results or --interval to the ExternalDNS container, exactly matching both customer reports.
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Direct problem it solves
Before the fix: Any customer using Infoblox with a DNS zone containing more than 1000 records is completely blocked. The ExternalDNS pod crashes on startup with a fatal error and never recovers. DNS records for OpenShift routes are never created or updated. There is no supported way to fix this -- the only workaround involves manually patching the Deployment and removing operator ownership, which is unsupported and fragile.
After the fix: Customers set maxResults: 5000 (or whatever value matches their zone size) in their ExternalDNS CR YAML, and it just works. The operator passes --infoblox-max-results=5000 to the ExternalDNS container, the WAPI query succeeds, and DNS records are created normally.

This is a small code change with high customer impact. It converts a non-functional, unsupported situation into a fully supported, one-line-of-YAML configuration. Every Infoblox customer with a production-scale DNS environment benefits.

swshende-cmd:
I also wanted to add the tests done summary:
Test Summary: --infoblox-max-results and --interval Support

Unit Tests
Ran 8 Infoblox-related unit tests (4 existing + 4 new) using go test. All passed. The 4 new cases cover: maxResults only, interval only, both together, and maxResults=0 (suppressed).
Cluster Tests — Operator Reconciliation (6 scenarios)
1. Both fields set (maxResults: 2000, interval: "5m") — Both --infoblox-max-results=2000 and --interval=5m appeared in the operand container args. PASS
2. Only maxResults (maxResults: 2000, interval removed) — Only --infoblox-max-results=2000 present, no --interval. PASS
3. Only interval (interval: "10m", maxResults: 0) — Only --interval=10m present, no --infoblox-max-results. PASS
4. Neither set (both fields removed) — Neither flag present. Behavior identical to unmodified operator. PASS
5. maxResults = 0 — Flag correctly suppressed, not added to container args. PASS
6. Dynamic update (changed from 1500/30s to 3000/1h) — Operator reconciled and updated the deployment args to --infoblox-max-results=3000 and --interval=1h. PASS
Cluster Tests — CRD Schema Validation (3 scenarios)
1. Negative maxResults (maxResults: -1) — Rejected by CRD validation with error: "should be greater than or equal to 0". PASS
2. Invalid interval (interval: "abc") — Rejected by CRD validation with error: "should match pattern ^([0-9]+(\.[0-9]+)?(s|m|h))+$". PASS
3. Complex interval (interval: "1h30m") — Accepted and correctly passed as --interval=1h30m. PASS
Cluster Tests — Resilience (1 scenario)
1. Pod restart (operand pod deleted) — New pod came up with both --infoblox-max-results=3000 and --interval=1h30m intact. PASS

Result: 10/10 PASS. All tests passed. The change is fully backward compatible and working as expected.

swshende-cmd:
Sorry if I am missing to create another story that it needs and to worked upon, please help me with that, I can do that with out any delay.

